PR #56, opened 2015-06-08
add enum type support for randoop

opticyclic:
cherry picked from from https://code.google.com/p/tpalus/source/detail?spec=svn122&r=122


mernst:
There are 742 commits in this pull request.  It looks like maybe errors in the history were fixed after you submitted it, and it should be re-submitted.

bjkeller:
The changes for enum support in this pull request treat the enum constants as primitive (non-message receiving) values. But, enums are a restricted form of class and *can* have methods, and so the constants should be registered as a kind of "statement" analogous to methods or constructors.

PR #57, opened 2015-10-08
Junit4 generation

bjkeller:
These are changes to transition Randoop to generate JUnit4 tests rather than JUnit3.
This primarily affects implementors of Check and ObjectContract as well as JunitFileWriter. I pulled the testing of the reflection flag out to GenTest, and now have separate methods for generating a single suite (for reflective case) and driver class (for non-reflective case). There is no longer a static JunitFileWriter.writeDriverFile method.


src/randoop/JunitFileWriter.java
```diff
@@ -5,12 +5,11 @@
 import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-
```
mernst: I personally like to put a blank line between groups of import statements, and JDK imports feel different than imports from this package itself.  It doesn't matter to me either way, but let's keep the files consistent.

src/randoop/JunitFileWriter.java
```diff
@@ -36,10 +35,12 @@
 
   private Map<String, List<List<ExecutableSequence>>> createdSequencesAndClasses = new LinkedHashMap<String, List<List<ExecutableSequence>>>();
 
+  private Map<String, Integer> methodCountMap = new LinkedHashMap<String,Integer>();
```
mernst: Please add a Javadoc comment for every method and field, to explain its meaning, purpose, or use.
(I realize this is not currently followed, but we can do better in code we are adding/changing.

src/randoop/JunitFileWriter.java
```diff
@@ -36,10 +35,12 @@
...
+
   public JunitFileWriter(String junitDirName, String packageName, String junitDriverClassName, int testsPerFile) {
     this.dirName = junitDirName;
-    this.packageName = packageName;
-    this.junitDriverClassName = junitDriverClassName;
+    this.packageName = packageName.trim();
```
mernst: Why does only this formal parameter need to be trimmed, but not other strings?

src/randoop/JunitFileWriter.java
```diff
@@ -36,10 +35,12 @@
 
   private Map<String, List<List<ExecutableSequence>>> createdSequencesAndClasses = new LinkedHashMap<String, List<List<ExecutableSequence>>>();
 
+  private Map<String, Integer> methodCountMap = new LinkedHashMap<String,Integer>();
+
   public JunitFileWriter(String junitDirName, String packageName, String junitDriverClassName, int testsPerFile) {
```
mernst: Given that you changed the field junitDriverClassName to masterTestClassName, shouldn't the formal parameter name be changed too?

src/randoop/JunitFileWriter.java
```diff
@@ -85,19 +87,12 @@ private void createOutputDir() {
    *
    * the default junit class name is the driver class name + index
    */
+
```
mernst: What is the point of this blank line?  I like to jam the comment against the method body, especially since the last line of the comment is practically a blank line already.  It's more standard -- but more importantly, it's what the rest of the codebase does, and consistency makes the code easier to read.

src/randoop/JunitFileWriter.java
```diff
@@ -105,46 +100,41 @@ private void createOutputDir() {
...
-        out.println();
-        // Replaced this printf by the below to avoid a dependence on Java
-        // 5 -- printf was added to the PrintStream class only in J2SE 5.0.
-        // out.println(indent("if (debug) System.out.printf(\"%n" + className + ".test" + testCounter + "\");"));
-        out.println(indent("if (debug) { System.out.println(); System.out.print(\"" + className + ".test" + testCounter + "\"); }"));
-        out.println();
-        out.println(indent(s.toCodeString()));
-        out.println("  }");
-        out.println();
         testCounter++;
```
mernst: I find it confusing that the counter is incremented in the middle of the loop, with parts of the operation of the loop before it and parts after it.  I suggest that the counter be incremented at the very beginning or very end of the loop, for clarity.

Even better, use
    for testcounter = 0; testcounter++; testcounter<sequences.size()) {
      ExecutableSequence s = sequences.get(testCounter);
      ...
    }

which is less efficient but easier to understand, and the efficiency shouldn't matter in this part of the code.
That way there is nothing at all clever that a reader has to follow.

src/randoop/JunitFileWriter.java
```diff
@@ -105,46 +100,41 @@ private void createOutputDir() {
...
+        writeTest(out, className, testCounter, s);
+        out.println();
       }
       out.println("}");
+      methodCountMap .put(className,testCounter);
```
mernst: Gratuitous space

src/randoop/JunitFileWriter.java
```diff
@@ -153,6 +143,19 @@ private File writeSubSuite(List<ExecutableSequence> sequencesForOneFile, String
...
   }
 
+/*
+ * Writes a test method to the output stream for the sequence s with name "test"+testCounter. 
+ */
+  private void writeTest(PrintStream out, String className, int testCounter, ExecutableSequence s) {
+    out.println("  @Test");
+    out.println("  public void test" + testCounter + "() throws Throwable {");
+    out.println();
+    out.println(indent("if (debug) { System.out.println(); System.out.print(\"" + className + ".test" + testCounter + "\"); }"));
```
mernst: It's weird that two spaces are used for two-space indent but indent() is used for 4-space indent.  I can see the need for indent() when its argument was generated elsewhere and has multiple lines, but it doesn't make so much sense to me here.  Maybe it's for consistency with the occurrence two lines below, which is necessary?

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
     return junitTestSuites;
   }
   
-  public static File writeDriverFile(File dir, String packageName, String driverClassName,
-      List<String> junitTestSuiteNames) {
-    File file = new File(dir, driverClassName + ".java");
+  /*
+   * writes JUnit4 suite based on list of class names
+   */
+  public File writeSuiteFile (List<String> testClassNames) {
```
mernst: As a minor point, elsewhere in the file, method names are abutted with their parameters.

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
     return junitTestSuites;
   }
   
-  public static File writeDriverFile(File dir, String packageName, String driverClassName,
-      List<String> junitTestSuiteNames) {
-    File file = new File(dir, driverClassName + ".java");
+  /*
+   * writes JUnit4 suite based on list of class names
```
mernst: Might this be multiple files?
Please write as a complete sentence with a capitalized first letter and final punctuation.
This needs to be a Javadoc comment, which starts with /** rather than just /*.

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
-      out.println("import junit.framework.*;");
-      out.println("import junit.textui.*;");
-      out.println("");
-      out.println("public class " + driverClassName + " extends TestCase {");
-      out.println("");
-      if (GenInputsAbstract.junit_reflection_allowed) {
-      out.println("  public static void main(String[] args) {");
-      } else {
-      out.println("  public static void main(String[] args) throws Throwable {");
+      outputPackageName(out,packageName);
```
mernst: Please separate argument names by a space.

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
+      
+      out.println();
+      out.println("import org.junit.runner.RunWith;");
+      out.println("import org.junit.runners.Suite;");
+      out.println();
+      out.println("@RunWith(Suite.class)");
+      out.println("@Suite.SuiteClasses({");
+     
+      Iterator<String> testIterator = testClassNames.iterator();
```
mernst: I would replace the next 7 lines by a call to join.  Theri is a [join](http://types.cs.washington.edu/plume-lib/api/plume/UtilMDE.html#join-java.util.List-java.lang.String-), though maybe a more canonical one than in plume-lib exists.
I would also put the commas at the ends of lines rather than at the beginnings of lines.

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
-      out.println("    TestResult result = runner.doRun(suite(), false);");
-      out.println("    if (! result.wasSuccessful()) {");
-      out.println("      System.exit(1);");
-      out.println("    }");
-      } else {
-      out.println("    new " + driverClassName + "(\"" + driverClassName + "\").runTest();");
+      
+      out.println("})");
+      out.println("public class " + suiteClassName + "{ }");
+    } finally {
```
mernst: I think a try-with-resources statement would be shorter and easier to read than the try-finally statement.
It's supported since Java 7, so it is fine to use.  (We are still avoiding use of Java 8 because some users haven't updated, but Java 7 is fair game.)

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
-      out.println("  public static Test suite() {");
-      out.println("    TestSuite result = new TestSuite();");
-      for (String junitTestsClassName : junitTestSuiteNames) {
-        out.println("    result.addTest(new TestSuite(" + junitTestsClassName + ".class));");
+    }
+    return file;
+  }
+  
+  /*
+   * write non-reflective driver for tests
```
mernst: Use capitalization and punctuation, here and elswhere, or the Javadoc will be unnecessarily hard to read.  Also use /** or else the comments won't be Javadoc comments at all.
You can use the Javadoc tag @see to refer to writeSubSuite.

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
+    File dir = this.getDir();
+    File file = new File(dir, masterTestClassName + "Driver.java");
+    PrintStream out = createTextOutputStream(file);
+    try {
+      outputPackageName(out, packageName);
+
+      out.println();
+      out.println("public class " + masterTestClassName + "Driver {");
+      out.println();
+      out.print("  public static void main(String[] args) {");
```
mernst: Shouldn't this be println so the following method call or return statement isn't jammed on the same line?

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
+        String testVariable = "t"+classIndex;
+        out.println(testClass + " " + testVariable + "= new " + testClass + "()");
+        
+        if (methodCountMap.containsKey(testClass)) {
+          int methodCount = methodCountMap.get(testClass);
+          for (int methodIndex = 1; methodIndex < methodCount; methodIndex++) {
+            String methodName = "test" + methodIndex;
+            out.println("try {");
+              out.println("  " + testVariable + "." + methodName + "()");
+              out.println("} catch (Throwable e) {");
```
mernst: This program always terminates successfully; the exit code is success regardless of whether tests fail.  Is that desirable?

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
+      for (String testClass : testClassNames) {
+        String testVariable = "t"+classIndex;
+        out.println(testClass + " " + testVariable + "= new " + testClass + "()");
+        
+        if (methodCountMap.containsKey(testClass)) {
+          int methodCount = methodCountMap.get(testClass);
+          for (int methodIndex = 1; methodIndex < methodCount; methodIndex++) {
+            String methodName = "test" + methodIndex;
+            out.println("try {");
+              out.println("  " + testVariable + "." + methodName + "()");
```
mernst: Why is writeTest (defined above) creating instance methods rather than static methods?  I would think that the test methods could be static (there is no state in the test class, is there?), and I think that would be clearer.

src/randoop/JunitFileWriter.java
```diff
@@ -153,6 +143,19 @@ private File writeSubSuite(List<ExecutableSequence> sequencesForOneFile, String
     return file;
   }
 
+/*
+ * Writes a test method to the output stream for the sequence s with name "test"+testCounter. 
+ */
+  private void writeTest(PrintStream out, String className, int testCounter, ExecutableSequence s) {
+    out.println("  @Test");
+    out.println("  public void test" + testCounter + "() throws Throwable {");
```
mernst: Why are the created tests instance methods?  Is there a reason they cannot be static methods?  Does JUnit require use of instance methods?

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
       }
-      out.println("    return result;");
-      out.println("  }");
-      } else {
-      out.println("  @Override");
-      out.println("  public void runTest() throws Throwable {");
-      for (String junitTestsClassName : junitTestSuiteNames) {
-        out.println("    new " + junitTestsClassName + "().runTest();");
+      
+      int classIndex = 0; //counter for test class object instances
```
mernst: I greatly prefer for-each loops to integral-bounded for loops.
However, I think it's even more important to keep all the loop iteration logic together and simple, so I would probably use a foreach loop here.  Otherwise, the reader will have to skim the code to understand how classindex is being updated.  I like to reserve loop-carried dependencies for unusual situations; they signal that something unusual is going on.

src/randoop/JunitFileWriter.java
```diff
@@ -195,58 +183,86 @@ public File writeDriverFile(String driverClassName) {
...
+      for (String testClass : testClassNames) {
+        String testVariable = "t"+classIndex;
+        out.println(testClass + " " + testVariable + "= new " + testClass + "()");
+        
+        if (methodCountMap.containsKey(testClass)) {
```
mernst: Under what circumstances might this fail?

If it cannot return false, then please don't put it in an if test that would silently mask an error.
If it can, write a brief comment saying how or why it would return false.

src/randoop/JunitFileWriter.java
```diff
@@ -254,6 +270,15 @@ public static File writeDriverFile(File dir, String packageName, String driverCl
     }
     return file;
   }
+  
+ 
+  /**
+   * Creates Junit tests for the faults.
```
mernst: What does "for the faults" mean here?  Do you mean tests?

src/randoop/JunitFileWriter.java
```diff
@@ -254,6 +270,15 @@ public static File writeDriverFile(File dir, String packageName, String driverCl
...
+   * Output is a set of .java files.
```
mernst: The return type is File rather than Set<File>.  Is the comment accurate?

src/randoop/main/GenTests.java
```diff
@@ -699,7 +699,7 @@ private void addClassLiterals(ComponentManager compMgr, List<Class<?>> allClasse
             compMgr.addGeneratedSequence(seq);
             break;
           default:
-            throw new Error("This can't happen");
+            throw new Error("Unexpected error in GenTests -- please report");
```
mernst: Could give the issue tracker URL

src/randoop/main/RandoopContractsTest.java
```diff
@@ -4,42 +4,45 @@
...
+    if (testResult.getFailureCount() != expectedFailures) {
+      StringBuilder b = new StringBuilder("RANDOOP TEST FAILED: EXPECTED GENERATED UNIT TESTS TO CAUSE " + expectedFailures + " FAILURES BUT GOT " + testResult.getFailureCount());
       b.append("\n\nJUNIT OUTPUT ON RANDOOP-GENERATED TESTS:");
       b.append(baos.toString());
       throw new RuntimeException(b.toString());
     }
 
-    if (result.errorCount() != expectedErrors) {
-      StringBuilder b = new StringBuilder("RANDOOP TEST FAILED: EXPECTED GENERATED UNIT TESTS TO CAUSE " + expectedErrors + " ERRORS BUT GOT " + result.errorCount());
+    /*
```
mernst: As with the above, you can delete this comment.

tests/randoop/test/SequenceTests.java
```diff
@@ -141,6 +141,7 @@ private static void parseRecord(List<String> lines) {
     checkListsEqual(expectedCode, Arrays.asList(ds.toCodeString().split(Globals.lineSep)), testId);
   }
   
+  @SuppressWarnings("deprecation")
```
mernst: Can you add a // comment on the same line indicating why this annotation is needed?
Should we update the code to eliminate the need for it?

src/randoop/JunitFileWriter.java
bjkeller: I added an Eclipse project specific setting for organize imports that recognizes randoop as a block, so running organize imports on a file will result in the randoop imports being separated from java ones.

src/randoop/JunitFileWriter.java
bjkeller: I changed the name to classMethodCounts

src/randoop/JunitFileWriter.java
bjkeller: I think we should be able to assume that by the time the package name gets here that it is clean. But probably should ensure that is true wherever the name comes from.

src/randoop/JunitFileWriter.java
bjkeller: The counter is an artifact of the name generation. I pulled the name generation into an inner class. (Yes, it does the same thing.)

src/randoop/JunitFileWriter.java
bjkeller: This was a refactored method, so the code is at least partly what was already there. The use of indent is old code. I'm going to leave the inconsistency now, since formatting is on my "maybe" list.

src/randoop/JunitFileWriter.java
bjkeller: Yes, it is verbose, but I'm going to leave this as is since it is the canonical solution within Java prior to Java 8.

src/randoop/JunitFileWriter.java
bjkeller: Generated code now captures whether any exceptions occurred, and if so has exit code of 1.

src/randoop/JunitFileWriter.java
bjkeller: Canonical JUnit4 test methods are instance methods.

src/randoop/JunitFileWriter.java
bjkeller: I haven't found anything that says why, but all JUnit4 examples are written as instance methods. The only reason it matters to us is the driver program generated by writeDriverFile, and that is a non-standard usage. I'm sticking with canonical JUnit4.

src/randoop/JunitFileWriter.java
bjkeller: This is another one of those name generation issues, which I resolved by using the inner class I originally added to generate method names. The explicit counter is gone.

src/randoop/JunitFileWriter.java
bjkeller: Removed. 
GenTest.write_junit_tests is called with a list of additional JUnit classes, which are then added to the list of generated tests. So, it would be possible for there to be classes for which we knew nothing of the methods. But, looking at the code, the only way this happens is through the plugin injecting other tests into the Randoop suite. Since the purpose of the driver is to make non-reflective Java that would not be used by the plugin, I removed the parameter and call the method getJunitTestSuiteNames directly. (at least until I change how that is done.)

src/randoop/JunitFileWriter.java
bjkeller: Preexisting comment.

src/randoop/JunitFileWriter.java
bjkeller: This is how the code was built previously. All of the methods return a collection of File objects so that the names can be printed out. I removed several of these methods, so this comment may be gone.

PR #62, opened 2015-11-19
refactor

bjkeller:
This is a refactoring branch consisting of essentially four kinds of changes:
1. class name changes
2. organization changes
3. refactoring of Statement to reduce dependencies from usage of Operation (formerly StatementKind) hierarchy in Sequence-related code. 
4. partitioning of Reflection based on function (outlined in last commit). What remains is just type information and I expect will go away as I deal with generics. The major part of this set of changes is the introduction of ReflectionManager, which can be used to standardize how classes and class members are visited prior to test generation.

Changes in the last two do affect behavior in terms of generated tests. The third allowing faux getter/setter objects to be treated as methods. And, the fourth affecting the order in which the Operation objects for class members are presented for testing.


bjkeller:
Travis failing because distribution build fails due to reorganization.  Will fix Makefile.

src/cov/Coverage.java
```diff
@@ -519,8 +519,13 @@ public static void initCoverage(Class<?> cls) {
    */
   public static Member getMemberContaining(CoverageAtom cov) {
     if (cov == null) throw new IllegalArgumentException("cov cannot be null.");
-    Class<?> cls = Reflection.classForName(cov.getClassName());
-    initCoverage(cls);
+
+    try {
+      initCoverage(Class.forName(cov.getClassName()));
+    } catch (ClassNotFoundException e) {
+      //ignor
```
mernst: Can you explain why this is ignored, or under what circumstances it will occur?

bjkeller: I was maintaining behavior in this case.

Looking at the code, the exception would occur if the class name of the CoverageAtom is not a class name, which appears unlikely. I changed the catch clause to return a null. This would match the intention in the comment.

src/randoop/FailureSet.java
```diff
@@ -5,17 +5,27 @@
...
+import randoop.sequence.Variable;
+
 public class FailureSet {
   
   private Set<Failure> failures = new LinkedHashSet<Failure>();
   
   public static class Failure {
-    public final StatementKind st;
+    
+    public final Statement st;
```
mernst: This isn't your fault because the documentation was missing before, but can you ensure that when you edit code it always has documentation for each field?
Maybe you needn't do this here because it is just a result of the rename refactoring.  I would have found the pull request easier to understand if you could break it into parts.  At least doing the rename first and then all the rest would help, because that would eliminate a bunch of lines from the diff.  (I hope it wouldn't be too hard to do all those refactorings on an unmodified version of the code.)

src/randoop/FailureSet.java
```diff
@@ -5,17 +5,27 @@
...
     public final Class<?> viocls;
-    public Failure(StatementKind st, Class<?> viocls) {
-      this.st = st;
+    
+    public Failure(Statement st2, Class<?> viocls) {
```
mernst: Please don't make this change.  It's much clearer to use the same variable name for the constructor formal parameter and the field name (and doing so is standard practice).

src/randoop/operation/MethodCall.java
```diff
@@ -340,13 +335,74 @@ public boolean isStatic() {
...
+    return list != null && list.contains(method);
+  }
+
+  /**
+   * callsTheMethod determines whether the method that this object calls is 
+   * method given in the parameter.
+   * @param m method to test against.
+   * @return true, if m corresponds to the method in this object, false, otherwise.
+   */
+  public boolean callsTheMethod(Method m) {
```
mernst: I would name this "callsMethod"

src/randoop/operation/MethodCall.java
```diff
@@ -340,13 +335,74 @@ public boolean isStatic() {
...
    */
+  @Override
   public String toParseableString() {
-    return Reflection.getSignature(method);
+    return this.getSignature();
   }
 
-  public static StatementKind parse(String s) {
-    return RMethod.getRMethod(Reflection.getMethodForSignature(s));
+  public String getSignature() {
```
mernst: For each method you add or modify, please write Javadoc documentation.

src/randoop/FailureSet.java
bjkeller: Sorry, didn't think about the confusion between renames and type substitutions.  This is the latter.

src/randoop/operation/MethodCall.java
```diff
@@ -340,13 +335,74 @@ public boolean isStatic() {
...
+   * method given in the parameter.
+   * @param m method to test against.
+   * @return true, if m corresponds to the method in this object, false, otherwise.
+   */
+  public boolean callsTheMethod(Method m) {
+    return method.equals(m);
   }
 
+  /**
+   * isMessage is a predicate to indicate that this method call is a message to a receiver object.
```
mernst: Can you clarify what is a message to a receiver object?  What wouldn't be one?
Should this use `@inheritDoc` instead?

src/randoop/operation/MethodParser.java
```diff
@@ -0,0 +1,51 @@
+package randoop.operation;
+
+import java.lang.reflect.Method;
+
+import randoop.types.TypeNames;
+
+/**
+ * MethodParser provides methods to recognize and extract methods from
```
mernst: If it's a library of static methods rather than representing anything, please state that explicitly here.

src/randoop/FailureSet.java
bjkeller: removed.  I will be more careful with Eclipse tools. Not sure where that came from.

src/randoop/operation/MethodParser.java
```diff
@@ -0,0 +1,51 @@
...
+  public static Method getMethodForSignature(String signature) throws OperationParseException {
+    if (signature == null) {
+      throw new IllegalArgumentException("signature may not be null");
+    }
+    
+    int openPar = signature.indexOf('(');
+    int closePar = signature.indexOf(')');
+    // Verify only one open/close paren, and close paren is last char.
+    assert openPar == signature.lastIndexOf('(') : signature;
+    assert closePar == signature.lastIndexOf(')') : signature;
```
mernst: Isn't this redundant with the next line?

src/randoop/operation/MethodParser.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * MethodParser provides methods to recognize and extract methods from
+ * string representations of their signature.
+ * 
+ * @author bjkeller
+ *
+ */
+public class MethodParser {
+
+  public static Method getMethodForSignature(String signature) throws OperationParseException {
```
mernst: Don't forget the documentation.  For example, what is the format of the parameter?  Java has 3 or 4 different representations for signatures.

src/randoop/operation/NonreceiverTerm.java
```diff
@@ -118,34 +123,24 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
...
-      b.append(newVar.getName());
-      b.append(" = ");
-      b.append(PrimitiveTypes.toCodeString(getValue()));
-      b.append(";");
-      b.append(Globals.lineSep);
-    }
+  /**
+   * appendCode adds the string representation of the Nonreceiver value
+   * to the StringBuilder.
+   * Note: do not explicitly box primitive values.
```
mernst: I don't understand this or how to take action on it.

src/randoop/operation/Operation.java
```diff
@@ -0,0 +1,121 @@
...
+
+import java.io.PrintStream;
+import java.util.List;
+
+import randoop.ExecutionOutcome;
+import randoop.reflection.ReflectionPredicate;
+import randoop.sequence.Variable;
+
+/**
+ * Represents an operation (in the term algebra sense) that can occur in a statement
```
mernst: Citation for people who don't understand term algebra?  Or examples?  I'd like to keep the code readable for ordinary programmers.

src/randoop/operation/Operation.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Represents an operation (in the term algebra sense) that can occur in a statement 
+ * as part of a test sequence. Operations include method calls, constructor calls, field
+ * accesses, enum constant values, or primitive values.
+ * 
+ * To support text-based serialization, an implementing class C should also provide:
+ * 
+ *  <ul>
+ *  <li> A public static String field that contains a unique ID for the statement kind.
```
mernst: Named what?

src/randoop/operation/Operation.java
```diff
@@ -0,0 +1,121 @@
...
+import randoop.ExecutionOutcome;
+import randoop.reflection.ReflectionPredicate;
+import randoop.sequence.Variable;
+
+/**
+ * Represents an operation (in the term algebra sense) that can occur in a statement 
+ * as part of a test sequence. Operations include method calls, constructor calls, field
+ * accesses, enum constant values, or primitive values.
+ * 
+ * To support text-based serialization, an implementing class C should also provide:
```
mernst: Should all implementations do so?  What are the tradeoffs and how does one choose whether to do this?  Or is it required?  In that case, if the method is abstract an implementing class must do so regardless.

src/randoop/operation/PublicField.java
```diff
@@ -160,5 +164,20 @@ public void setValue(Object object, Object value) {
   protected Object writeReplace() throws ObjectStreamException {
     return new SerializablePublicField(field);
   }
+
+  public boolean isStatic() {
```
mernst: Every method is worth at least a one-liner Javadoc description.

src/randoop/operation/PublicField.java
```diff
@@ -160,5 +164,20 @@ public void setValue(Object object, Object value) {
...
+    return false;
+  }
+
+  /**
+   * satisfies checks whether the enclosed {@link Field} object satisfies
```
mernst: Conventionally, one does not mention the method name in the Javadoc.
If you want to retain it, then put "satisfies" in tt font to make the sentence easier to parse.

src/randoop/operation/SerializableArrayCreation.java
```diff
@@ -0,0 +1,25 @@
+package randoop.operation;
+
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+
+import randoop.types.TypeNames;
+
+public class SerializableArrayCreation implements Serializable {
```
mernst: I realize this is a renaming, but can you document it and/or its relationship to ArrayCreation.

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,35 @@
...
+
+    // If inner classes are involved, Class.getName() will return
+    // a string with "$" characters. To make it compilable, must replace with
+    // dots.
+    retval = retval.replace('$', '.');
+
+    return retval;
+  }
+  
+  public static Class<?> recognizeType(String typeName) throws ClassNotFoundException {
```
mernst: This method is new but not documented.

src/randoop/reflection/VisibilityPredicate.java
```diff
@@ -0,0 +1,15 @@
+package randoop.reflection;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+public interface VisibilityPredicate {
```
mernst: It's especially important to document anything that is an interface or abstract class or intended to be overridden.  I came here trying to understand some other code but this didn't explain my questions.

src/randoop/sequence/ExecutableSequence.java
```diff
@@ -392,6 +391,10 @@ public static boolean getRuntimeInputs(Sequence s, List<ExecutionOutcome> outcom
     return true;
   }
   
+  public static Object[] getRuntimeValuesForVars(List<Variable> vars, Execution execution) {
```
mernst: What is the format for the variables?  (More generally, please add documentation.)

src/randoop/sequence/ExecutableSequence.java
```diff
@@ -860,4 +863,32 @@ public static int checks_count (List<ExecutableSequence> seqs) {
 
     return cnt;
   }
+
+  /**
+   * initialize - used by ExecutionVisitors
+   */
+  public void initializeResults() {
+    // TODO Auto-generated method stub
```
mernst: This was auto-generated?

src/randoop/sequence/ExecutableSequence.java
```diff
@@ -860,4 +863,32 @@ public static int checks_count (List<ExecutableSequence> seqs) {
 
     return cnt;
   }
+
+  /**
+   * initialize - used by ExecutionVisitors
```
mernst: What does it do?
(Knowing how it is supposed to be used is useful but not enough.)

src/randoop/sequence/ForwardGenerator.java
```diff
@@ -566,17 +575,16 @@ private InputsAndSuccessFlag selectInputs(StatementKind statement) {
       // If we were unlucky and selected a null value as the receiver
       // for a method call, return with failure.
       if (i == 0
-          && (statement instanceof RMethod)
-          && (!((RMethod) statement).isStatic())
-          && chosenSeq.getCreatingStatement(randomVariable) instanceof PrimitiveOrStringOrNullDecl)
+          && operation.isMessage() && !(operation.isStatic())
```
mernst: I would find it easier to read these if every clause was on its own line.

src/randoop/sequence/MappedSequences.java
```diff
@@ -57,4 +59,11 @@ public void addSequence(T key, Sequence seq) {
     emptyList = new ListOfLists<Sequence>(emptyJDKList);
   }
 
+  public Set<Sequence> getAllSequences() {
```
mernst: Add documentation.

src/cov/Coverage.java
mernst: Under what circumstances is CoverageAtom not a class name?  If that situation is an error, then I feel strongly that Randoop should throw an exception rather than silently masking the error.
Also, when I asked the circumstances, I meant to document them in the code rather than just in the code review conversation, but that might not have been clear.

src/randoop/operation/NonreceiverTerm.java
bjkeller: I moved the generation of the code for initialization and assignment into the Statement class, so the only remaining part of the method is the call to PrimitiveTypes.toCodeString(-).

PR #65, opened 2015-11-23
Review of refactor

bjkeller:
[NOT TO BE MERGED]
This is a merge of master into the refactoring branch, and is very broken.
The intent is for review of refactoring changes, and the stmt-refactoring branch will be deleted.


src/cov/Coverage.java
```diff
@@ -528,12 +537,17 @@ public static Member getMemberContaining(CoverageAtom cov) {
...
-      getFalseBranches(cls)[br.branchNumber]++;
+    try {
+      Class<?> cls = Class.forName(br.getClassName());
+      if (br.branch) {
+        getTrueBranches(cls)[br.branchNumber]++;
+      } else {
+        getFalseBranches(cls)[br.branchNumber]++;
+      }
+    } catch (ClassNotFoundException e) {
+      //ignore
```
mernst: I don't think we should ignore any thrown exceptions.  Can you throw an exception here?

src/randoop/FailureSet.java
```diff
@@ -6,35 +6,56 @@
...
+   *
+   */
   public static class Failure {
-    public final Operation st;
-    public final Class<?> viocls;
-    public Failure(Operation st, Class<?> viocls) {
-      this.st = st;
-      this.viocls = viocls;
+    
+    /**
```
mernst: It's up to you, but for one-line Javadoc comments I put them all on one line:

    /** .... */

rather than

    /**
     * ...
     */

src/randoop/FailureSet.java
```diff
@@ -6,35 +6,56 @@
...
+     */
+    public final Class<?> violationClass;
+    
+    public Failure(Statement st, Class<?> viocls) {
+      this.statement = st;
+      this.violationClass = viocls;
     }
+    
     public boolean equals(Object o) {
       if (o == null) return false;
```
mernst: This isn't your fault because the poor code was alraeady there, but this will crash if called with a non-Failure argument.  The canonical way that most `equals()` methods should start is

    if (! o instanceof Failure) {
      return null;
    }

which doesn't crash and also makes the `o == null` test unnecessary (since `null` in an `instanceof` expression yields `false`).

src/randoop/operation/NonreceiverTerm.java
```diff
@@ -164,17 +164,22 @@ public Object getValue() {
...
   /**
-   * Returns the appropriate PrimitiveOrStringOrNullDeclInfo representative of 
+   * createCanonicalTerm returns a NonreceiverTerm representative of 
    * the specified class c.
+   * 
+   * @param c type of value desired.
+   * @return a {@link NonreceiverTerm} with a canonical representative of the given type.
    */
-  public static NonreceiverTerm nullOrZeroDecl(Class<?> c) {
+  public static NonreceiverTerm createCanonicalTerm(Class<?> c) {
```
mernst: I find the name "canonical" confusing here; it makes me think of canonicalization, which is not what this does.  For this one, I prefer the original name.  In the change, the documentation and name doesn't indicate that it will always be null or zero.

src/randoop/SerializableCheckRepContract.java
```diff
@@ -4,19 +4,25 @@
...
 import java.lang.reflect.Method;
 
-import randoop.util.Reflection;
+import randoop.operation.MethodSignatures;
+import randoop.operation.OperationParseException;
 
+/**
+ * Serialized form of {@link CheckRep} allowing tests to be serialized.
+ * 
+ * Also see <code>CheckRep.writeReplace</code>.
```
mernst: Use `{@link CheckRep#writeReplace}` Javadoc formatting, rather than `<code>...</html>` HTML markup.

src/randoop/operation/MethodCall.java
```diff
@@ -94,40 +95,32 @@ public MethodCall(Method method) {
   }
 
   /**
-   * Returns the statement corresponding to the given constructor.
+   * getMethodCall a static method that creates a {@link MethodCall} corresponding
```
mernst: In the first word here a typo?
I don't think you need to state it is a static method -- that will be obvious to anyone who reads this documentation.  I would just start, "Creates a ...".
This documentation is much better than the misleading previous documentation.  However, the name is still confusing.  This is a creator (constructor) method, so I propose that its name should start with "create" or "make" or "new" rather than "get" which implies that it returns something that already exists.  Please see if any other methods should also be ranamed.
Also, this renaming and a few others appear on a disproportionate number of lines (at least in the part of the diff I have read so far).  If true, would it make sense to create a pull request for those renamings in order to simplify the other changes?

src/randoop/experiments/RandomWalkGenerator.java
```diff
@@ -85,8 +84,7 @@
   private long gentime;
   private long exectime;
 
-  private static SequenceCollection prims = new SequenceCollection(
-      SeedSequences.defaultSeeds());
+  private static SequenceCollection prims = new SequenceCollection(SeedSequences.defaultSeeds());
```
mernst: Your change is fine and does not need to be modified.
If I were to break the line, I would personally do it before the "=".

src/randoop/experiments/RandomWalkGenerator.java
```diff
@@ -571,7 +569,7 @@ void repInvariantCheck2() {
       assert st != null;
     }
 
-    // TODO move this to repCheck for List<Operation>.
+    // TODO move this to repCheck for List<StatementKind>.
```
mernst: Should `repCheck` be `checkRep`?

src/randoop/main/ClassTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import randoop.types.TypeNames;
+import randoop.util.Files;
+
+import plume.EntryReader;
+
+/**
+ * ClassTypeLoader provides methods that allow a list of class names to be loaded
```
mernst: You mean loaded into the JVM?  I would clarify that.  (When I first read it, I thought it might be loading from a file.)

src/randoop/main/ClassTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+ * as {@link Class<?>} objects.
+ *
```
mernst: Blank lines are ignored in the generated Javadoc.  You need to write `<p>` in order to start a new paragraph.  (I typically put it at the end of the previous paragraph rather than the start of the new one, but I don't write it at the end of the last paragraph.)

src/randoop/main/ClassTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * ClassTypeLoader provides methods that allow a list of class names to be loaded
+ * as {@link Class<?>} objects.
+ * 
+ * (Extracted from class previously known as randoop.util.Reflection.)
+ * 
+ * @author bjkeller
+ *
+ */
+public class ClassTypeLoader {
```
mernst: Why is this called *ClassType* Loader?  I'm not sure it has anything to do with types.

src/randoop/main/ClassTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * loadClassesFromList returns a list of {@link Class} objects given a list of class names.
+   * In the case that the typename recognition throws an exception, an error is only thrown if
+   * <code>noerr</code> is false.
+   * 
+   * @param classNames list of strings for classnames.
```
mernst: I would just say "list of fully-qualified class names".  It will be obvious they are strings.  If you want to say that, put it later rather than first because it's not the key information.

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
...
+    // If inner classes are involved, Class.getName() will return
+    // a string with "$" characters. To make it compilable, must replace with
+    // dots.
+    retval = retval.replace('$', '.');
+
+    return retval;
+  }
+  
+  /**
+   * Recognizes a type from a string and returns the {@link Class} object for the type.
```
mernst: "Recognize" is not a standard term of art.  I would just say "Returns the {@link Class} for the given fully-qualified class name or primitive type name."

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
...
+   * Handles primitive types as well.
+   * 
+   * @param typeName string representation of a type.
```
mernst: Is it fully-qualified?

src/randoop/main/ClassTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param noerr a boolean flag to indicate whether to report classes not found.
+   * @return list of {@link Class} objects corresponding to elements of list.
+   */
+  public static List<Class<?>> loadClassesFromList(List<String> classNames, boolean noerr) {
+    List<Class<?>> result = new ArrayList<Class<?>>(classNames.size());
+    for (String className : classNames) {
+      try {
+        Class<?> c = TypeNames.recognizeType(className);
```
mernst: I mention this lower down, but it's confusing that this method is named "load...", it calls one named "recognize...", and that calls two named "forName".  Let's use consistent names for these throughout.  Ones with "forName" seem reasonable to me, especially since that is the name used in the JDK.

src/randoop/main/GenBranchDir.java
```diff
@@ -181,9 +182,13 @@ public static void main(String[] args) {
     }
     Set<CoverageAtom> allBranches = new LinkedHashSet<CoverageAtom>();
     for (String className : covClassNames) {
-      Class<?> cls = Reflection.classForName(className);
+      try {
+      Class<?> cls = TypeNames.recognizeType(className);
       covClasses.add(cls);
       allBranches.addAll(Coverage.getBranches(cls));
+      } catch (ClassNotFoundException e) {
+        //ignore
```
mernst: Every time I see a catch clause ignoring a thrown exception, it gives me the heebie jeebies.  Can you please eliminate all uses of that undesirable code pattern from Randoop?

src/randoop/main/GenTests.java
```diff
@@ -777,6 +786,7 @@ private void addClassLiterals(ComponentManager compMgr, List<Class<?>> allClasse
 
   /**
    * Execute the init routine (if user specified one)
+   * @param phase
```
mernst: Empty Javadoc for this silences the compiler warning but doesn't help a programmer...

src/randoop/operation/AbstractOperation.java
```diff
@@ -0,0 +1,59 @@
...
+ * AbstractOperation is an abstract implementation of the Operation interface
+ * to provide default implementations of Operation predicates that are false
+ * except for a few kinds of operations.
+ * 
+ * @author bjkeller
+ *
+ */
+public abstract class AbstractOperation implements Operation {
+
+  @Override
```
mernst: For every method, please add a Javadoc comment -- at least `@inheritDoc` so it shows up in the generated HTML API documentation.

src/randoop/operation/AbstractOperation.java
```diff
@@ -0,0 +1,59 @@
...
+    return false;
+  }
+  
+  @Override
+  public Object getValue() {
+    throw new IllegalArgumentException("No value for this kind of operation.");
+  }
+  
+  /**
+   * compareTo(Operation) compares two {@link Operation} objects.
```
mernst: I would use either `@inheritDoc` or `@link` to point to the superclass, and then say "This implementation uses ..."

src/randoop/operation/AbstractOperation.java
```diff
@@ -0,0 +1,59 @@
...
+   * Uses canonical string representation of operations to compare.
+   */
+  @Override
+  public int compareTo(Operation op) {
+    return (this.toParseableString()).compareTo(op.toParseableString());
+  }
+
+  /**
+   * satisfies checks to see if reflective object contained in this {@link Operation}
```
mernst: I have trouble parsing the first sentence without `sasisfies` either within <code>...</code> or omitted.

src/randoop/operation/AbstractOperation.java
```diff
@@ -0,0 +1,59 @@
...
+   * satisfies the predicate. Since there is no reflective object, this returns false.
```
mernst: Please add "in an AbstractOperation" after "object", for clarity.

src/randoop/operation/ArgumentParser.java
```diff
@@ -0,0 +1,36 @@
+package randoop.operation;
+
+import randoop.types.TypeNames;
+
+/**
+ * ArgumentParser recognizes method/constructor arguments in a string
```
mernst: What kind of string?  A method signature?  In what format?  Also, do you mean it recognizes parameters or arguments?  Or maybe parameter types?  Please indicate the exact format expected of the input, and give a concrete example string in the documentation.
Don't forget to have some kind of documentation for the method itself, as well.

src/randoop/operation/ArrayCreation.java
```diff
@@ -9,20 +9,23 @@
...
 
 /**
- * Immutable.
- * Represents a one-dimensional
- * array creation statement, e.g. "int[] x = new int[2] { 3, 7 };"
+ * ArrayCreation is an {@link Operation} representing the construction of a one-dimensional
+ * array such as "int[] x = new int[2] { 3, 7 };"
+ * 
+ * In terms of the notation used for {@link Operation}, an array creation of an array of elements 
+ * of type e and length n has a signature [e,...,e] -> t where [e,...,e] is a list of length n,
```
mernst: What might _n_ be?  Is it hardcoded to some specific set of small values?

src/randoop/operation/ConstructorCall.java
```diff
@@ -7,51 +7,39 @@
...
 import randoop.util.Util;
 
 /**
- * Represents a constructor call. The inputs are parameters to the constructor
- * and outputs include the new object.
- *
- * The "R" stands for "Randoop", to underline the distinction from
- * java.lang.reflect.Constructor.
- *
+ * ConstructorCall is an {@link Operation} that represents a call to a constructor, and holds a reference to
```
mernst: If you are creating new comment lines, then I prefer to keep them to 80 columns if possible, for consistency and readability in many terminals/printouts.

src/randoop/operation/ConstructorCall.java
```diff
@@ -7,51 +7,39 @@
...
+ * a reflective {@link java.lang.reflect.Constructor} object.  
+ * 
+ * As an {@link Operation}, a call to constructor c with n arguments is represented as 
+ * c : [t1,...,tn] -> c, where the output type c is also the name of the class.
```
mernst: c appears twice at the beginning of this line.  Is the first one representing "new C"?

src/randoop/operation/ArrayCreation.java
```diff
@@ -120,45 +130,47 @@ public String toStringVerbose() {
...
 
   /**
-   * Returns constraint to represent new reference to this statement,
-   * namely the receiver that is generated.
+   * {@inheritDoc}
+   * @return type of created array.
    */
+  @Override
   public Class<?> getOutputType() {
-    if (outputTypeCached == null) {
```
mernst: This deletion does not look right to me.

src/randoop/operation/ConstructorCall.java
```diff
@@ -170,11 +161,12 @@ public boolean equals(Object o) {
     if (!(o instanceof ConstructorCall))
```
mernst: See comment elsewhere about equals methods.  This instanceof test should be first, and then the null test is not needed any more.  (The `this == o` test is still worthwhile; it could go before or after the instanceof test.)

src/randoop/operation/ConstructorCall.java
```diff
@@ -216,31 +220,68 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
   }
 
   /**
-   * Returns the return type of this constructor.
+   * {@inheritDoc}
+   * @return type of the object created (i.e., class for constructor).
    */
+  @Override
   public Class<?> getOutputType() {
-    if (outputTypeCached == null) {
```
mernst: This deletion seems suspicous; see comment about a similar deletion above.

src/randoop/operation/ConstructorCall.java
```diff
@@ -216,31 +220,68 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
...
-   * CONSTRUCTOR
+   * {@inheritDoc}
+   * Generates a string representation of the constructor signature.
    *
-   * Where CONSTRUCTOR is a string representation of the constrctor signature. Examples:
-   *
-   * java.util.ArrayList.&lt;init&gt;()
-   * java.util.ArrayList.&lt;init&gt;(java.util.Collection)
+   * Examples:
+   *  java.util.ArrayList.&lt;init&gt;()
```
mernst: To put these examples on separate lines in the generated HTML, you should probably put them within <pre>...</pre>.

src/randoop/operation/ConstructorCall.java
```diff
@@ -216,31 +220,68 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
...
+   */
+  @Override
+  public boolean isConstructorCall() {
+    return true;
+  }
+
+  /**
+   * {@inheritDoc}
+   * Determines whether enclosed {@link Constructor<?>} satisfies the given predicate.
+   * @return true only if the constructor in this object satisfies the canUse(Constructor) of predicate.
```
mernst: turn `canUse` into an `@link`?

src/randoop/operation/ConstructorSignatures.java
```diff
@@ -0,0 +1,76 @@
+package randoop.operation;
+
+import java.lang.reflect.Constructor;
+
+/**
+ * ConstructorParser provides static methods to write as well as recognize
```
mernst: Change "recognize and extract" to "parse".
(The class creates and parses string representations of the constructor.)

src/randoop/operation/AbstractOperation.java
bjkeller: if you don't add anything then the inheritdoc is redundant.
I will double-check the javadoc though

src/randoop/operation/ConstructorCall.java
bjkeller: I just went ahead and saved the type into the variable rather than checking every call whether it had been saved into the variable.

PR #67, opened 2015-11-25
For Review: st-refactor changes from 11/24

bjkeller:
This is current set of changes based one comments to PR65


src/cov/Coverage.java
```diff
@@ -512,15 +513,23 @@ public static void initCoverage(Class<?> cls) {
 
 
   /**
-   *
+   * getMemberContaining retrieves the class member that contains the {@link CoverageAtom}
+   * 
    * @param cov
    * @return Returns null if this coverage atom does not belong to a method or
    *         constructor.
+   * @throws IllegalArgumentException if {@link CoverageAtom} argument is null.
```
mernst: Would it be better to specify that the cov parameter should be non-null (in its `@param` block), rather than specifying precisely what happens if it is?

src/cov/Coverage.java
```diff
@@ -528,12 +537,17 @@ public static Member getMemberContaining(CoverageAtom cov) {
...
-    } else {
-      getFalseBranches(cls)[br.branchNumber]++;
+    try {
+      Class<?> cls = Class.forName(br.getClassName());
+      if (br.branch) {
+        getTrueBranches(cls)[br.branchNumber]++;
+      } else {
+        getFalseBranches(cls)[br.branchNumber]++;
+      }
+    } catch (ClassNotFoundException e) {
```
mernst: Please narrow the scope of every try...catch block to be as small as possible -- in this case, to just be the Class.forName call.  That makes its purpose more clear.

src/cov/Coverage.java
```diff
@@ -528,12 +537,17 @@ public static Member getMemberContaining(CoverageAtom cov) {
...
+      throw new BugInRandoopException("Error in Coverage.touch: " + e);
     }
+
```
mernst: Is this added whitespace necessary?

src/randoop/FailureSet.java
```diff
@@ -6,35 +6,55 @@
...
+ * {@link ExecutableSequence}. 
+ *
+ */
 public class FailureSet {
   
   private Set<Failure> failures = new LinkedHashSet<Failure>();
   
+  /**
+   * FailureSet.Failure represents an execution failure and holds the
+   * statement and class where the failure occurred.
```
mernst: I don't think you mean the class where the failure occurred in the same sense as the statement where the failure occurred.  (Zeugma is good in fiction prose but bad in technical writing!)  Is it the class in which the failure occurred, or a class that represents the type of failure?

src/randoop/FailureSet.java
```diff
@@ -6,35 +6,55 @@
...
-      this.st = st;
-      this.viocls = viocls;
+    
+    /**
+     * statement is the statement in the {@link ExecutableSequence} where the failure occurred.
+     */
+    public final Statement statement;
+    
+    /**
+     * violationClass is the type of the {@link ObjectContract}
```
mernst: Add "that was violated"?

src/randoop/FailureSet.java
```diff
@@ -6,35 +6,55 @@
...
     }
+    
     public boolean equals(Object o) {
-      if (o == null) return false;
-      if (o == this) return true;
-      Failure other = (Failure)o;
-      if (!st.equals(other.st)) return false;
-      if (!viocls.equals(other.viocls)) return false;
-      return true;
+      if (o instanceof Failure) {
```
mernst: It's more idiomatic to test `if (! o instanceof Failure)` and immediately return false.  That uses a common coding pattern, and it keeps from having long indented blocks that the reader must keep track of.  I suggest using the standard idiom throughout (even though I agree yours is equally correct!).

src/randoop/JunitFileWriter.java
```diff
@@ -175,6 +178,11 @@ private void writeTest(PrintStream out, String className, String methodName, Exe
     out.println("  }");
   }
   
+  /**
+   * Generates the list of test class names for previously generated test suite.
```
mernst: "suite" => "suites"

src/randoop/SerializableExpectedExceptionChecker.java
```diff
@@ -3,8 +3,14 @@
...
 
-import randoop.util.Reflection;
-
+import randoop.types.TypeNames;
+
+/**
+ * Serialized form of {@link ExpectedExceptionChecker} allowing tests to
+ * be serialized.
+ * 
+ * Also see <code>ExpectedExceptionChecker.writeReplace</code>.
```
mernst: No such class.  I think you mean ExpectedExceptionCheck.  Please change to a `@link` and make sure that "make javadoc" runs without any warnings.

src/randoop/SerializableExpectedExceptionChecker.java
```diff
@@ -17,9 +23,9 @@ public SerializableExpectedExceptionChecker(Class<? extends Throwable> exception
     this.statementIdx = statementIdx;
   }
 
-  private Object readResolve() throws ObjectStreamException {
-    @SuppressWarnings("unchecked")
-    Class<? extends Throwable> c = (Class<? extends Throwable>) Reflection.classForName(exceptionClass);
+  @SuppressWarnings("unchecked")
```
mernst: What is the reason for changing the `@SuppressWarnings`?  Before it applied to only one variable declaration; now it applies to the whole body of the method.  I think it would be better to keep its scope as small as possible, for clarity as to what it is doing and to avoid suppressing unintended warnings.

src/randoop/experiments/Execute.java
```diff
@@ -32,9 +32,13 @@ public static void main(String[] args) {
       throw new Error(e);
     }
     for (String className : covClassNames) {
-      Class<?> cls = Reflection.classForName(className);
-      System.out.println(cls.toString() + " " + Coverage.getBranches(cls).size());
-      covClasses.add(cls);
+      try {
```
mernst: Can you make the try clause cover fewer lines?  It doesn't need to cover the `covclasses.add(cls);" line, and I don't think it is necessary for the `getBranches` line either.  More generally, please make sure all try clauses (or at least the ones you added) have minimal scope.

src/randoop/main/GenBranchDir.java
```diff
@@ -181,9 +182,13 @@ public static void main(String[] args) {
     }
     Set<CoverageAtom> allBranches = new LinkedHashSet<CoverageAtom>();
     for (String className : covClassNames) {
-      Class<?> cls = Reflection.classForName(className);
-      covClasses.add(cls);
-      allBranches.addAll(Coverage.getBranches(cls));
+      try {
```
mernst: Scope of try block seems too big.

src/randoop/main/GenTests.java
```diff
@@ -183,17 +189,23 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
 
     // Remove private (non-.isVisible) classes and abstract classes
     // and interfaces.
     List<Class<?>> classes = new ArrayList<>(allClasses.size());
     for (Class<?> c : allClasses) {
-      if (Reflection.isAbstract (c)) {
+      if (Modifier.isAbstract (c.getModifiers()) && !c.isEnum()) {
         System.out.println("Ignoring abstract " + c + " specified via --classlist or --testclass.");
-      } else if (! Reflection.isVisible (c)) {
+      } else if ( !visibility.isVisible (c) ) {
```
mernst: I would leave the space between the `!` operator and the expression it applies to.  Otherwise it's too easy to overlook, is harder to read (in my opinion), and is inconsistent with elsewhere in the file.
I would remove the added space at the end of the line, however -- again, for consistency.

src/randoop/main/TypeReader.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import randoop.types.TypeNames;
+import randoop.util.Files;
+
+import plume.EntryReader;
+
+/**
+ * ClassReader is a library of static methods that converts a list of
```
mernst: Class is named `TypeReader`.  However, since it returns a list of `Class` objects, renaming it to `ClassReader` makes sense to me.

src/randoop/main/TypeReader.java
```diff
@@ -0,0 +1,102 @@
...
+ * (Extracted from class previously known as randoop.util.Reflection.)
+ */
+public class TypeReader {
+
+  /**
+   * Returns a list of {@link Class} objects given a list of class names.
+   * In the case that the type name recognition throws an exception, an error is only thrown if
+   * <code>ignoreBadNames</code> is false.
+   * 
+   * @param classNames  the list of fully qualified class names.
```
mernst: "fully-qualified" has a hyphen

src/randoop/main/TypeReader.java
```diff
@@ -0,0 +1,102 @@
...
+   * @param ignoreBadNames  a flag to indicate that {@code ClassNotFoundExceptions} should be ignored. 
+   * @return list of {@link Class} objects corresponding to elements of list.
+   */
+  public static List<Class<?>> getTypesForNames(List<String> classNames, boolean ignoreBadNames) {
```
mernst: I would call this "getClassesForNames" since it returns a list of Classes.

src/randoop/main/UniversalDriverHandler.java
```diff
@@ -31,6 +31,7 @@ public boolean handle(String[] args) throws RandoopTextuiException {
 
     Options parsedArgs = new Options(usage_synopsis,
         WriteModelCheckerDriver.class);
+    @SuppressWarnings("unused")
```
mernst: For every `@SuppressWarnings`, please write a code comment on the same line, starting with //, explaining why it is necessary.

src/randoop/main/UniversalDriverHandler.java
```diff
@@ -31,6 +31,7 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
     String non_options[] = null;
```
mernst: This isn't your fault, but could you remove the initialization?  It isn't necessary because of the assignment in a following line.

src/randoop/operation/ArrayCreation.java
```diff
@@ -9,20 +9,30 @@
...
+ * <code>int[] x = new int[2] { 3, 7 };</code><br>
+ * with the initializer list as inputs.
+ * <p>
+ * In terms of the notation used for {@link Operation}, an array creation of an 
+ * array of elements of type <i>e</i> and length <i>n</i> has a signature 
+ * [<i>e,...,e</i>] &rarr; <i>t</i> 
+ * where [<i>e,...,e</i>] is a list of length <i>n</i>, and <i>t</i> is the 
+ * array type.
+ * <p>
+ * Objects are immutable.
```
mernst: I would add "ArrayCreation" at the front of this sentence.

src/randoop/operation/ArrayCreation.java
```diff
@@ -9,20 +9,30 @@
...
+ * ArrayCreation is an {@link Operation} representing the construction of a 
+ * one-dimensional array with a given element type and length.
+ * The operation requires a list of elements in an initializer.
+ * For instance, <code>new int[2]</code> is the {@code ArrayCreation} in the 
+ * initialization<br>
+ * <code>int[] x = new int[2] { 3, 7 };</code><br>
+ * with the initializer list as inputs.
+ * <p>
+ * In terms of the notation used for {@link Operation}, an array creation of an 
+ * array of elements of type <i>e</i> and length <i>n</i> has a signature
```
mernst: It took me a couple of readings to parse this sentence.  Maybe add commas (even though they are not grammatical) or rephrase?

src/randoop/operation/ArrayCreation.java
```diff
@@ -54,30 +67,41 @@ public ArrayCreation(Class<?> elementType, int length) {
     // Set state variables.
     this.elementType = elementType;
     this.length = length;
+    this.outputTypeCached = Array.newInstance(elementType, 0).getClass();
   }
 
+  /**
```
mernst: I'm wondering why this class defines writeReplace (along with another class that defines readResolve) rather than this class just defining writeObject and readObject, which would keep all the code together in this file.

src/randoop/operation/ArrayCreation.java
```diff
@@ -54,30 +67,41 @@ public ArrayCreation(Class<?> elementType, int length) {
...
+   */
   public int getLength() {
     return this.length;
   }
 
   /**
-   * Extracts the input constraints for this ArrayDeclarationInfo
-   * @return list of input constraints
+   * {@inheritDoc}
+   * @return list of element types matching length of created array.
```
mernst: Is every element of the returned list identical?  (I think so.)

src/randoop/operation/ArrayCreation.java
```diff
@@ -120,45 +143,48 @@ public String toStringVerbose() {
...
         b.append(", ");
       
-      // In the short output format, statements like "int x = 3" are not added to a sequence; instead,
-      // the value (e.g. "3") is inserted directly added as arguments to method calls.
-      Operation statementCreatingVar = inputVars.get(i).getDeclaringStatement(); 
-      if (!GenInputsAbstract.long_format
-          && ExecutableSequence.canUseShortFormat(statementCreatingVar)) {
-        b.append(PrimitiveTypes.toCodeString(((NonreceiverTerm) statementCreatingVar).getValue()));
+      // In the short output format, statements like "int x = 3" are not added 
+      // to a sequence; instead, the value (e.g. "3") is inserted directly added
```
mernst: Is "inserted directly added" a typo?

src/randoop/operation/ArrayCreation.java
```diff
@@ -120,45 +143,48 @@ public String toStringVerbose() {
...
-        b.append(PrimitiveTypes.toCodeString(((NonreceiverTerm) statementCreatingVar).getValue()));
+      // In the short output format, statements like "int x = 3" are not added 
+      // to a sequence; instead, the value (e.g. "3") is inserted directly added 
+      // as arguments to method calls.
+      Statement statementCreatingVar = inputVars.get(i).getDeclaringStatement(); 
+      if (!GenInputsAbstract.long_format && 
+          statementCreatingVar.isPrimitiveInitialization() &&
+          !statementCreatingVar.isNullInitialization()) {
+        String shortForm = statementCreatingVar.getShortForm();
+        if (shortForm != null) {
```
mernst: This code looks wrong to me.  If shortForm is null, then the statement doesn't get output at all.  I think if shortForm is null, then the long form should be used instead.
(As usual, same comment elsewhere in the diffs.)

src/randoop/operation/ArrayCreation.java
```diff
@@ -185,29 +211,48 @@ public boolean equals(Object o) {
...
    */
-  public static Operation parse(String str) {
+  public static Operation parse(String str) throws OperationParseException {
     int openBr = str.indexOf('[');
     int closeBr = str.indexOf(']');
     String elementTypeStr = str.substring(0, openBr);
     String lengthStr = str.substring(openBr + 1, closeBr);
-    Class<?> elementType = Reflection.classForName(elementTypeStr);
+    try {
+    Class<?> elementType = TypeNames.getTypeForName(elementTypeStr);
```
mernst: I would personally restrict the scope of the try...catch form to just this one statement rather than all three, but it's not a disaster to include the two extra statements.

src/randoop/operation/ConstructorCall.java
```diff
@@ -7,68 +7,69 @@
...
 
+  /**
+   * Converts this object to a form that can be serialized.
+   * 
+   * @return serializable form of this object
+   * @see SerializableConstructorCall
+   */
   private Object writeReplace() throws ObjectStreamException {
-    return new SerializableConstructorCall(constructor);
+    return new SerializableConstructorCall(this.constructor);
```
mernst: Why is it needed to add "this."?  (Here & elsewhere.)

src/randoop/operation/ConstructorCall.java
```diff
@@ -149,32 +146,39 @@ public void appendCode(Variable varName, List<Variable> inputVars, StringBuilder
...
       // instead, the value (e.g. "3") is inserted directly added as
       // arguments to method calls.
-      Operation statementCreatingVar = inputVars.get(i).getDeclaringStatement(); 
-      if (!GenInputsAbstract.long_format
-          && ExecutableSequence.canUseShortFormat(statementCreatingVar)) {
-        b.append(PrimitiveTypes.toCodeString(((NonreceiverTerm) statementCreatingVar).getValue()));
+      Statement statementCreatingVar = inputVars.get(i).getDeclaringStatement(); 
+      String shortForm = statementCreatingVar.getShortForm();
+      if (!GenInputsAbstract.long_format && shortForm != null) {
+        b.append(shortForm);
```
mernst: Here the test seems right (but it's a touch inefficient to generate the short form whether or not it will be needed, so that is a negative).

src/randoop/operation/ConstructorCall.java
```diff
@@ -149,32 +146,39 @@ public void appendCode(Variable varName, List<Variable> inputVars, StringBuilder
...
-      return false;
-    if (this == o)
-      return true;
-    if (!(o instanceof ConstructorCall))
-      return false;
-    ConstructorCall other = (ConstructorCall) o;
-    if (!this.constructor.equals(other.constructor))
-      return false;
-    return true;
+    if (o instanceof ConstructorCall) {
```
mernst: Same comment as before about canonical way to write an equals() method implementation.

src/randoop/operation/ConstructorSignatures.java
```diff
@@ -0,0 +1,73 @@
+package randoop.operation;
+
+import java.lang.reflect.Constructor;
+
+/**
+ * ConstructorParser provides static methods to write as well as parse
```
mernst: Wrong class name, or line 9 of the file uses the wrong class name.

src/randoop/operation/ConstructorSignatures.java
```diff
@@ -0,0 +1,73 @@
...
+public class ConstructorSignatures {
+
+  /**
+   * Parses a constructor signature as produced by 
+   * {@link ConstructorSignatures#getSignatureString(Constructor)} 
+   * and returns the corresponding reflective {@link Constructor} object.
+   * 
+   * @param signature a string representing a constructor signature.
+   * @return reflective {@link Constructor} method corresponding to signature.
+   * @throws OperationParseException if signature parameter does not match
```
mernst: This doesn't look right.  For example, it might throw an AssertionException if the format is wrong and one of the assertions fails.

src/randoop/operation/ConstructorSignatures.java
```diff
@@ -0,0 +1,73 @@
...
+   * @return reflective {@link Constructor} method corresponding to signature.
+   * @throws OperationParseException if signature parameter does not match 
+   *         expected format.
+   */
+  public static Constructor<?> getConstructorForSignatureString(String signature) throws OperationParseException {
+    if (signature == null) {
+      throw new IllegalArgumentException("signature may not be null");
+    }
+    
+    int openPar = signature.indexOf('(');
```
mernst: Would matching against a single regular expression be easier to read than this sequence of character operations?

src/randoop/operation/ConstructorSignatures.java
```diff
@@ -0,0 +1,73 @@
...
+   * Parses a constructor signature as produced by 
+   * {@link ConstructorSignatures#getSignatureString(Constructor)} 
+   * and returns the corresponding reflective {@link Constructor} object.
+   * 
+   * @param signature a string representing a constructor signature.
+   * @return reflective {@link Constructor} method corresponding to signature.
+   * @throws OperationParseException if signature parameter does not match 
+   *         expected format.
+   */
+  public static Constructor<?> getConstructorForSignatureString(String signature) throws OperationParseException {
```
mernst: Most of this implementation is identically duplicated in getMethodForSignatureString, and it would be good to eliminate that duplication.

src/randoop/operation/DummyStatement.java
```diff
@@ -6,11 +6,16 @@
...
 import randoop.NormalExecution;
 import randoop.sequence.Variable;
 
-public class DummyStatement implements Operation, Serializable {
+/**
+ * DummyStatement is an {@link Operation} representing a noop.
+ * As an operation a DummyStatement with name noop is represented as 
+ * <i>noop</i> : [] &rarr; void, 
+ * an operation with no input types and void output. It does nothing when executed.
+ */
```
mernst: Can you remark on why is is useful or how it is used?
(This comment is already a big improvement over the lack of comments in the previous version of the code, of course.)

src/randoop/operation/DummyStatement.java
```diff
@@ -50,43 +55,60 @@ public String toStringVerbose() {
...
+  @Override
   public int hashCode() {
     return name.hashCode();
   }
 
+  /**
+   * {@inheritDoc}
+   * Generates a string is of the form:
+   * <code>
+   * (NAME)
```
mernst: I might prefix this by "dummy:", for explicitness.

src/randoop/operation/EnumConstant.java
```diff
@@ -9,18 +6,17 @@
...
-import randoop.util.Reflection;
+import randoop.types.TypeNames;
 
 /**
- * EnumConstant represents a constant value from an enum.
+ * EnumConstant is an {@link Operation} representing a constant value from an enum.
  * 
- * @author bjkeller
- *
+ * As a formal operation, a constant named BLUE from the enum Colors is an operation
```
mernst: What does "formal operation" mean in this context?

src/randoop/operation/FieldGetter.java
```diff
@@ -89,19 +86,12 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
   /**
    * appendCode adds the text for an initialization of a variable from a field to 
    * the StringBuilder.
-   * @param newVar - variable to be initialized.
    * @param inputVars - list of variables to be used (ignored).
```
mernst: There should not be a hyphen in an `@param` line.  Could you remove them here and anywhere else they appear in the code?

src/randoop/operation/FieldGetter.java
```diff
@@ -131,15 +121,15 @@ public boolean equals(Object obj) {
   public int hashCode() { return field.hashCode(); }
 
   /**
-   * parse recognizes a getter for a field in a string.
-   * A getter description has the form "<get>( field-descriptor )"
-   * where "<get>" is literal ("<" and ">" included, and field-descriptor
+   * Parses a getter for a field in a string.
```
mernst: I would say "from a string" rather than "in a string".

src/randoop/operation/FieldGetter.java
```diff
@@ -131,15 +121,15 @@ public boolean equals(Object obj) {
...
+   * A getter description has the form "&lt;get&gt;( field-descriptor )"
+   * where &lt;get&gt;" is literal ("&lt;" and "&gt;" included), and field-descriptor
    * is as recognized by {@link PublicFieldParser#parse(String)}.
+   * @see PublicFieldParser#parse(String)
    * 
-   * @param descr - string containing descriptor of getter for a field.
-   * @return - getter object in string.
+   * @param descr  the string containing descriptor of getter for a field.
+   * @return the getter object in string.
```
mernst: what does "in string" mean here?

src/randoop/operation/FieldSetter.java
```diff
@@ -99,17 +95,18 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
   }
 
   /**
-   * appendCode generates code for setting a field.
-   * Should look like
-   *   field = <value/variable>;
+   * Generates code for setting a field.
+   * Should look like<br>
+   *   <code>field = value;</code><br>
```
mernst: If you want code on its own line, `<pre>...</pre>` is more idiomatic than `<br><code>...</code><br>`.

src/randoop/operation/FieldSetter.java
```diff
@@ -99,17 +95,18 @@ public ExecutionOutcome execute(Object[] statementInput, PrintStream out) {
...
+   * or<br>
+   * <code>field = variable;</code>   
    * 
-   * @param newVar - variable that is ignored in this method.
    * @param inputVars - list of input variables. Last element is value to assign.
```
mernst: Same comment as above about hyphens in `@param` lines.

src/randoop/operation/FieldSetter.java
```diff
@@ -119,20 +116,23 @@ public void appendCode(Variable newVar, List<Variable> inputVars, StringBuilder
...
 
     //TODO this is duplicate code from RMethod - should factor out behavior
-    Operation statementCreatingVar = inputVars.get(index).getDeclaringStatement();
-    if (!GenInputsAbstract.long_format && ExecutableSequence.canUseShortFormat(statementCreatingVar )) {
-      Object val = ((NonreceiverTerm) statementCreatingVar).getValue();
-      b.append(PrimitiveTypes.toCodeString(val));
+    Statement statementCreatingVar = inputVars.get(index).getDeclaringStatement();
+    if (!GenInputsAbstract.long_format) {
+      String shortForm = statementCreatingVar.getShortForm();
+      if (shortForm != null) {
```
mernst: Same comment as before about silently dropping a statement with a null shortForm.

src/randoop/operation/MethodCall.java
```diff
@@ -150,26 +152,25 @@ public void appendCode(Variable newVar, List<Variable> inputVars, StringBuilder
...
-      if (!GenInputsAbstract.long_format
-          && ExecutableSequence.canUseShortFormat(statementCreatingVar)) {
-        Object val = ((NonreceiverTerm) statementCreatingVar).getValue();
-        sb.append(PrimitiveTypes.toCodeString(val));
+      // In the short output format, statements like "int x = 3" are not added 
+      // to a sequence; instead, the value (e.g. "3") is inserted directly added 
+      // as arguments to method calls.
+      Statement statementCreatingVar = inputVars.get(i).getDeclaringStatement();
+      String shortForm = statementCreatingVar.getShortForm();
+      if (!GenInputsAbstract.long_format && shortForm != null) {
```
mernst: Same comment as above about unnecessarily computing the shortForm.

src/randoop/operation/MethodCall.java
```diff
@@ -329,9 +346,10 @@ public boolean isVoid() {
   }
 
   /**
-   * Returns true if method represented by this MethodCallInfo is a static
-   * method.
+   * {@inheritDoc}
+   * @return true if this method is static, and false, otherwise.
```
mernst: I would drop the second comma.

src/randoop/operation/MethodCall.java
```diff
@@ -341,18 +359,80 @@ public boolean isStatic() {
...
   }
 
+  /**
+   * callsMethodIn determines whether the current method object
+   * calls one of the methods in the list.
+   * @param list method objects to compare against.
+   * @return true if method called by this object is in the given list.
+   */
+  public boolean callsMethodIn(List<Method> list) {
+    return list != null && list.contains(method);
```
mernst: Why does this test whether the list is null?  It seems that could mask errors, especially since the specification does not say that list is allowed to be null.  I don't see any clients that would actually pass null, but the test strongly suggests to me that they are allowed to do so, despite the lack of documentation about it.  I suggest you shorten the body to just `return list.contains(method);`

src/randoop/operation/NonreceiverTerm.java
```diff
@@ -284,14 +307,16 @@ public static NonreceiverTerm parse(String s) throws OperationParseException {
...
+      type = TypeNames.getTypeForName(typeString);
+    } catch (ClassNotFoundException e1) {
       String msg = "Error when parsing type/value pair " + s + ". A primitive value declaration description must be of the form "
-        + "<type>:<value>" + " but the <type> given (\"" + typeString + "\") was unrecognized.";
-      throw new OperationParseException(msg);
+          + "<type>:<value>" + " but the <type> given (\"" + typeString + "\") was unrecognized.";
+        throw new OperationParseException(msg);
     }
+    
     Object value = null;
```
mernst: Can you remove the initialization?  The rhs there suggests that value might not be set; leaving it off makes clear that value is always set.

src/randoop/operation/Operation.java
```diff
@@ -4,62 +4,166 @@
...
+ * The concept of an operation comes from logic and universal algebra, where 
+ * operations are used to build terms, which represent values. 
+ * (To dive into the formality, see, e.g., the 
+ * <a href="https://en.wikipedia.org/wiki/Term_algebra">Term Algebra</a> Wikipedia entry.)
+ * <p>
+ * An operation op has a type-signature 
+ * op: [T1, T2, ..., Tn] &rarr; T, where [T1, T2, ..., Tn] is the list of input 
+ * types, and T is the output type. 
+ * The input types are represented by an ordered list of 
+ * {@link Class} objects, and the output type is a single {@link Class} object.
```
mernst: It's true that the output type is a single Class object, but it's also the case that the operation may side-effect some of the inputs and thus may be viewed as having additional possible outputs.  Is that fact expressed?

src/randoop/operation/Operation.java
```diff
@@ -4,62 +4,166 @@
...
+ * <a href="https://en.wikipedia.org/wiki/Term_algebra">Term Algebra</a> Wikipedia entry.)
+ * <p>
+ * An operation op has a type-signature 
+ * op: [T1, T2, ..., Tn] &rarr; T, where [T1, T2, ..., Tn] is the list of input 
+ * types, and T is the output type. 
+ * The input types are represented by an ordered list of 
+ * {@link Class} objects, and the output type is a single {@link Class} object. 
+ * <p>
+ * For a non-static method call or instance field access, the first input type
+ * is always the class to which the method or field belongs. If we have a method
```
mernst: I would say "in which it is defined" and maybe also mention "receiver".  (I see that terminology is used below in the method javadoc.)  Your text is correct, but "belongs" isn't standard terminology.

src/randoop/operation/Operation.java
```diff
@@ -4,62 +4,166 @@
...
+   * 
+   * @return true if operation corresponds to static method or field of a class, 
+   *         and false, otherwise.
+   */
+  boolean isStatic();
+
+  /**
+   * Predicate to indicate whether object represents a method-call-like 
+   * operation (either static or instance). 
+   * This include non-method operations that access fields.
```
mernst: "include" => "includes".
I would be more specific:
"This includes field getters and setters."

src/randoop/operation/Operation.java
```diff
@@ -4,62 +4,166 @@
...
+   * 
+   * @return true if operation is method-like, and false otherwise.
+   */
+  boolean isMessage();
+
+  /**
+   * getDeclaringClass returns the type to which the operation belongs. 
+   * If a constructor or value, then the type should be the same as the output
```
mernst: That is, where it is defined, except for values where it is their type.  ("belongs" isn't a technical term with a well-known definition.)

src/randoop/operation/Operation.java
```diff
@@ -4,62 +4,166 @@
...
+  /**
+   * Predicate to indicate whether object represents a call to a constructor.
+   * 
+   * @return true if operation is a constructor call, and false otherwise.
+   */
+  boolean isConstructorCall();
+
+  /**
+   * Predicate to indicate whether this object represents a value of a
+   * non-receiving type (includes numbers, strings and null).
```
mernst: As a trivial point, it's good to use a serial comma in lists of 3 or more elements: https://homes.cs.washington.edu/~mernst/advice/write-technical-paper.html#serial-comma

src/randoop/reflection/ReflectionPredicate.java
```diff
@@ -11,10 +11,36 @@
  */
 public interface ReflectionPredicate {
 
+  /**
+   * Indicate whether a class meets the use criteria for predicate.
+   * 
+   * @param c  the class to test.
+   * @return true if the class meets the predicate criteria, false, otherwise.
+   */
   public boolean canUse(Class<?> c);
```
mernst: Calls to the canUse methods are often wrapped in a call named "satisfies".  Would it be better to use consistent naming and rename these methods as well?

src/randoop/operation/OperationParser.java
```diff
@@ -8,24 +8,28 @@
 
   /**
    * Parses a string representing a StatementKind. The string is expected to be
-   * of the form:
+   * of the form:<br>
```
mernst: In some places in this javadoc comment you added <br> tags and in other places <p> tags.  Is there a reason for the difference?

src/randoop/operation/PublicFieldParser.java
```diff
@@ -75,8 +74,10 @@ public PublicField parse(String s) throws OperationParseException {
...
 
 
-    Class<?> type = Reflection.classForName(typeName,true);
-    if (type == null) {
+    Class<?> type;
+    try {
+      type = TypeNames.getTypeForName(typeName);
+    } catch (ClassNotFoundException e) {
       String msg = errorPrefix + " The type given \"" + typeName +"\" was not recognized.";
       throw new OperationParseException(msg);
```
mernst: Does it make sense to chain the exception (pass `e` into the OperationParseException constructor)?

src/randoop/operation/SerializableConstructorCall.java
```diff
@@ -4,19 +4,24 @@
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
 
-import randoop.util.Reflection;
-
+/**
+ *  Serializable version of {@link ConstructorCall} allowing tests to be
+ *  serialized. In particular, uses string representation of the constructor based on
```
mernst: Thanks for this useful comment that explains what is different than the standard representation.

src/randoop/operation/TypeArgumentList.java
```diff
@@ -0,0 +1,66 @@
...
+ * <li><code>int</code>
+ * <li><code>int,double,java.lang.String</code>
+ * <li><code>randoop.operation.Operation</code>
+ * </ul>
+ * 
+ * @see MethodSignatures
+ * @see ConstructorSignatures
+ *
+ */
+public class TypeArgumentList {
```
mernst: This name suggests that the class represents a list.  I would call it "TypeArgumentListUtils" or "TypeArgumentUtils" or even "TypeArguments", to emphasize that it is a collection of static methods.

src/randoop/reflection/NotPrivateVisibilityPredicate.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * {@inheritDoc}
+   * @return true if the field access modifier is not private, and false, otherwise.
+   */
+  @Override
+  public boolean isVisible(Field f) {
+    return isVisible(f.getModifiers());
+  }
+  
+  private boolean isVisible(int mods) {
```
mernst: I would add Javadoc, as you have done for NotPrivateVisibilityPreficate.isVisible(int).

src/randoop/reflection/OperationExtractor.java
```diff
@@ -0,0 +1,134 @@
...
+ *
+ */
+public class OperationExtractor implements ClassVisitor {
+
+  private Set<Operation> operations;
+ 
+  /**
+   * OperationExtractor() creates a visitor object that collects Operation objects corresponding
+   * to class members visited by {@link ReflectionManager}. Stores {@link Operation} objects in
+   * ordered set to ensure strict order once flattened to list --- needed to guarantee determinism
```
mernst: missing "an" before "ordered".

"ordered set" is an oxymoron, and I think you mean "ordered collection".

src/randoop/reflection/OperationExtractor.java
```diff
@@ -0,0 +1,134 @@
...
+  public OperationExtractor() {
+    this.operations = new TreeSet<>();
+  }
+  
+  public List<Operation> getOperations() {
+    return new ArrayList<Operation>(operations);
+  }
+  
+  /**
+   * getStatements collects the methods, constructor and enum constants for a collection of classes.
```
mernst: minor: use serial comma

src/randoop/sequence/ForwardGenerator.java
```diff
@@ -579,18 +573,21 @@ private InputsAndSuccessFlag selectInputs(Operation statement) {
 
       // If we were unlucky and selected a null value as the receiver
       // for a method call, return with failure.
-      if (i == 0
-          && (statement instanceof MethodCall)
-          && (!((MethodCall) statement).isStatic())
-          && chosenSeq.getCreatingStatement(randomVariable) instanceof NonreceiverTerm)
+      if (i == 0 && 
+          operation.isMessage() &&
```
mernst: I personally find it easier to read a multi-line expression when the operators are at the beginning of the line.  (I agree the code doesn't line up quite as nicely with the operators at the beginning of the line, though.)
More importantly, the rest of the file puts the operators at the beginning of the line, and we should try to keep them consistent since inconsistent code is harder to read than either style.  Thanks!
(This applies a few other places in the diff, too.)

src/randoop/sequence/MappedSequences.java
```diff
@@ -59,6 +59,10 @@ public void addSequence(T key, Sequence seq) {
     emptyList = new ListOfLists<Sequence>(emptyJDKList);
   }
 
+  /**
+   * Returns all sequences by merging all of the sequence collections.
```
mernst: Does "merging" mean "unioning" or something else?

src/randoop/sequence/Sequence.java
```diff
@@ -658,36 +650,49 @@ protected void checkIndex(int i) {
...
    * statement appended at the end.
    */
-  public final Sequence extend(Operation statement, Variable... inputs) {
-    return extend(statement, Arrays.asList(inputs));
+  public final Sequence extend(Operation operation, Variable... inputs) {
+    return extend(operation, Arrays.asList(inputs));
   }
 
+  /**
+   * extend adds a new statement to a sequence using the operation of the given statement.
```
mernst: Better to say "to this sequence" rather than "to a sequence" which led me to look for a parameter to the method.

src/randoop/sequence/Sequence.java
```diff
@@ -839,9 +843,11 @@ public boolean isOwnerOf(Variable v) {
    * Returns a string representing this sequence. The string can be parsed back
    * into a sequence using the method Sequence.parse(String). In particular, the
    * following invariant holds:
-   *  <pre>st.equals(parse(st.toParseableCode()))</pre>
-   *  See the {@link randoop.Sequence#parse(List) parse} for the 
+   *  <code>st.equals(parse(st.toParseableCode()))</code>
```
mernst: I would have left it as `<pre>`, but if you are change it to `<code>` then add a period at the end of the sentence.

src/randoop/sequence/Sequence.java
```diff
@@ -1076,13 +1073,9 @@ public int lastUseBefore(int idx, Variable var) {
...
-    if (size() != 1) {
-      return false;
-    }
-    if (!(getOperation(0) instanceof NonreceiverTerm)) {
-      return false;
-    }
-    return true;
+    return (size() == 1 && getStatement(0).isPrimitiveInitialization());
   }
+
```
mernst: Minor: probably don't need these two added blank lines.

src/randoop/sequence/Statement.java
```diff
@@ -1,65 +1,92 @@
...
+ * 
  */
 public final class Statement implements Serializable {
 
   private static final long serialVersionUID = -6876369784900176443L;
 
-  // The kind of statement (method call, constructor call,
-  // primitive values declaration, etc.).
-  public final Operation statement;
+  // The operation (method call, constructor call, primitive values declaration, etc.).
```
mernst: Use /** ... */ Javadoc-style comment.

src/randoop/sequence/Statement.java
```diff
@@ -1,65 +1,92 @@
...
+   * @return true if output type is void.
    */
   public boolean isVoidMethodCall() {
-    return statement.getOutputType().equals(void.class);
+    return operation.getOutputType().equals(void.class);
   }
 
+  /**
+   * equals tests whether two Statement objects are equal:
+   * @return true if operation is the same, the number of inputs are the same, and inputs are equal.
```
mernst: "number is" rather than "number are"

src/randoop/sequence/Statement.java
```diff
@@ -67,11 +94,198 @@ public boolean equals(Object o) {
   @Override
   public int hashCode() {
```
mernst: It's not your fault, but this is a bizarre and non-optimal hash function.  I suggest using instead
`return java.util.Objects.hash(statement, inputs)` and also simplify any other undesirable hash functions.

src/randoop/sequence/Statement.java
```diff
@@ -67,11 +94,198 @@ public boolean equals(Object o) {
...
+      String typeName = type.getCanonicalName();
+      b.append(typeName);
+      b.append(" " + Variable.classToVariableName(type) + variable.index + " = ");
+    }
+    operation.appendCode(inputs, b);
+    b.append(";" + Globals.lineSep);
+  }
+
+  
+  public String toParseableString(String variableName, List<Variable> inputs) {
```
mernst: Please add Javadoc about how this is intended to be used, especially since it is not an overriding definition.

src/randoop/sequence/Statement.java
```diff
@@ -67,11 +94,198 @@ public boolean equals(Object o) {
...
+    b.append(" : ");
+    for (Variable v : inputs) {
+      b.append(v.toString());
+      b.append(" ");
+    }
+    return b.toString();
+  }
+
+  /**
+   * toModifiableStatement converts the statement to the mutable form.
```
mernst: Rather than converts, I would say it creates a new mutable version.

src/randoop/sequence/MutableStatement.java
```diff
@@ -13,7 +13,7 @@
  */
```
mernst: On line above, reference `MutableSequence`.  The MutableSequence class should indicate when one would choose to use each of the two types of sequence.  Also, its comment
<pre> * The "M" stands for "Mutable".</pre>
should be removed.

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
+package randoop.types;
+
+import randoop.util.PrimitiveTypes;
+
+import plume.UtilMDE;
+
+public class TypeNames {
+  
+  /**
+   * Generates a name for the given type in a format that can be compiled.
```
mernst: I would rephrase as "Returns a compilable Java name for the type"

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
...
+   * 
+   * @param cls is the type for which name is to be generated.
```
mernst: Don't need a period at end of `@param` and `@return` Javadoc clauses.

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
...
+    // If inner classes are involved, Class.getName() will return
+    // a string with "$" characters. To make it compilable, must replace with
+    // dots.
+    retval = retval.replace('$', '.');
+
+    return retval;
+  }
+  
+  /**
+   * Returns {@link Class} object for a fully qualified class name or primitive
```
mernst: "fully-qualified" has a hyphen.

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
+package randoop.types;
+
+import randoop.util.PrimitiveTypes;
+
+import plume.UtilMDE;
+
+public class TypeNames {
```
mernst: Please add Javadoc.  Also, should this be called "ClassNames" since that is what it manipulates?

src/randoop/types/TypeNames.java
```diff
@@ -0,0 +1,50 @@
...
+  
+  /**
+   * Generates a name for the given type in a format that can be compiled.
+   * 
+   * @param cls is the type for which name is to be generated.
+   * @return string that is a compilable version of type name.
+   */
+  public static String getCompilableName(Class<?> cls) {
```
mernst: This does not return a result in compilable form.  It looks like it returns it in some variant of JVML form, but I'm not sure what format it actually returns.  Please use a precise term such as "field descriptor", "binary name", "fully-qualified name", etc. rather than JVML format or compilable form.  A brief description of the four (!) formats defined by Java appears at http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#signature-annotations .
Also, is there a need to re-implement what exists in other packages?  For example, could you use plume-lib's `fieldDescriptorToBinaryName` (if that is what this is doing, but I now think it's not)?  Or should we add more methods that do the desired conversion, to some library?  (Maybe plume-lib should have a comprehensive ClassNames class that defines all possible conversions, or maybe some other library already does.)

src/randoop/util/ClassComplexityCalculator.java
```diff
@@ -125,13 +126,13 @@ public static void main(String[] args) throws ClassNotFoundException {
     Set<Class<?>> classes= new LinkedHashSet<Class<?>>();
     args= args[0].split(Globals.lineSep);
     int classCount= 0;
-    for (String className : args) {
-      String classNa = className.substring(0, className.length() - ".class".length());
-      System.out.println("loading " +classNa + " " + (classCount++) + " of " + args.length);
+    for (String arg : args) {
+      String className = arg.substring(0, arg.length() - ".class".length());
+      System.out.println("loading " +className + " " + (classCount++) + " of " + args.length);
```
mernst: As long as you are editing this line, please add space around the + operator.

src/randoop/util/ListOfLists.java
```diff
@@ -95,4 +98,44 @@ public T get(int index) {
...
+          elemIterator = listIterator.next().iterator();
+        }
+        return elemIterator.next();
+      }
+      throw new NoSuchElementException("end of ListOfLists reached");
+    }
+    
+    @Override
+    public void remove() {
+      //does nothing
```
mernst: If this has no effect, its body should be `throw new UnsupportedOperationException();`.

src/randoop/util/ListOfLists.java
```diff
@@ -95,4 +98,44 @@ public T get(int index) {
...
+    public LOLIterator(Iterator<SimpleList<T>> listIterator) {
+      this.listIterator = listIterator;
+      if (this.listIterator.hasNext()) {
+        this.elemIterator = (this.listIterator).next().iterator();
+      }
+    }
+
+    @Override
+    public boolean hasNext() {
+      return (listIterator.hasNext() || (elemIterator != null && elemIterator.hasNext()));
```
mernst: This implementation is not correct, because listIterator might contain (only) an empty iterator, in which case the LOLIterator doesn't acutally contain any elements and hasNext should return false.

src/randoop/util/ListOfLists.java
```diff
@@ -95,4 +98,44 @@ public T get(int index) {
   public String toString() {
     return toJDKList().toString();
   }
+
+  @Override
+  public Iterator<T> iterator() {
+    return new LOLIterator(lists.iterator());
+  }
+  
+  private class LOLIterator implements Iterator<T> {
```
mernst: Is LOL "list of lists"?  Please document.

But, couldn't we reuse functionality from some library instead of redefining it here?
For instance, plume-lib has MergedIterator that does the same as this, and other libraries must too.  And they won't contain the errors that I pointed out.

src/randoop/util/OneMoreElementList.java
```diff
@@ -45,4 +45,42 @@ public String toString() {
     return toJDKList().toString();
   }
 
+  @Override
+  public Iterator<T> iterator() {
+    return new OMEIterator(list.iterator(), lastElement);
+  }
+
+  private class OMEIterator implements Iterator<T> {
```
mernst: Stands for "one more element"?  Please document.  I'm inclined to implement this by calling something else (such as plume-lib's MergedIterator), even though it's less efficient, because that other code is probably better-tested.

src/randoop/util/OneMoreElementList.java
```diff
@@ -2,7 +2,9 @@
```
mernst: (I now see you didn't add this and other utility routines even though it is showing up as a new file in the diff.  The comments still stand.)

src/randoop/util/SimpleList.java
```diff
@@ -10,7 +10,8 @@
  * IMPLEMENTATION NOTE
  * <p>
  *
- * Randoop's main generator ({@link randoop.ForwardGenerator} creates new
+ * Randoop's main generator 
+ * ({@link randoop.sequence.ForwardGenerator ForwardGenerator}) creates new
  * sequences by concatenation existing sequences and appending a
```
mernst: Typo "concatenation"

src/randoop/util/SimpleList.java
```diff
@@ -40,11 +41,11 @@
  * creation time) proportional to N, not to the length of the new sequence.
  * <p>
  * When extending a Sequence with a new statement, we store the old sequence's
- * statements plus the new statement in a OneMoreElementList, which takes up
+ * statements plus the new statement in a {@code OneMoreElementList}, which takes up
  * only 2 references in memory (and constant creation time).
```
mernst: "only 2 references": well, there's the object header overhead too.  No need to make any change, though.

src/randoop/util/SimpleList.java
```diff
@@ -40,11 +41,11 @@
...
  * <p>
  */
-public abstract class SimpleList<T> {
+public abstract class SimpleList<T> implements Iterable<T> {
```
mernst: SimpleList is not descriptive of the point of this data structure.  (Like some other comments, you could fix this one after pushing the changes if you want, but let's not forget about them.)
The point of the list is that it is immutable and supports very efficient append and concatenation; the documentation should say that and maybe the name should reflect it as well.

src/randoop/operation/Operation.java
bjkeller: There is no representation that an operation might have side effects. Operations really are just formal constructs used to create statements, and the inputs/outputs are used solely for determining the types of values that are needed.

src/randoop/reflection/ReflectionPredicate.java
bjkeller: The code before used the name "filter" and said when we "can use" a reflective object.
I changed the terminology to predicates, and kept the method names.  I am open to changing them, but the calling methods are from Operation and correspond to saying
  "this operation satisfies the predicate"
while these methods are saying
  "this predicate is satisfied by the operation".

The Java 8 Predicate has the method test(T), which would be
  "this predicate tests true/false on the operation"

I'll go with the Java 8 convention.

bjkeller:
merged changes on st-refactor branch.

PR #68, opened 2015-11-25
Fix array variable name and final instance field omission

bjkeller:
This fixes two things:
1) the mismatch between variable names for arrays where declared variable had additional "_array".
2) the omission of final instance variables from the PublicField hierarchy. Final instance variables were being treated as mutable.

user doc and version numbers updated to v.2.0.1

### DIFF
--- a/src/randoop/operation/FinalInstanceField.java
+++ b/src/randoop/operation/FinalInstanceField.java
@@ -1,0 +1,34 @@
+package randoop.operation;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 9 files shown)


src/randoop/operation/FinalInstanceField.java
```diff
@@ -0,0 +1,34 @@
+package randoop.operation;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+import randoop.sequence.Variable;
+
+public class FinalInstanceField extends PublicField {
```
mernst: It would be delightful to add a spot of documentation for the class if possible.

PR #73, opened 2015-12-17
Fix NPE flakiness and update command-line

bjkeller:
This PR started out as a fix to the complaint about the behavior of no-NPE-contract under the flag --forbid-null, but also includes a major refactor of the command-line interface and the output of tests.  Specifically, regression and error-revealing tests are now output separately. The user also has finer control over what is output, and what is considered a failure. Also, removes command line arguments that were residue from previous research projects.
It also fixes an issue where bridge methods that restore the visibility of inherited public methods defined in a package private class were being filtered out of the operations model by the reflection predicate.


doc/CHANGES.txt
```diff
@@ -20,6 +20,17 @@ command-line arguments --no-regression-assertions and
...
+The option simplify-failed-tests has been removed. The command option exec has 
+been removed. The command line argument --inputlimit has been removed. Removed 
+--output-nonexec.
+
+The interface for ExecutionVisitor has been changed to visitAfterStatement and
+visitBeforeStatement to clarify that the methods are invoked relative to the
+execution of each statement. Note that an ExecutionVisitor may no longer modify
+the Checks for a Sequence.
+
+Support for the eclipse plugin has been removed. The option --comp-port has been
```
mernst: Capitalize "Eclipse"

src/randoop/DummyVisitor.java
```diff
@@ -9,13 +10,19 @@
...
     // do nothing.
   }
-  
-  public void visitBefore(ExecutableSequence sequence, int i) {
-    // do nothing.
+
+  @Override
+  public void visitBeforeStatement(ExecutableSequence sequence, int i) {
+    // do nothing
+
```
mernst: I wouldn't add this blank line.  I personally find it distracting, but more importantly it's inconsistent with code elsewhere in the file.  Same comment for other additions of blank lines.

src/randoop/EmptyExceptionCheck.java
```diff
@@ -0,0 +1,57 @@
...
+ * An {@code ExceptionCheck} that doesn't enforce the expectation of an
+ * exception at a particular statement. Allows for execution at the statement 
+ * to be either normal or throw an exception.
+ */
+public class EmptyExceptionCheck extends ExceptionCheck {
+  
+  private static final long serialVersionUID = 8159839339961663076L;
+  
+  /**
+   * Creates an exception check at the given statement.
```
mernst: More specifically, an EmptyExceptionCheck (or an empty exception check).
What is "the given statement"?  One isn't supplied, but `@param exception` mentions it too.  I would refer to statementIndex somehow.

src/randoop/EmptyExceptionCheck.java
```diff
@@ -0,0 +1,57 @@
...
+    super(exception, statementIndex);
+  }
+
+  /**
+   * {@inheritDoc}
+   * Appends comment indicating that catch is being ignored.
+   */
+  @Override
+  protected void appendCatchBehavior(StringBuilder b, String exceptionClassName) {
+    b.append("  // ignoring fact that caught expected exception " + exceptionClassName);
```
mernst: I find this a but too telegraphic; I suggest you use a few more words to make it clearer.

src/randoop/EmptyExceptionCheck.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  @Override
+  protected void appendTryBehavior(StringBuilder b, String exceptionClassName) {
+    b.append("  // ignoring fact that expected exception " + exceptionClassName + " not thrown");
+  }
+
+  /**
+   * {@inheritDoc}
+   * This check passes if execution is either normal or an exception is thrown.
+   * @return true if execution outcome is normal or and exception, false otherwise
```
mernst: "and" => "an"

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+ * An {@code ExceptionCheck} is used to indicate that an exception is expected
+ * at a particular statement in a sequence. Depending on command line arguments
+ * to Randoop, an instance may be either a {@link ExpectedExceptionCheck} or 
+ * {@link EmptyExceptionCheck}. Both will catch the exception if it occurs, but
+ * differ on whether the expectation of the exception is enforced.
+ */
+public abstract class ExceptionCheck implements Check {
+  
+ /**
+   *
```
mernst: Remove empty Javadoc comment.

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
+package randoop;
+
+import java.io.ObjectStreamException;
+import java.lang.reflect.Modifier;
+import java.util.Objects;
+
+/**
+ * An {@code ExceptionCheck} is used to indicate that an exception is expected
+ * at a particular statement in a sequence. Depending on command line arguments
```
mernst: Minor nit: "command-line" as an adjective contains a hyphen.  ("command line" as a noun does not.)

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+  // Indicates which statement results in the given exception. 
+  protected final int statementIndex;
+
+  public ExceptionCheck(Throwable exception, int statementIndex) {
+    this.exception = exception;
+    this.statementIndex = statementIndex;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
mernst: Is this intended to be overridden?  Because it seems like subclasses would not want to use it as is because it ignores any fields in the subclass.

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /**
+   * The "try" half of the try-catch wrapper.
+   */
+  @Override
+  public String toCodeStringPreStatement() {
+    StringBuilder b = new StringBuilder();
+    b.append("// The following exception was thrown during execution." + Globals.lineSep);
+    b.append("// This behavior will recorded for regression testing." + Globals.lineSep);
```
mernst: Missing word "be".  But do you really mean "will be" or "was"?

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    b.append("try {" + Globals.lineSep + "  ");
+    return b.toString();
+  }
+  
+  /** The nearest visible superclass -- usually the argument itself. */
+  // TODO: handle general visibility based on Randoop command-line
```
mernst: Is this related to the user-reported problem?  Are you deferring it to a later pull request, or is this worthy of examination now?

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+   * what code to include in the catch block.
+   */
+  @Override
+  public String toCodeStringPostStatement() {
+    StringBuilder b = new StringBuilder();
+    String exceptionClassName = getExceptionName();
+    if (exceptionClassName == null) {
+      exceptionClassName = "Exception";
+    }
+    appendTryBehavior(b, exceptionClassName);
```
mernst: If this is the catch half, and is post-statement, why is appendTryBehavior being called?

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    if (Modifier.isPublic(exception.getClass().getModifiers())) {
```
mernst: Why isn't appendCatchBehavior called in the then clause?  I expected the if...then...else to compute the type being caught, but then to re-merge so that appendCatchBehavior is always called.

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Appends code for catch block behavior corresponding to expected exception.
+   * 
+   * @param b  the string builder to which code text is to be added
+   * @param exceptionClassName  the class name of the expected exception
+   */
+  protected abstract void appendCatchBehavior(StringBuilder b, String exceptionClassName);
+
+  /**
+   * Appends code to follow statement throwing expected exception in try block.
```
mernst: It took me a little while to parse whether "follow statement" should be "following statement" or "follow the statement".  I think it's usually worth taking a few extra words to make the result unambigious.

src/randoop/ExecutionVisitor.java
```diff
@@ -39,4 +39,4 @@
    */
   void initialize(ExecutableSequence executableSequence);
 
-}
+}
```
mernst: Minor:  I would retain newlines at the end of files.

src/randoop/ExpectedExceptionCheck.java
```diff
@@ -1,138 +1,68 @@
...
 import randoop.sequence.Execution;
 import randoop.util.Reflection;
 
 /**
- * A checker that checks for an expected exception from a method call.
+ * An {@code ExceptionCheck} that enforces the expectation of an exception being
+ * thrown. In particular,
+ * <ul>
+ * <li> fails if exception is not thrown, and
+ * <li> succeeds only when expected exception is thrown.
```
mernst: Does this mean exactly, or is some subclass acceptable?

src/randoop/HelperSequenceCreator.java
```diff
@@ -38,7 +38,7 @@
     } else {
       SimpleList<Sequence> candidates = components.getSequencesForType(cls.getComponentType(), false);
       if (candidates.isEmpty()) {
-        if (GenInputsAbstract.forbid_null) {
+        if (GenInputsAbstract.null_ratio == 0) {
           // No sequences that produce appropriate component values found, and null forbidden.
```
mernst: The Randoop manual says of `--null-ratio`:  "If no non-null value is available, Randoop will try null regardless of the value of this parameter."  This code does not seem consistent with that statement.  (Do you think that Randoop's behavior should be different than what the manual says?)

src/randoop/NoExceptionCheck.java
```diff
@@ -58,7 +60,7 @@ public String toCodeStringPostStatement() {
 
   @Override
   public String toCodeStringPreStatement() {
-    return "";
+    return "//contract failure: this statement threw " + exceptionName + Globals.lineSep;
```
mernst: Please put a space between the comment characters "//" and the following text.

src/randoop/NoExceptionCheck.java
```diff
@@ -12,9 +12,11 @@
   
   // Indicates which statement is expected to return normally. 
   private final int statementIdx;
+  private String exceptionName;
   
-  public NoExceptionCheck(int statementIdx) {
+  public NoExceptionCheck(int statementIdx, String exceptionName) {
```
mernst: It's a bit surprising that the documentation says "Checks that no exception is thrown by a given statement" but the constructor takes a specific exception name as an argument.  Can you clarify the semantics or intended use?

src/randoop/main/GenInputsAbstract.java
```diff
@@ -175,14 +146,10 @@ public GenInputsAbstract(String command, String pitch,
...
    * Determines the maximum number of tests to output.
-   * 
-   * This command-line option is generally more appropriate than
-   * --inputlimit, which controls how many test candidates Randoop
-   * generates internally.
    */
   @Option ("Maximum number of tests to ouput; contrast to --inputlimit")
   public static int outputlimit = 100000000;
-
+
```
mernst: Very minor: avoid whitespace changes when possible.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -219,19 +184,6 @@ public GenInputsAbstract(String command, String pitch,
   @OptionGroup("Values used in tests")
   @Option("Use null as an input with the given frequency")
   public static double null_ratio = 0;
-  
-  /**
-   * Never use null as input to methods or constructors.
-   * 
-   * This option causes Randoop to abandon the method call rather than providing
```
mernst: As noted elsewhere, this isn't consistent with the Randoop manual.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -366,36 +318,42 @@ public GenInputsAbstract(String command, String pitch,
...
+  
+  /** The type of test to which a checked exception should be assigned. */
+  @Option("Type of test to which a checked exception is assigned")
+  public static ExceptionAllocation checked_exception = ExceptionAllocation.REGRESSION;
+  
+  /** The type of test to which an unchecked exception should be assigned. */
+  @Option("Type of test to which an unchecked exception is assigned")
+  public static ExceptionAllocation unchecked_exception = ExceptionAllocation.REGRESSION;
+  
+  /** The type of test to which a NullPointerException should be assigned */
```
mernst: I would expand this Javadoc, which will go in the manual.  It's not for all NullPointerExceptions.
The `@Option` text can still be short.
However, for all of those, rather than "type of test ... should be assigned" which might be confusing, can you reword to be something more like how the behavior is viewed?

src/randoop/main/GenInputsAbstract.java
```diff
@@ -366,36 +318,42 @@ public GenInputsAbstract(String command, String pitch,
...
+  /**
+   * The possible values for exception allocation command-line arguments.
+   */
+  public static enum ExceptionAllocation {
+    /** Occurrence of exception should be considered an error */
+    ERROR,
+    /** Occurrence of exception should be considered invalid */
+    INVALID,
+    /** Occurrence of exception should be considered expected behavior */
+    REGRESSION
```
mernst: Given the description just above, how about renaming this as EXPECTED?  That focuses more on the behavior rather than Randoop's mechanics of outputting tests.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -408,16 +366,6 @@ public GenInputsAbstract(String command, String pitch,
...
-   * Output sequences even if they do not complete execution.
-   * 
-   *  Randoop's default behavior is to output only tests consisting of
-   *  method call sequences that execute every statement, rather than throwing
-   *  an exception or failing a contract check before the last statement.
-   */
-  @Option("Output sequences even if they do not complete execution")
-  public static boolean output_nonexec = false;
-  
   @Option("specifies regex of classes that must be in any regression tests")
```
mernst: I know this isn't your fault, but I don't find this documentation descriptive.  Also, please capitalize the first letter of every `@Option` argument, for consistency.

src/randoop/main/Help.java
```diff
@@ -5,13 +5,13 @@
 import java.util.Arrays;
 import java.util.List;
 
+import randoop.Globals;
```
mernst: This reordering is weird to me.  It seems that before, the utility classes were first (JDK classes, then plume-lib).  Now what is the order?

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -69,7 +67,13 @@
    * Sequence limit for generation. If generation reaches the specified sequence
    * limit, the generator stops generating sequences.
    */
-  public final int maxSequences;
+  public final int maxGeneratedSequences;
+  
+  /**
+   * Limit for output. If more than specified number of sequences are in the
```
mernst: Should it be equal to rather than more than?  If I specify 500 as the max, I will be surprised to see 501 in the output.

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -113,41 +117,50 @@
...
+  public AbstractGenerator(List<Operation> operations, long timeMillis, 
+      int maxGenSequences, int maxOutSequences, ComponentManager componentManager,
+      IStopper stopper, RandoopListenerManager listenerManager) {
     assert operations != null;
 
     this.maxTimeMillis = timeMillis;
-
-    this.maxSequences = maxSequences;
-
+    this.maxGeneratedSequences = maxGenSequences;
```
mernst: Whenever setting fields in a constructor, please always use the exact same name for the formal parameter as for the field.  That's standard style.

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -69,7 +67,13 @@
    * Sequence limit for generation. If generation reaches the specified sequence
    * limit, the generator stops generating sequences.
    */
-  public final int maxSequences;
+  public final int maxGeneratedSequences;
```
mernst: How is this related to outputlimit?  Should you just use the same variable?

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -113,41 +117,50 @@
   public static Sequence currSeq = null;
   
   /**
-   * The list of final sequences that are printed out as JUnit tests (i.e. Randoop's output). 
+   * The list of failure sequences to be saved as JUnit tests
+   */
+  public List<ExecutableSequence> outFailureSeqs = new ArrayList<>();
```
mernst: The command-line arguments call these error tests.  Please use consistent terminology throughout the documentation and codebase.

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -113,41 +117,50 @@
...
-
+    this.maxGeneratedSequences = maxGenSequences;
+    this.maxOutputSequences = maxOutSequences;
     this.operations = operations;
-
-    this.executionVisitor = new MultiVisitor();
-
+    this.executionVisitor = new DummyVisitor();
+    this.outputTest = new AlwaysFalse<>();
+
```
mernst: Your editor seems to be leaving spaces on blank lines.  Can you configure it not to do so?

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -155,34 +168,74 @@ public AbstractGenerator(List<Operation> operations, long timeMillis, int maxSeq
...
+   */
+  public void addExecutionVisitor(ExecutionVisitor executionVisitor) {
+    if (executionVisitor == null) {
+      throw new IllegalArgumentException("executionVisitor must be non-null");
+    }
+    this.executionVisitor = executionVisitor;
+  }
+  
+  /**
+   * Registers a visitor with this object to generate checks following
```
mernst: I initially read this as "run the visitor after all generated tests have been executed".  But I think you mean that it is run after each generated test is executed.  Please clarify here and for other nearby methods' documentation.

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -155,34 +168,74 @@ public AbstractGenerator(List<Operation> operations, long timeMillis, int maxSeq
...
+  
   /**
-   * Creates and executes new sequences in a loop.
+   * Creates and executes new sequences until stopping criteria is met.
+   * @see AbstractGenerator#stop()
+   * @see AbstractGenerator#step()
    */
   public void explore() {
-
+    if (checkGenerator == null) {
```
mernst: The indentation looks wrong on these lines.

src/randoop/DummyVisitor.java
bjkeller: This is from an Eclipse template that has a ${body-template} substitution that doesn't have a definition for a void method.

src/randoop/ExceptionCheck.java
bjkeller: inheriting classes are pure behavior

src/randoop/ExceptionCheck.java
bjkeller: Only in the sense that it is using reflection to look at superclasses.  This method should probably be somewhere else, and the issue is that it is not using the VisibilityPredicate that has the policy from the command line.

src/randoop/ExceptionCheck.java
bjkeller: removed the condition on isPublic

src/randoop/HelperSequenceCreator.java
bjkeller: OK, I'll change the manual.

src/randoop/ExceptionCheck.java
mernst: Sorry, I don't understand your comment.

src/randoop/ExceptionCheck.java
mernst: When do you plan to address that issue?  Now, or in the future?

src/randoop/ExceptionCheck.java
bjkeller: I just meant there is no state in the implementing classes, they just
provide alternative behaviors based on type.

On Fri, Dec 18, 2015 at 9:23 AM, Michael Ernst <notifications@github.com>
wrote:

> In src/randoop/ExceptionCheck.java
> <https://github.com/randoop/randoop/pull/73#discussion_r48048156>:
>
> > +   *
> > +   */
> > +  private static final long serialVersionUID = 4806179088639914364L;
> > +
> > +  protected final Throwable exception;
> > +
> > +  // Indicates which statement results in the given exception.
> > +  protected final int statementIndex;
> > +
> > +  public ExceptionCheck(Throwable exception, int statementIndex) {
> > +    this.exception = exception;
> > +    this.statementIndex = statementIndex;
> > +  }
> > +
> > +  @Override
> > +  public boolean equals(Object o) {
>
> Sorry, I don't understand your comment.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/randoop/randoop/pull/73/files#r48048156>.
>



-- 
Ben Keller
bjkeller@uw.edu

src/randoop/main/Help.java
bjkeller: I noticed this earlier.  Apparently, I had put the order in the project settings and those are not in the repo.  I changed my global settings so that other packages are first, then plume, then randoop

src/randoop/sequence/AbstractGenerator.java
bjkeller: it is set from inputlimit via a parameter

src/randoop/HelperSequenceCreator.java
mernst: It seems that the behavior as specified in the manual, which is to pass a null value as a last resort, is desirable.  I wouldn't want Randoop to give up on creating sequences just because it is unwilling to pass a null value?

(My question "Do you think that Randoop's behavior should be different than what the manual says?" must have been ambiguous, sorry.  It was not intended to be, "Do you think inconsistency is good?", but "Do you disagree with the specification in the manual, and why?")

src/randoop/ExceptionCheck.java
mernst: In that case, I would add a brief comment so stating.
(PS: when replying to a GitHub pull request discussion by email, there is no need to quote the message.)

doc/CHANGES.txt
```diff
@@ -9,17 +9,37 @@ removed.  Use --error-test-filename and --regression-test-filename to set
...
+You can also control whether assertions appear in regression tests with the new
+command-line argument --no-regression-assertions; old command-line argument
 --check-regression-behavior has been removed.
 
-Old command-line argument --forbid-null has been removed; use
---null-ratio=0 instead.
+The command line --test-classes has been changed to --include-only-classes to
+distance meaning from --testclass argument. (Addresses #66)
+
+Fixes a bug in which certain inherited public methods were being excluded from
```
mernst: This sentence has no subject, making it hord to read.
Also, I would organize the changelog from most important and user-visible changes, to bug fixes, to non-user-visible changes (if they even appear).  A bug fix should appear after changes to command-line options.

doc/CHANGES.txt
```diff
@@ -9,17 +9,37 @@ removed.  Use --error-test-filename and --regression-test-filename to set
 the file names; old command-line argument --junit-classname has been
 removed.
 
-New command-line argument --no-npe-contract disables the "No null pointer
-exceptions without null" heuristic contract.
-
-You can control which assertions appear in regression tests with new
-command-line arguments --no-regression-assertions and
---regression-assertions-about-exceptions; old command-line argument
+You can now indicate whether you want to view exceptions as expected, error or
```
mernst: "Whether you want to view" is vague; I don't think users will find this helpful.  Please avoid first person and also use more specific language.
I would start by saying that Randoop now splits its generated tests into three categories, and explain what they are and how they are output into tests.  Then you can say that the user can customize the categorization of a generated test into those three categories.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -219,20 +184,18 @@ public GenInputsAbstract(String command, String pitch,
   @OptionGroup("Values used in tests")
   @Option("Use null as an input with the given frequency")
   public static double null_ratio = 0;
-  
+
   /**
-   * Never use null as input to methods or constructors.
+   * Suppress use of null when no other argument value can be generated.
```
mernst: I find this change makes the option harder to understand, especially in the manual (the confusing text there led me to this Javadoc comment).  The previous text was clear about use of null as an input to methods and constructors.  This uses "suppress", which does not appear elsewhere in the manual and is not defined or explained.
Adding details like "when no other argument value can be generated" is helpful, but not in concert with confusing changes.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -219,20 +184,18 @@ public GenInputsAbstract(String command, String pitch,
...
    * 
-   * This option causes Randoop to abandon the method call rather than providing
-   * null as an input, when no non-null value of the appropriate type is available.
+   * If true, Randoop will fail input generation rather than use null as an
```
mernst: What does "fail input generation" mean?  A user might imagine that Randoop crashes or has some other behavior.  Please make the text explicit so that users understand what is happening and the consequences of their choices, in words that are linked to concepts in the Randoop manual and that they can understand.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -219,20 +184,18 @@ public GenInputsAbstract(String command, String pitch,
...
+   * Suppress use of null when no other argument value can be generated.
    * 
-   * This option causes Randoop to abandon the method call rather than providing
-   * null as an input, when no non-null value of the appropriate type is available.
+   * If true, Randoop will fail input generation rather than use null as an 
+   * input.
    * 
-   * To ask Randoop to calls methods with null with greater frequency,
-   * see option --null-ratio.
+   * Does not affect the behavior based on {@link #null_ratio}.
```
mernst: Can you make the relationship between the two options more explicit?
Also, I see you removed the reference to --null-ratio.  I think you should reinstate it.  This text appears in the Randoop manual (many more people will read it there than in the source code), and it should make sense there using concepts that a Randoop user is familiar with.  Right now it appears in the manual as a confusing literal `#null_ratio`.
Other uses should be fixed too, such as the reference to --forbid-null from --null-ratio.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -349,53 +305,61 @@ public GenInputsAbstract(String command, String pitch,
...
+    INVALID,
+    /** Occurrence of exception should be considered expected behavior */
+    EXPECTED
   }
-
+  
   /**
-   * If a test throws an exception, should it be included in the regression test suite?
+   * Indicates whether a checked exception should be considered as error,
+   * expected, or invalid behavior. Determines how the exception is handled
```
mernst: "Determines how ..." feels like an implementation detail that I would omit.  This determines the classification of the test.  (Then it's the classfication of the test, not this, that determines how the exception is handled.)
Also, the text has become more obscure.  How can an exception be an error? The previous text was explicit:  it's the test that is being classified, not the exception, and what's relevant is that the test threw an exception.  So, please do improve the documentation, but don't take away from it.  I feel that you have changed this user-visible documentation to be terser and maybe more comprehensible to you, but that it will be confusing to users because it is less explicit and assumes they know much more about the internals of Randoop than they have any right to know.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -349,53 +305,61 @@ public GenInputsAbstract(String command, String pitch,
   public static boolean no_regression_assertions = false;
 
   /**
-   * The possible values of the regression_assertions_about_exceptions command-line argument.
-   * @see #regression_assertions_about_exceptions
+   * The possible values for exception allocation command-line arguments.
```
mernst: What is allocation?  I don't see this term elsewhere in the documentation.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -349,53 +305,61 @@ public GenInputsAbstract(String command, String pitch,
...
-  @Option("Which exceptional tests to include in the regression test suite")
-  public static ExceptionAssertionsMode regression_assertions_about_exceptions = ExceptionAssertionsMode.ALL;
-
-
-  // TODO make an enum. (But presently Options package requires upper-case
-  // strings for enums, which will require changes to Make targets, plugin, etc.)
-  /** For details, see the Javadoc documentation for {@link randoop.DefaultTestFilter}. */
-  @Option("What kinds of tests to output: pass, fail, or all")
-  public static String output_tests = "all";
+  @Option("Type of behavior assigned to a checked exception")
```
mernst: "Type of behavior" does not appear in the manual, and "behavior" is used many times in the manual, usually to mean "how the code behaves" (which includes the values it computes) rather than "classification of test outcome".  Can we find terminology that will be clearer to users?  (If we do want to reserve "behavior" for "test outcome", then we need to find a different name for the other uses.)

src/randoop/main/GenInputsAbstract.java
```diff
@@ -326,13 +289,6 @@ public GenInputsAbstract(String command, String pitch,
   @Option("Whether to output error-revealing tests")
   public static boolean no_error_revealing_tests = false;
```
mernst: Currently this appears in the option group "outputting the JUnit tests", which also has details like the file name for the test files.  I think that conceptual issues like which tests to output and how to treat exceptions should be separated from filename handling and implementation details like whether JUnit is used.  I suggest that the former should be made into a new option group that appears earlier, probably right before or right after the "limiting test generation"  option group.

doc/index.html
```diff
@@ -590,11 +572,7 @@ <h3 id="command-line-options">Command-line options</h3>
 
  Note that if you use this option, Randoop is nondeterministic: it
  may generate different test suites on different runs. [default 100]</li>
-      <li id="option:outputlimit"><b>--outputlimit=</b><i>int</i>. Determines the maximum number of tests to output.
- 
- This command-line option is generally more appropriate than
- --inputlimit, which controls how many test candidates Randoop
- generates internally. [default 100000000]</li>
+      <li id="option:outputlimit"><b>--outputlimit=</b><i>int</i>. Determines the maximum number of tests to output. [default 100000000]</li>
```
mernst: For consistency with following line (and to avoid excess verbiage), cut "Determines the".  Also, it's OK to remove the guidance about what option to use, but what replaces it?  Randoop has a lot of command-line options and a user may feel overwhelmed; how should a user decide which ones to use?

src/randoop/main/GenInputsAbstract.java
```diff
@@ -201,17 +167,16 @@ public GenInputsAbstract(String command, String pitch,
...
   /**
    * Use null with the given frequency as an argument to method calls.
    * 
    * For example, a null ratio of 0.05 directs Randoop to use
    * <code>null</code> as an input 5 percent of the time when a
-   * non-<code>null</code> value of the appropriate type is available.  If
-   * no non-<code>null</code> value is available, Randoop will try
-   * <code>null</code> regardless of the value of this parameter.
+   * non-<code>null</code> value of the appropriate type is available.
+   *
```
mernst: Note that line breaks introduced in the Javadoc code comments will not appear in either the HTML API documentation or in the manual unless you add an explicit `<p>`.  This may be intended only for the code comments, but it's a bit inconsistent and confusing to have line breaks that are considered important enough for the code comments but not for the formatted documentation.   I would be inclined to remove them for consistency.  (I realize that they predate your changes.)

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,157 @@
+package randoop;
+
+import java.util.Objects;
+
+/**
+ * An {@code ExceptionCheck} is used to indicate that an exception is expected
+ * at a particular statement in a sequence. Depending on command-line arguments
+ * to Randoop, an instance may be either a {@link ExpectedExceptionCheck} or 
+ * {@link EmptyExceptionCheck}. Both will catch the exception if it occurs, but
```
mernst: "Both will catch" makes it sound like the ExceptionCheck catches the exception.  I think you mean that code in the generated test will catch the exception, and that would be clearer to say.  Reading this comment, I'm not clear on whether that code is generated by ExceptionCheck or elsewhere, and clarifying that here would be helpful to people who want to understand the code.

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,157 @@
...
+  
+  // Indicates which statement results in the given exception. 
+  protected final int statementIndex;
+
+  private String catchClassName;
+
+  /**
+   * Creates an exception check for the statement at the statement index.
+   * The generated code for this check will include a try-catch block with
+   * behaviors determined by implementing sub-classes. The catch class name
```
mernst: I might drop this last sentence.  It's true but is a bit of an implementation detail and might be confusing to someone who is just trying to get an understanding of the class.  I had to read it a couple of times even though we just talked about this issue.

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,157 @@
...
+   */
+  public ExceptionCheck(Throwable exception, int statementIndex, String catchClassName) {
+    this.exception = exception;
+    this.statementIndex = statementIndex;
+    this.catchClassName = catchClassName;
+  }
+
+  /**
+   * Determines if two {@code ExceptionCheck} objects are equal.
+   * Assumes that implementing classes are purely behavior.
```
mernst: I would say "have no state" in addition to or instead of "are purely behavior".

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,157 @@
...
+   */
+  @Override
+  public boolean equals(Object o) {
+    if (o == null) return false;
+    if (o == this) return true;
+    if (this.getClass() != o.getClass()) { //match implementing class
```
mernst: minor: put space between "//" punctuation and following text.

src/randoop/ExceptionCheck.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * {@inheritDoc}
+   * The pre-statement prefix of the try-catch wrapper.
+   */
+  @Override
+  public final String toCodeStringPreStatement() {
+    StringBuilder b = new StringBuilder();
+    b.append("// The following exception was thrown during execution." + Globals.lineSep);
+    b.append("// This behavior will be recorded for regression testing." + Globals.lineSep);
```
mernst: "will be recorded": it seems like it already has been recorded.

src/randoop/ExpectedExceptionCheck.java
```diff
@@ -1,138 +1,76 @@
...
-import java.io.ObjectStreamException;
-import java.lang.reflect.Modifier;
-
 import randoop.sequence.Execution;
 import randoop.util.Reflection;
 
 /**
- * A checker that checks for an expected exception from a method call.
+ * An {@code ExceptionCheck} that enforces the expectation of an exception
+ * or a subclass being thrown. In particular,
```
mernst: For clarity, I would cut "or a subclass" here and move it three lines below.  That moves a detail out of the overview but makes the detailed explanation more accurate, so I think it improves them both.

src/randoop/NoExceptionCheck.java
```diff
@@ -3,8 +3,10 @@
 import randoop.sequence.Execution;
 
 /**
- * Checks that no exception is thrown after a given statement
- * in a sequence.
+ * Checks that an exception is not thrown by a given statement in a sequence.
```
mernst: "an exception is not thrown" could be interpreted as "no exception is thrown" or "some specific exception is not thrown".  Can you make the text explicit?

src/randoop/NoExceptionCheck.java
```diff
@@ -3,8 +3,10 @@
 import randoop.sequence.Execution;
 
 /**
- * Checks that no exception is thrown after a given statement
- * in a sequence.
+ * Checks that an exception is not thrown by a given statement in a sequence. 
+ * This is meant for an error-revealing check where, in fact, an exception
+ * was observed at the statement, and this check is meant to represent that
```
mernst: "was observed" is at test generation time, but this check will occur when the user runs the generated tests, right?  Clarifying what happens at different points in time would be helpful.

src/randoop/NoExceptionCheck.java
```diff
@@ -3,8 +3,10 @@
...
+ * the throws is actually an expected failure.
```
mernst: "expected failure" seems strange.  Also, "throws" sounds like a static code structure; do you mean the thrown exception?

After reading this a few times, I think it means:
 * The code under test is not supposed to throw an exception.  
 * The generated test checks for an exception and fails if one is thrown.
 * (Implementation detail that should be stated at the end:  This kind of test is generated when an exception was observed during test generation.)

I think it's most helpful to first state what something is or means, and then discuss typical uses.  Those uses can be illuminating, but mixing them together with the specification proper is confusing.

src/randoop/NoExceptionCheck.java
mernst: This comment doesn't seem to have been addressed.
If this check indicates that no exception should be thrown, then why is one exception name being specially used?

src/randoop/NoExceptionCheck.java
```diff
@@ -58,7 +62,7 @@ public String toCodeStringPostStatement() {
 
   @Override
   public String toCodeStringPreStatement() {
-    return "";
+    return "// contract failure: this statement threw " + exceptionName + Globals.lineSep;
```
mernst: Wouldn't it be a contract failure to throw *any* exception?
Or is this just about some specific exception and other exceptions are OK?

src/randoop/main/GenInputsAbstract.java
```diff
@@ -125,35 +124,6 @@ public GenInputsAbstract(String command, String pitch,
...
-   * The default set of contracts includes:
-   *   equals(Object) is reflexive,
-   *   equals(Object) is symmetric,
-   *   equals(Object) and hashCode() are consistent,
-   *   x.equals(null) returns false,
-   *   any nullary method annotated with {@code @CheckRep} returns true.
-   * </p>
-   */
-  ///////////////////////////////////////////////////////////////////
-  @OptionGroup("Creating test oracles")
```
mernst: Why did your remove this option group?  (Maybe it was on oversight.)  As noted elsewhere, I think that there should be an option group (even if not of this exact name) about which tests are generated and how behavior is classified.

src/randoop/main/GenInputsAbstract.java
mernst: This extraneous whitespace has not been removed.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -201,17 +167,16 @@ public GenInputsAbstract(String command, String pitch,
...
    * Use null with the given frequency as an argument to method calls.
    * 
    * For example, a null ratio of 0.05 directs Randoop to use
    * <code>null</code> as an input 5 percent of the time when a
-   * non-<code>null</code> value of the appropriate type is available.  If
-   * no non-<code>null</code> value is available, Randoop will try
-   * <code>null</code> regardless of the value of this parameter.
+   * non-<code>null</code> value of the appropriate type is available.
+   * 
+   * Unless {@link #forbid_null} is true, a null value will still be used if no
```
mernst: Refer to --forbid-null command-line option rather than using a Javadoc link.

src/randoop/main/GenInputsAbstract.java
```diff
@@ -409,17 +373,12 @@ public GenInputsAbstract(String command, String pitch,
...
+   * Indicate which classes that any generated test must include. Tests that
+   * do not have classes matching the regular expression are not written to 
+   * the test suites. 
    */
-  @Option("Output sequences even if they do not complete execution")
-  public static boolean output_nonexec = false;
-  
-  @Option("specifies regex of classes that must be in any regression tests")
-  public static Pattern test_classes = null;
+  @Option("Regular expression for class names that must occur in generated tests")
```
mernst: Looking at the code, it's not just that the generated test must include the given class names; if any variable has any class name that isn't in this list, the test is discarded.  That is, the condition seems stronger than stated here.  Can you please double-check and update the documentation?

src/randoop/main/GenTests.java
```diff
@@ -346,71 +344,41 @@ public boolean handle(String[] args) throws RandoopTextuiException {
     addClassLiterals(componentMgr, allClasses);
 
     RandoopListenerManager listenerMgr = new RandoopListenerManager();
-
-    AbstractGenerator explorer = null;
-
-    LinkedList<ITestFilter> outputTestFilters = new LinkedList<ITestFilter>();
-    outputTestFilters.add(new DefaultTestFilter());
+
```
mernst: Extraneous whitespace here.

src/randoop/main/GenTests.java
```diff
@@ -346,71 +344,41 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
-    // Determine what visitors to install.
-    // NOTE that order matters! Regression capture visitor
-    // should come after contract-violating visitor.
-
-    CodeCoverageTracker covTracker = new CodeCoverageTracker(covClasses);
-    listenerMgr.addListener(covTracker);
+    ////// setup for check generation 
+    TestCheckGenerator testGen = createTestCheckGenerator(visibility, classes);
+    
+    //Define test predicate to decide which test sequences will be output
```
mernst: I suggest leaving space after "//" before the text.

src/randoop/main/GenTests.java
```diff
@@ -346,71 +344,41 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+    Predicate<ExecutableSequence> outputTest = createTestOutputPredicate(objectConstructor);
```
mernst: This is a slightly funny name for a predicate.

src/randoop/main/GenTests.java
```diff
@@ -418,7 +386,9 @@ public boolean handle(String[] args) throws RandoopTextuiException {
           Class<ExecutionVisitor> cls =
             (Class<ExecutionVisitor>)Class.forName(visitorClsName);
           ExecutionVisitor vis = cls.newInstance();
-          visitors.add(vis);
+          if (vis != null) {
```
mernst: How can newInstance return null?  (That is, this test seems extraneous to me.)

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
         }
       }
     }
+    
+    ExecutionVisitor visitor;
+    if (visitors.isEmpty()) {
```
mernst: I find this idiom (requiring exactly one, non-null ExecutionVisitor) clumsy, and I wonder why not have a list (that could be empty or even null) rather than having to have exactly one.  That could make it possible to eliminate both DummyVisitor and MultiVisitor from Randoop.  Maybe that is something to change at a future date.  Or maybe this idiom is cleaner for some reason?

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+    explorer.addTestCheckGenerator(testGen);
+    explorer.addExecutionVisitor(visitor);
 
-    explorer.executionVisitor.visitors.addAll(visitors);
-
+    if (!GenInputsAbstract.noprogressdisplay) {
+      System.out.printf ("Explorer = %s\n", explorer);
+    }
+    
+    //Generate tests
```
mernst: Was this comment inaccurate or just unnecessary?  As someone not familiar with the Randoop codebase, I found it useful.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
-            sequences.get(i).getChecksResults() + "\n"
-            + seqsfromfile.get(i) + "\n"
-            + eseq.getChecksResults() + "\n" + sequences.get(i);
-
+  
+    if (! GenInputsAbstract.no_error_revealing_tests) {
+      List<ExecutableSequence> errorSequences = explorer.getErrorTestSequences();
+      if (errorSequences.size() > 0) {
+        if (! GenInputsAbstract.noprogressdisplay) {
+          System.out.format("%nError-revealing test output:%n");
```
mernst: System.out.format and System.out.printf are the same, I believe.  Randoop uses the latter more than the former; I would continue to use printf, for consistency.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+   * Builds the test predicate that determines whether a particular sequence will
+   * be included in the output based on command-line arguments.
+   * 
+   * @param objectConstructor  the constructor for the Object class
+   * @return the predicate 
+   */
+  private Predicate<ExecutableSequence> createTestOutputPredicate(ConstructorCall objectConstructor) {
+    Predicate<ExecutableSequence> outputTest;
+    if (GenInputsAbstract.dont_output_tests) {
+      outputTest = new AlwaysFalse<>();
```
mernst: Similar comments to above about predicate names; the variables look to me like they contain tests.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+   * @param objectConstructor  the constructor for the Object class
+   * @return the predicate 
+   */
+  private Predicate<ExecutableSequence> createTestOutputPredicate(ConstructorCall objectConstructor) {
+    Predicate<ExecutableSequence> outputTest;
+    if (GenInputsAbstract.dont_output_tests) {
+      outputTest = new AlwaysFalse<>();      
+    } else {
+      Predicate<ExecutableSequence> baseTest;
+      //base case: exclude sequences with just "new Object()", keep everything else
```
mernst: Why is this base case necessary?  This test will go in the pool, and then when it is extended Randoop will remove it as a subsumed test.  I would be inclined to just use that general mechanism rather than adding a seemingly-redundant special case here.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+      Set<Sequence> excludeSet = new LinkedHashSet<>();
+      excludeSet.add(newObj);
+      baseTest = new ExcludeTestPredicate(excludeSet);
+      if (GenInputsAbstract.include_only_classes != null) { //keep only tests with test classes
+        baseTest = baseTest.and(new IncludeTestPredicate(GenInputsAbstract.include_only_classes));
       }
-      if (!GenInputsAbstract.noprogressdisplay) {
-        System.out.printf("%d subsumed tests removed%n", sequences.size() - unique_seqs.size());
+
+      //Use arguments to determine which kinds of tests to output
```
mernst: I had assumed that test generation and test outputting would be distinct -- that is, the same tests would be generated and put into two lists, regardless of the outputting parameters.  Then the no_error_revealing_tests and no_regression_tests flags would be used when deciding whether to output the tests.  That would simplify the logic during generation.  (Might it help to ensure determinism too?  I'm not sure.)  But maybe this approach is better for some reason such as efficiency?

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
-      write_junit_tests ("./before_clean", sequences, null);
+      outputTest = baseTest.and(checkTest);
+    }
+    return outputTest;
+  }
 
-      write_sequences (sequences, tmpfile.getPath());
-      generate_clean_checks (tmpfile.getPath());
+  /**
+   * Outputs JUnit tests for the sequence list, and if user indicates, also,
```
mernst: I would split this comment into two sentences.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+  /**
+   * Outputs serialized version of test sequences using given string as a
+   * suffix for the file name.
+   * 
+   * @param sequences  the list of test sequences to serialize
+   * @param suffix  the string to use as a suffix
+   * @throws Error if an exception is thrown while writing to the file
+   */
+  private void outputSerializedTests(List<ExecutableSequence> sequences, 
+      String suffix) throws Error {
```
mernst: I would change the variable name suffix to fileSuffix or filenameSuffix

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+    System.out.println("Serializing tests...");
+    try {
 
-      remove_diff_checks (tmpfile.getPath());
-      sequences = read_sequences (tmpfile.getPath());
+      FileOutputStream fileos = new FileOutputStream(output_tests_serialized + suffix);
```
mernst: I would capitalize the "o", to make the variable name easier to read, here and on the next line.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+      ObjectOutputStream objectos = new ObjectOutputStream(new GZIPOutputStream(fileos));
+      objectos.writeObject(sequences);
+      objectos.close();
+      fileos.close();
+
+    } catch (Exception e) {
+      throw new Error(e);
```
mernst: Does it make sense to include the file name in the error message?

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+  /**
+   * Serializes component sequences used by the generator.
+   *  
+   * @param gen  the sequence generator
+   * @throws Error if an exception is thrown while writing to the file
+   */
+  private void outputComponentSequences(ForwardGenerator gen) throws Error {
+    // Output component sequences.
+    System.out.print("Serializing component sequences...");
+    try {
```
mernst: It seems that much of this method and the previous one can be extracted into a single method that takes an object and a filename and calls writeObject into the file, handling all the streams and exception catching.

src/randoop/main/GenTests.java
```diff
@@ -429,264 +399,335 @@ public boolean handle(String[] args) throws RandoopTextuiException {
...
+    contracts.add(new EqualsToNullRetFalse());
+    return contracts;
+  }
 
-
-    if (pluginBridge != null) {
-      IMessage msg = new ClosingStream();
-      pluginBridge.msgSender.send(msg);
+  /**
+   * Constructs an {@code ExceptionPredicate} that checks whether each kind of
```
mernst: From this documentation, I couldn't figure out what this does.  It sounded like it was verifying correctness of command-line arguments.  Is it creating an predicate that indicates exceptional behavor?  Are there two other similar predicates for normal behavior or regression behavior?

Would it be cleaner or less error-prone to have a single routine that, given a test, categorizes it into one of the three categories?
I'm concerned that having multiple tests that run independently could lead to incorrect results.  Suppose that exception E is supposed to be invalid, but all other unchecked exceptions are supposed to become regression tests.  Then the predicate for regression tests will return true for even exception E.  Since exceptions need to be checked in leaf-to-root-in-type-hierarchy order, I don't see how a separate predicate for each type of behavior can work.

src/randoop/main/Help.java
mernst: You did not yet re-re-order the imports for this .java file, though.  (When you do so, the comment will be removed from the view in GitHub.)

src/randoop/reflection/DefaultReflectionPredicate.java
```diff
@@ -58,6 +57,18 @@ public boolean test(Class<?> c) {
     return visibility.isVisible (c);
   }
 
+  /**
+   * {@inheritDoc}
+   * Does checks for the following cases:
+   * <ul>
+   * <li>Main methods
```
mernst: Thanks for documenting this case.  Is this a good policy?  Is it a problem to call main?

src/randoop/sequence/AbstractGenerator.java
mernst: To make sure you don't overlook any comments, review the comments at a URL like https://github.com/randoop/randoop/pull/73/files

src/randoop/main/GenInputsAbstract.java
bjkeller: so priority is manual over javadoc, yes?

mernst: Yes, please.  The rationale is that more people will read it, and that even though this is marginally less nice for the HTML API documentation, it's still better than having two copies of the documentation.
A feature for the Options doclet that creates the option listing in the manual would be to automatically change occurrences of `{@link #null_ratio}` to `--null-ratio`, but that doesn't seem like a high priority right now.

src/randoop/main/GenInputsAbstract.java
bjkeller: I changed these to say 

Whether checked exception is an ERROR, EXPECTED or INVALID

Remember you told me we'd take care of the manual after all this was done.

src/randoop/NoExceptionCheck.java
bjkeller: I've revised this.

Incidentally, there is no actual check for exceptions in error-revealing tests. Only a comment is output. If the test isn't flaky, the exception will be thrown when the error-revealing test is run.

I'm thinking we should revisit what error-revealing tests do, but not in this PR

src/randoop/NoExceptionCheck.java
bjkeller: The point of the comment is to provide information about what was observed during generation.

But, it will only be added because a particular exception that is tagged as an error behavior was thrown during test generation.  Any "expected" exceptions would be OK, but any invalid or error exceptions would not. We should probably wrap the statement in a try-catch.

src/randoop/main/GenInputsAbstract.java
bjkeller: because we removed the two options that were under it.
I'm not sure "oracles" makes sense anyway.

I did add a new group.

src/randoop/main/GenInputsAbstract.java
bjkeller: The predicate actually was only returning false if there were variables and none of them matched the pattern, but I changed it so that it also returns false when there are no variables.

But otherwise the description is correct. The predicate loops through the variables of a test sequence until it finds one that matches the regex, and only returns true if one is found.

src/randoop/main/GenTests.java
bjkeller: I guess I'm responsible for forcing the visitor to be non-null, but otherwise the single visitor within the test generator is as it was.

I'd be OK removing ExecutionVisitors altogether. Whether there is a good use case for them is not clear.

src/randoop/main/GenTests.java
bjkeller: ? I'm not seeing what you are referring to here. The "Generate tests" comment you linked to is still there.

src/randoop/main/GenTests.java
bjkeller: I'm not really sure.  All of this is from previous code.  The Object constructor call *is* specifically added at the beginning, so it is probably an artifact of that. 

It seems like if being able to generate Object is an issue, it should go into the "seeds" instead.

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -113,40 +117,49 @@
   public static Sequence currSeq = null;
   
   /**
-   * The list of final sequences that are printed out as JUnit tests (i.e. Randoop's output). 
+   * The list of failure sequences to be saved as JUnit tests
+   */
+  public List<ExecutableSequence> outErrorSeqs = new ArrayList<>();
```
mernst: Please use consistent terminology; here, the comment and the variable name use failure and error respectively.

src/randoop/sequence/AbstractGenerator.java
```diff
@@ -113,40 +117,49 @@
...
+
+  /**
+   * The list of regression sequences to be save as JUnit tests
```
mernst: "save" => "saved".  Also, here and above I would find "to be output" or "to be written" clearer than "to be saved" since "save" also has different meanings in computing, such as saving in a data structure.

src/randoop/sequence/ExecutableSequence.java
```diff
@@ -86,40 +86,44 @@
...
-  // Invariant: Invariant: sequence.size() == executionResults.size().
-  // Transient because it can contain arbitrary objects that may not be
-  // serializable.
+  /** The checks for this sequence */ 
+  private TestChecks checks;
+  
+  /**
+   * Contains the runtime objects created and exceptions thrown (if any)
+   * during execution of this sequence.
+   * Invariant: Invariant: sequence.size() == executionResults.size().
```
mernst: Duplicated "Invariant:"

src/randoop/sequence/ExecutableSequence.java
```diff
@@ -310,87 +280,74 @@ public void execute (ExecutionVisitor visitor) {
...
+    visitor.initialize(this);
 
+    //reset execution result values
+    hasNullInput = false;
     executionResults.theList.clear();
     for (int i = 0 ; i < sequence.size() ; i++) {
       executionResults.theList.add(NotExecuted.create());
     }
-
+
```
mernst: Extraneous whitespace

src/randoop/test/ContractCheckingVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+   */
+  private void checkBinary(ExecutableSequence s, ObjectContract c, Set<Integer> values, ErrorRevealingChecks checks) {
+    for (Integer i : values) {
+      for (Integer j : values) {
+
+        ExecutionOutcome result1 = s.getResult(i);
+        ExecutionOutcome result2 = s.getResult(j);
+        
+        if (! (result1 instanceof NormalExecution) 
+            && ! (result2 instanceof NormalExecution)) {
```
mernst: Why is this "and" rather than "or"?  (I would write it as "(! ( ... and))" rather than having two negations.)

src/randoop/NoExceptionCheck.java
bjkeller: I edited the class comment to say

> This check represents the fact that a statement should not throw an exception.
> It is used in an error-revealing test to indicate that an exception that is
> considered to be an error (e.g., not "expected" or "invalid") was thrown by 
> the statement during test generation. Only a comment is included when the test is
> output noting the occurrence of the exception during test generation, though
> the statement is expected to throw the method when the error-revealing test 
> is run.

I suppose I did change the intent of the class by adding the exception name, but my feeling is that if the point of an Error-Revealing test is to *reveal* then we should say what exception was observed during test generation.

src/randoop/sequence/AbstractGenerator.java
bjkeller: I had not changed this because it would be a large indentation change.

src/randoop/main/GenTests.java
bjkeller: I changed the text to try to make it more clear.

As far as the independent predicates is concerned, there is a problem, but it is opposite of what you are saying.  Exceptions are classified by checking for behaviors in the order: invalid, error, expected. So, the problem is if the more general case is invalid/error and the more specific is expected.  I added the checks for this case to the constructed predicate. The construction assumes the enum constant order is invalid < error < expected, and this is noted in comments in the enum.

src/randoop/reflection/DefaultReflectionPredicate.java
bjkeller: No, I don't see any reason not to call main. Though we might want to make sure that the timeout handling is better.

src/randoop/NoExceptionCheck.java
mernst: You are right that the comment provides information about what was 
observed during generation.  I was suggesting that the comment should say 
that, because I think the current wording is potentially confusing to 
someone reading the test (and likewise for other implementations of 
toCodeStringPreStatement).

src/randoop/main/GenInputsAbstract.java
mernst: The description is ambiguous:  I thought it meant that every class name 
must occur in every generated test, but you clarified that some class name 
must occur in every generated test.  I think that the documentation should 
be clarified.
(Also, this is only affects outputting tests, right?  Every generated test 
should be put in the pool even if it does not contain any of the given 
class names.)

src/randoop/main/GenTests.java
mernst: When you make a change on a line, you remove the code review conversation 
related to that line.  I believe you made a whitespace change on this 
line, so I don't know the referent for "what you are referring to here".  
Unfortunately, that means I can't respond to your comment.

src/randoop/main/GenTests.java
mernst: Your suggestion of putting "new Object()" in the seeds sounds reasonable.

                    -Mike

src/randoop/main/GenInputsAbstract.java
bjkeller: Yes, it only affects what goes into the output lists.

I am a little concerned that it is filtering tests that don't have any instances of a class --- since the only way that a variable would match a class name is if the value is an instance with that type. So, I think we wouldn't get any tests of purely static methods or fields.

mernst: That concern sounds right.  You could add that as low-priority item to 
your to-do list.

bjkeller: I already changed it so that it checks the declaring class of the operation
in each statement instead of matching against the variable names.

On Mon, Dec 28, 2015 at 1:30 PM, Michael Ernst <notifications@github.com>
wrote:

> In src/randoop/main/GenInputsAbstract.java
> <https://github.com/randoop/randoop/pull/73#discussion_r48506890>:
>
> >     */
> > -  @Option("Output sequences even if they do not complete execution")
> > -  public static boolean output_nonexec = false;
> > -
> > -  @Option("specifies regex of classes that must be in any regression tests")
> > -  public static Pattern test_classes = null;
> > +  @Option("Regular expression for class names that must occur in generated tests")
>
> That concern sounds right. You could add that as low-priority item to your
> to-do list.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/randoop/randoop/pull/73/files#r48506890>.
>



-- 
Ben Keller
bjkeller@uw.edu

PR #75, opened 2016-01-01
Package Visibility (aka, the Happy New Year PR)

bjkeller:
This PR includes a set of changes that allow Randoop to use package accessibility in collecting classes and class members. 

And, yes, the branch name is missing a 'b'.


src/randoop/reflection/PackageVisibilityPredicate.java
```diff
@@ -0,0 +1,88 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+
+/**
+ * A predicate that tests for visibility of a class, method, constructor, or
+ * field relative to a particular local package.
+ * For classes, tests if the class is either public or package private in
+ * the local package.
+ * Otherwise, implements a test for Java package accessibility rules, and
```
mernst: Parts of this class use the term "accessibility", and parts of it use "visibility".  Please use only one or the other, or explain the distinction between the two terms.

src/randoop/reflection/ReflectionManager.java
```diff
@@ -107,8 +107,28 @@ public void apply(Class<?> c) {
...
-
-
+  /**
+   * Applies the visitors to the methods of the class. 
+   * Will visit all methods unless visiting only public members.
+   * 
+   * @param c  the class whose methods should be visited
+   */
+  private void applyMethods(Class<?> c) {
+    Set<Method> methods = new HashSet<>();
```
mernst: This set is never side-effected, so it is always empty and the test `! methods.contains(m)` is always false.
I would find this method easier to understand if it was one loop, and that loop contained a single if test (albeit one with a slightly more elaborate test).  In particular, I couldn't tell from the class documentation how publicOnly and the predicate are combined:  is it an and-style test or an or-style test?  The class documentation could be made more explicit on this point as well.

src/randoop/reflection/PackageVisibilityPredicate.java
```diff
@@ -0,0 +1,95 @@
+package randoop.reflection;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+
+// TODO: Parts of this class use the term "accessibility", and parts of it
```
mernst: Please address this TODO

src/randoop/reflection/ReflectionManager.java
bjkeller: Thanks for spotting that.  I will check into what is going on with the tests.  My test should have caught that.

I don't understand what you are thinking about in terms of a single loop.  The method Class.getMethods() returns all of the public methods of the class, and the method Class.getDeclaredMethods() returns the methods declared within the class (not just public). There is no way to "get all methods", so to obtain public + not-private both methods have to be used.

bjkeller: The point of this set methods is to avoid visiting a method twice. The fact that it wasn't doing its job was hidden by the facts that visiting twice has no side-effect other than wasting time, and that the collected objects are placed in a set by the OperationExtractor visitor.

bjkeller:
I removed the public-members-only flag from ReflectionManager so that all visibility tests are handled through the reflection predicate.  I had originally introduced the flag for an optimization that is probably not needed, and it was just complicating the logic.  A public-only generation will unnecessarily dredge through non-private methods, but it only happens once and, unless the classes under test are really complex, the cost should be small.  Several of the test changes are due to the fact that this changed the interface of OperationExtractor.getOperations. 

The other big change is that I extended the VisibilityTest to include constructors, enums, and fields. This exposed a public-only bug in the field collection.

PR #83, opened 2016-02-05
Coverage agent

bjkeller:
Adds feature to filter output tests based on covering (exercising) a particular test class.


src/randoop/instrument/CoveredClassVisitor.java
```diff
@@ -0,0 +1,82 @@
...
+import randoop.sequence.ExecutableSequence;
+
+/**
+ * A {@link ExecutionVisitor} that polls a set of coverage instrumented
+ * classes and adds each covered class to an {@link ExecutableSequence} after
+ * it is executed.
+ */
+public class CoveredClassVisitor implements ExecutionVisitor {
+
+  /** the classes to be polled */
```
mernst: Minor:  Please capitalize the first letter.  I would add a period at the end too, though that isn't required if it's just a sentence fragment.

PR #84, opened 2016-02-09
Library updates

bjkeller:
This PR is centered around updating/simplifying the set of jar files included in randoop/lib. 
This includes removing the "Branch-directed Generation" option group along with related classes and test files. This removed code used Eclipse libraries that have been removed.
Other code changes mostly relate to:
1. updating agent and constant file class to BCEL 6, already included in plume.jar file
2. removing "long-format" command-line argument and code generation.


Makefile
```diff
@@ -39,7 +39,7 @@ JAVAC_TARGET ?= ${TARGET_DEFAULT}
 
 # User may set JAVAC_JAR
 # User may set JAVAC_EXTRA_ARGS
-JAVAC_COMMAND ?= ${JAVAC} ${JAVAC_TARGET} ${JAVAC_EXTRA_ARGS}
+JAVAC_COMMAND ?= ${JAVAC} ${JAVAC_TARGET} ${JAVAC_EXTRA_ARGS} -Xlint:-classfile
```
mernst: Can you comment on why you added -Xlint:-classfile?

Makefile
```diff
@@ -343,78 +341,11 @@ perf2: bin
...
-	   --experiment=test \
-	   --seqs-per-method=1 \
-	   --print-coderep-comments=true
-	gunzip frontier*.gz
-	cat frontier[1-9] \
-	  > systemtests/resources/arraylist.dfin.txt
-# Cleanup scratch files
-	rm -f frontier[1-9]* test.dftargets.txt
 
+### Note that this is not a target under tests
```
mernst: Does this get run regularly?  Should it?

Makefile
```diff
@@ -436,91 +367,11 @@ test-generation: bin
...
-# NOT A TEST! I use this target to communicate problems to Jeff.
-dferr%: $(DYNCOMP) bin
-	rm -rf systemtests/df-scratch
-	java -ea -classpath $(RANDOOP_HOME)/systemtests/src/java_collections:${JAVAC_JAR}:$(CLASSPATH) \
-	   randoop.main.DataFlow --debug_df \
-	   --scratchdir=systemtests/df-scratch \
-	   --overwrite \
-	   systemtests/resources/$@.txt
-
 execerr:
```
mernst: Execerr is not called by other targets.  Should it be tested/run regularly?

doc/dev.html
```diff
@@ -64,8 +64,26 @@
...
+Assuming this directory is <code>randoop-working</code>, get started by these steps:
+<code><pre>
+  cd randoop-working
+  git clone git@github.com:randoop/randoop.git myfeature
+  git clone git@github.com:randoop/randoop.git randoop-gh-pages
+  cd randoop-gh-pages
+  git checkout gh-pages
+</pre></code>
+You would then work in the directory <code>myfeature</code>, and follow the
+directions below to move changes into <code>gh-pages</code>.
```
mernst: I suggest putting two blank lines before a heading, to make it stand out visually when editing the .html file.

doc/dev.html
```diff
@@ -64,8 +64,26 @@
 
 <hr />
 
-<h1 id="tasks">Compiling, testing, releasing, etc.</h1>
+<h1 id="setup">Setting up</h1>
+
+Randoop development involves working with both Randoop's code base and the
+website and documentation.
+The first is is in the master (and other branches) of the repository, while
```
mernst: Is "and other branches" relevant here?  I would omit it, as it tends to suggest that a reader should start looking at all the other branches, when in fact they are transient.

doc/dev.html
```diff
@@ -64,8 +64,26 @@
...
+the website is in the gh-pages branch.
+These instructions assume that you keep subdirectories for both your working
+branch, and the gh-pages branch of Randoop in the same directory.
+Assuming this directory is <code>randoop-working</code>, get started by these steps:
+<code><pre>
+  cd randoop-working
+  git clone git@github.com:randoop/randoop.git myfeature
```
mernst: This seems like odd advice.  Why wouldn't I name my clone of randoop "randoop"?  That seems a lot better choice than "myfeature".

doc/dev.html
```diff
@@ -64,8 +64,26 @@
...
+
+Randoop development involves working with both Randoop's code base and the
+website and documentation.
+The first is is in the master (and other branches) of the repository, while
+the website is in the gh-pages branch.
+These instructions assume that you keep subdirectories for both your working
+branch, and the gh-pages branch of Randoop in the same directory.
+Assuming this directory is <code>randoop-working</code>, get started by these steps:
+<code><pre>
+  cd randoop-working
```
mernst: I find the name "randoop-working" for the parent directory misleading; I would think "randoop-working" would be a place I do work on randoop as opposed to a container directory.  Why not just omit any recommendation for a parent directory?  That's especially apt because (so far as I know) you aren't outlining a standard practice but something you just made up.

doc/dev.html
```diff
@@ -64,8 +64,26 @@
...
+  git clone git@github.com:randoop/randoop.git myfeature
+  git clone git@github.com:randoop/randoop.git randoop-gh-pages
+  cd randoop-gh-pages
+  git checkout gh-pages
+</pre></code>
+You would then work in the directory <code>myfeature</code>, and follow the
+directions below to move changes into <code>gh-pages</code>.
```
mernst: There is no reference to where to find "the directions below".  (It's in the release instructions, but that will not be obvious to readers, who are going to be confused while reading this.)  In fact, the developer doesn't have to even clone the gh-pages branch because the release instructions do so automatically if the gh-pages branch doesn't already exist.  Therefore, this entire section about setup seems redundant.  Developers will create a clone (and will probably name it `randoop`, though it doesn't matter to you whether they do), and then everything else happens automatically without need for the developer to do anything more.

lib/README
```diff
@@ -1,31 +1,18 @@
...
-org.eclipse.text_3.3.0.v20070606-0010.jar
-plume.jar
-servlet.jar
-swtgraphics2d.jar
+
+junit-4.12.jar:
+  http://search.maven.org/#search%7Cgav%7C1%7Cg:%22junit%22%20AND%20a:%22junit%22
+hamcrest-core-1.3.jar:
+  http://search.maven.org/#search%7Cgav%7C1%7Cg:%22org.hamcrest%22%20AND%20a:%22hamcrest-core%22
+
```
mernst: I would remove this blank line to emphasize that the comment refers just to these two.  But if the above URLs work, isn't that enough?

lib/README
```diff
@@ -1,31 +1,18 @@
...
+Or, use https://github.com/junit-team/junit/wiki/Download-and-Install
+  which has links to Maven Central Repository
+
+plume.jar:
+  Run "make lib/plume.jar" in Randoop
```
mernst: Is this really where it is coming from?  Is that what we want?  If a random developer does this, then we can get a version of plume.jar without type annotations.
Also, you need to indicate what version, such as what date plume-lib was pulled, so that someone reading this repository knows what version is checked in.
However, I strongly suggest that we use an official release from https://github.com/mernst/plume-lib/releases

I would think we would want a version with type annotations, but someone who runs the

mernst:
This pull request combines multiple different changes, making it more difficult to review.
In the future, please separate out unrelated changes like changing from BCEL 5.2 to BCEL 6 into their own pull requests -- or just commit it directly to master to keep it from cluttering the pull request.  (The pull request is so large that GitHub cannot show it and I'm having to use other tools to view it.)  Other changes, such as removing the --long-option command-line argument or the whitespace changes in doc/index.html, also don't need to be reviewed and should be commited directly to master to keep them out of the pull request.
The clutter makes it more likely that I will overlook important changes.  Although you made the changes over time, it forces me to attempt to keep in my mind at once, again making the review less effective.

src/randoop/main/Help.java
```diff
@@ -80,7 +80,7 @@ public boolean handle(String[] args_raw) throws RandoopTextuiException {
       // User wants help on a specific command.
       List<CommandHandler> allHandlers = new ArrayList<CommandHandler>();
       allHandlers.addAll(Main.handlers);
-      allHandlers.addAll(Main.invisibleHandlers);
+      // allHandlers.addAll(Main.invisibleHandlers);
```
mernst: All code related to invisibleHandlers is commented out.  Should it be removed instead?  (I guess my question is why it's kept around -- what do you plan to do with it later?)

src/randoop/operation/ConstructorCall.java
```diff
@@ -22,21 +21,21 @@
 import randoop.util.Util;
 
 /**
- * ConstructorCall is an {@link Operation} that represents a call to a
```
mernst: These whitespace changes are all good, but they bloat and clutter the diffs.
If your editor is configured to remove all whitespace from the ends of lines, then in order to avoid these undesirable whitespace-only diffs, could you remove all trailing whitespace from all files in Randoop and commit that to the repository once and for all?

src/randoop/operation/MethodCall.java
```diff
@@ -228,9 +235,8 @@ private void appendReceiverOrClassForStatics(String receiverString,
       Class<?> expectedType = getInputTypes().get(0);
       String canonicalName = expectedType.getCanonicalName();
       boolean mustCast = canonicalName != null
```
mernst: I personally like to parenthesize compound boolean expressions, especially if they span multiple lines.  I think it makes the code easier to read.

src/randoop/util/ClassFileConstants.java
```diff
@@ -41,7 +41,7 @@
 // the constant table.  Integer constants less that 64K are in the code.
 // There are also special opcodes to push values from -1 to 5.  This code
 // does not include them, but it would be easy to add them.  This code also
-// does not include class literals as constants.
+// does not include class literals as Const.
```
mernst: This line looks like mistaken string replacement.  (Your string replacement should have matched only "Constants" with a capital C, not "constants".)

mernst:
I wrote some scripts to help me manage the diffs and I think that I have made all the comments that I will.  Overall it looks good (modulo the need to tease apart all the different changes).

doc/dev.html
bjkeller: I removed this. I didn't understand that you were doing a checkout when I wrote it.

But I do think something is needed to orient the developer to how things work.   It is not unusual for dev documentation to give the clone command.  It would be appropriate to give the clone command and then explain that make distribution-files will dump a gh-pages directory into the parent.  It should also be explained that the command-line argument documentation comes from the code.

lib/README
bjkeller: This is built into the Makefile, and not my creation.

See targets plume-lib-update and lib/plume.jar.

doc/dev.html
mernst: I agree with the documentation suggestions you made.  Thanks!

lib/README
mernst: I agree it's not your fault.  However, I propose that we change it to use a real plume-lib release, which will reduce confusion.

bjkeller: So, remove the plume-lib repo, and its management from the Makefile. Yes?

On Wed, Feb 10, 2016 at 8:00 AM, Michael Ernst <notifications@github.com>
wrote:

> In lib/README
> <https://github.com/randoop/randoop/pull/84#discussion_r52477718>:
>
> > -org.eclipse.osgi_3.3.2.R33x_v20080105.jar
> > -org.eclipse.text_3.3.0.v20070606-0010.jar
> > -plume.jar
> > -servlet.jar
> > -swtgraphics2d.jar
> > +
> > +junit-4.12.jar:
> > +  http://search.maven.org/#search%7Cgav%7C1%7Cg:%22junit%22%20AND%20a:%22junit%22
> > +hamcrest-core-1.3.jar:
> > +  http://search.maven.org/#search%7Cgav%7C1%7Cg:%22org.hamcrest%22%20AND%20a:%22hamcrest-core%22
> > +
> > +Or, use https://github.com/junit-team/junit/wiki/Download-and-Install
> > +  which has links to Maven Central Repository
> > +
> > +plume.jar:
> > +  Run "make lib/plume.jar" in Randoop
>
> I agree it's not your fault. However, I propose that we change it to use a
> real plume-lib release, which will reduce confusion.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/randoop/randoop/pull/84/files#r52477718>.
>



-- 
Ben Keller
bjkeller@uw.edu

mernst: The scripts in plume-lib are needed; for example, see lines 630 and following in the top-level Makefile:

	utils/plume-lib/bin/html-update-toc doc/index.html

However, we can obtain the jar file from the release.

bjkeller: I'm told that to be able to include the annotated plume.jar without warnings, we also need to add checker-qual.jar to the lib.

bjkeller:
changes were made directly to master

lib/README
mernst: If true (and I think you are right), then this is a bug in plume-lib that should be corrected upstream.

PR #91, opened 2016-05-01
Utility classes should not have public constructors

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1118 - “Utility classes should not have public constructors ”. You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1118
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/src/main/java/randoop/JavaFileWriter.java
+++ b/src/main/java/randoop/JavaFileWriter.java
@@ -15,6 +15,9 @@ import randoop.util.Log;
  */
 public class JavaFileWriter {
 
+  private JavaFileWriter() {
+  }
+
   // Creates Junit tests for the faults.
   // Output is a set of .java files.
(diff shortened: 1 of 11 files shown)


mernst:
Thanks for the suggestions.

I'm not going to merge this pull request, because the patch degrades rather than improves code quality.

A method or constructor should never be added without documentation.  The added constructors have no indication of their purpose.

These classes should be examined to determine whether they should be instantiated at all.  If not, their private constructor should throw an error rather than having an empty body.  This is a worthwhile task that is inspired by your pull request.

Another problem with the patch is that it has formatting errors:
 * it makes some indentation incorrect
 * it re-indents commented lines gratuitiously

mernst:
I've created PR #103 for better fixes for these issues.

PR #99, opened 2016-05-22
Equals Transitive Contract

huangwaylon:
My additions for this fork are:
1. EqualsTransitive contract
2. Adding the contract to OperationModel
3. Updating ContractCheckingVisitor to accept contracts with three arguments
4. And adding to Buggy and buggyclasses.txt to get Randoop to generate Error tests using EqualsTransitive


src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -8,11 +8,7 @@
 import java.util.List;
 import java.util.Set;
 
-import randoop.contract.EqualsHashcode;
-import randoop.contract.EqualsReflexive;
-import randoop.contract.EqualsSymmetric;
-import randoop.contract.EqualsToNullRetFalse;
-import randoop.contract.ObjectContract;
+import randoop.contract.*;
```
bjkeller: Please change these back to explicit imports

src/systemtest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -113,6 +116,9 @@ public void runCollectionsTest() {
             options,
             timeout);
 
+    long endTime = System.currentTimeMillis();
+    System.out.println("runCollectionTest took: " + (endTime - startTime) + " milliseconds");
```
mernst: I don't think this should be printed unconditionally.  It's useful for debugging, but I suggest you comment it out.

src/main/java/randoop/contract/EqualsTransitive.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * The contract:
+ * <code>!(x0.equals(x1) &amp;&amp; x1.equals(x2))||x0.equals(x2)</code>.
+ */
+
+public class EqualsTransitive implements ObjectContract {
+    // Immutable, single instance
+    private static final EqualsTransitive instance = new EqualsTransitive();
+
+    // Private constructor
```
mernst: This comment seems obvious to me.  I would omit it.

In addition, you should use /** ... */ Javadoc-style comments even for private members.  The reasons are:
 * there is a Javadoc mode for generating HTML that includes private members
 * IDEs will display Javadoc for private members when you hover/click on a use

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -390,6 +391,7 @@ private void addOperations(Set<String> methodSignatures) throws OperationParseEx
    *   Note: if you add to this list, also update the Javadoc for check_object_contracts.
    */
   private void addDefaultContracts() {
+    contracts.add(EqualsTransitive.getInstance());
     contracts.add(new EqualsReflexive());
```
mernst: Should other contracts should be singletons, too?
If so, could you create a separate pull request that makes that change?

src/systemtest/java/randoop/main/RandoopSystemTest.java
huangwaylon: Would you suggest that I comment out the System.currentTimeMillis as well if it isn't being used after I comment out the println?

src/main/java/randoop/test/ContractCheckingVisitor.java
```diff
@@ -99,6 +101,71 @@ public TestChecks visit(ExecutableSequence s) {
...
+              // Create an check that records the actual value
+              // returned by the expression, marking it as invalid
+              // behavior.
+              checks.add(new ObjectCheck(c, i, s.sequence.getVariable(i), s.sequence.getVariable(j), s.sequence.getVariable(k)));
+            }
+          } else if (exprOutcome instanceof ExceptionalExecution) {
+            Throwable e = ((ExceptionalExecution) exprOutcome).getException();
+            if (Log.isLoggingOn()) Log.logLine("Contract threw exception: " + e.getMessage());
+            if (e instanceof BugInRandoopException) {
+              throw new BugInRandoopException(e);
```
mernst: Why are you wrapping a BugInRandoopException within another BugInRandoopException?  Why not just `throw e`?

src/main/java/randoop/test/ContractCheckingVisitor.java
```diff
@@ -99,6 +101,71 @@ public TestChecks visit(ExecutableSequence s) {
...
+            }
+            // Execution of contract resulted in exception. Do not create
+            // a contract-violation decoration.
```
mernst: The only user code that was executed was `equals`.  If `equals` fails, then that indicates a bug in the program.  I think that should be surfaced to the user and a failing test case created.

(This shouldn't happen because Randoop should have previously checked symmetry of equals.  However, the code should still do something in this case rather than silently ignoring it, because nondeterminism or side effects might have caused a problem.)

src/systemtest/java/randoop/main/RandoopSystemTest.java
mernst: Either way is OK with me.

src/main/java/randoop/test/ContractCheckingVisitor.java
huangwaylon: It looks like throw e won't compile because the compiler doesn't know whether it's a checked or unchecked exception. So instead, I will cast e:
`throw (BugInRandoopException) e`

src/main/java/randoop/test/ContractCheckingVisitor.java
huangwaylon: Are you referring to when `e.getRuntimeValue().equals(true)` evaluates to false?
I am not sure what else other than:
`if (Log.isLoggingOn())
                Log.logLine("Contract returned false. Will add ExpressionEqFalse check");
              // Create an check that records the actual value
              // returned by the expression, marking it as invalid
              // behavior.
              checks.add(new ObjectCheck(c, i, s.sequence.getVariable(i), s.sequence.getVariable(j), s.sequence.getVariable(k)));`

should be done.

mernst: I don't think this is  just when `e.getRuntimeValue().equals(true)` evaluates to false, but I'm not sure.

I'm trying to understand under what circumstances execution reaches line 156, which seems to be whenever an exception other than `BugInRandoopException` is thrown.  The code seems to silently ignore all such exceptions.  

Ignoring an exception is a code smell and I wonder what is the justification for ignoring the exception.

bjkeller: Randoop has not been doing anything with exceptions thrown by contracts
that are not Randoop errors.  This is probably because the contract is not
part of the test method sequence, and exceptions from statements in a
sequence are dealt with by a different visitor.





On Mon, May 30, 2016 at 8:23 PM, Michael Ernst <notifications@github.com>
wrote:

> In src/main/java/randoop/test/ContractCheckingVisitor.java
> <https://github.com/randoop/randoop/pull/99#discussion_r65119673>:
>
> > +            } else {
> > +              if (Log.isLoggingOn())
> > +                Log.logLine("Contract returned false. Will add ExpressionEqFalse check");
> > +              // Create an check that records the actual value
> > +              // returned by the expression, marking it as invalid
> > +              // behavior.
> > +              checks.add(new ObjectCheck(c, i, s.sequence.getVariable(i), s.sequence.getVariable(j), s.sequence.getVariable(k)));
> > +            }
> > +          } else if (exprOutcome instanceof ExceptionalExecution) {
> > +            Throwable e = ((ExceptionalExecution) exprOutcome).getException();
> > +            if (Log.isLoggingOn()) Log.logLine("Contract threw exception: " + e.getMessage());
> > +            if (e instanceof BugInRandoopException) {
> > +              throw new BugInRandoopException(e);
> > +            }
> > +            // Execution of contract resulted in exception. Do not create
> > +            // a contract-violation decoration.
>
> I don't think this is just when e.getRuntimeValue().equals(true)
> evaluates to false, but I'm not sure.
>
> I'm trying to understand under what circumstances execution reaches line
> 156, which seems to be whenever an exception other than
> BugInRandoopException is thrown. The code seems to silently ignore all
> such exceptions.
>
> Ignoring an exception is a code smell and I wonder what is the
> justification for ignoring the exception.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/randoop/randoop/pull/99/files/3e0cffc1ccd68c0fc501fae40637c3701e7fa66b#r65119673>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AFOdvXiW51zmtQFd6TvvmeHp7J1C8INhks5qG6nCgaJpZM4Ij6Nm>
> .
>



-- 
Ben Keller
bjkeller@uw.edu

src/main/java/randoop/test/ContractCheckingVisitor.java
```diff
@@ -99,6 +101,70 @@ public TestChecks visit(ExecutableSequence s) {
...
+              throw (BugInRandoopException) e;
+            }
+            // Execution of contract resulted in exception. Do not create
+            // a contract-violation decoration.
+            // TODO are there cases where exception in contract check is a
+            // failure?
+          } else {
+            throw new Error("Contract failed to execute during evaluation");
+          }
+        }
```
bjkeller: Let's make both checkBinary and checkTernary look more like checkUnary: if there is an exception then the check should be added.

doc/CHANGES.txt
```diff
@@ -1,10 +1,13 @@
-Version 2.1.5, released ????
+Version 3.0.0, released ????
 
 Randoop now tracks types independently of the Java reflection classes. This
 means for generic classes it will select a type argument for each type parameter
 and use that selection consistently. This will avoid compile and run time type
 errors related to inconsistent use of a parameterized type.
 
+Randoop implements the Object.equals() transitivity contract, which was
+previously documented by unimplemented.
```
mernst: "by" => "but"

src/main/java/randoop/contract/EqualsTransitive.java
```diff
@@ -0,0 +1,62 @@
+package randoop.contract;
+
+import randoop.Globals;
+
+/**
+ * The contract:
```
mernst: I would put some text about transitivity is this Javadoc comment.

src/main/java/randoop/contract/EqualsTransitive.java
```diff
@@ -0,0 +1,62 @@
...
+ * <code>!(x0.equals(x1) &amp;&amp; x1.equals(x2))||x0.equals(x2)</code>.
```
mernst: I would find the specification easier to read as an implication rather than when converted into more primitive operators (though the latter is how it would be implemented).

The spacing is confusing, with the || jammed against its operators even though it is the top-level operator.

src/main/java/randoop/contract/EqualsTransitive.java
huangwaylon: How about: "Checks the transitivity of equals for an object"

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -8,10 +8,11 @@
 import java.util.List;
 import java.util.Set;
 
-import randoop.contract.EqualsHashcode;
+import randoop.contract.EqualsTransitive;
 import randoop.contract.EqualsReflexive;
 import randoop.contract.EqualsSymmetric;
 import randoop.contract.EqualsToNullRetFalse;
+import randoop.contract.EqualsHashcode;
```
mernst: Why has this been moved later?  I would keep them in their current order to minimize diffs.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -390,6 +391,7 @@ private void addOperations(Set<String> methodSignatures) throws OperationParseEx
    *   Note: if you add to this list, also update the Javadoc for check_object_contracts.
    */
   private void addDefaultContracts() {
+    contracts.add(EqualsTransitive.getInstance());
```
mernst: Simpler contracts should appear before later ones, to make output more comprehensible.
That will be especially true once the implementation of transitivity assumes that reflexivity and symmetry have already been checked.

src/main/java/randoop/contract/EqualsTransitive.java
huangwaylon: I had initially written 
`<code>(x0.equals(x1) &amp;&amp; x1.equals(x2)) -> x0.equals(x2)</code>`
but this caused issues with the javadoc since -> wasn't a valid character. Is there a way to write - and > ?

src/main/java/randoop/test/ContractCheckingVisitor.java
```diff
@@ -99,6 +101,70 @@ public TestChecks visit(ExecutableSequence s) {
...
+
+          if (exprOutcome instanceof NormalExecution) {
+            NormalExecution e = (NormalExecution) exprOutcome;
+            if (e.getRuntimeValue().equals(true)) {
+              if (Log.isLoggingOn()) Log.logLine("Contract returned true.");
+              // Behavior ok.
+            } else {
+              if (Log.isLoggingOn())
+                Log.logLine("Contract returned false. Will add ExpressionEqFalse check");
+              // Create an check that records the actual value
```
mernst: "an check" => "a check"

src/main/java/randoop/test/ContractCheckingVisitor.java
```diff
@@ -99,6 +101,70 @@ public TestChecks visit(ExecutableSequence s) {
...
+              checks.add(new ObjectCheck(c, i, s.sequence.getVariable(i), s.sequence.getVariable(j), s.sequence.getVariable(k)));
+            }
+          } else if (exprOutcome instanceof ExceptionalExecution) {
+            Throwable e = ((ExceptionalExecution) exprOutcome).getException();
+            if (Log.isLoggingOn()) Log.logLine("Contract threw exception: " + e.getMessage());
+            if (e instanceof BugInRandoopException) {
+              throw (BugInRandoopException) e;
+            }
+            // Execution of contract resulted in exception. Do not create
+            // a contract-violation decoration.
```
mernst: I think a slightly more extensive explanation, taken from our discussion, would be useful.

src/main/java/randoop/contract/EqualsTransitive.java
bjkeller: You can use the HTML code for a right arrow: &#8594;
The way to literally get "->" is to use "-&gt;"

src/main/java/randoop/test/ContractCheckingVisitor.java
bjkeller: I think both checkBinary and checkTernary should be handling exceptions as in checkUnary, which includes the exceptions as failures, but has a mechanism to skip the inclusion depending on the contract. I'm not sure if we want to deal with this now.

src/main/java/randoop/contract/EqualsTransitive.java
mernst: I prefer `&rarr;` to `&#8594;` because it's clearer when you are reading.

src/main/java/randoop/contract/EqualsTransitive.java
mernst: Sure.  Please update the code.

PR #101, opened 2016-06-11
CompareTo Contracts and Singletons

huangwaylon:
I have created the following contracts over compareTo:
- CompareToReflexive - Checks the compare to is reflexive
- CompareToAntiSym - Checks antisymmetry of compareTo
- CompareToEquals - Checks consistency of compareTo with equals
- CompareToSubs - Checks substitutability of compareTo
- CompareToTransitive - Checks transitivity of compareTo

I have also created test input classes for the contracts listed above.

I made the following classes singletons:
- EqualsReflexive
- EqualsSymmetric
- EqualsHashcode
- EqualsToNullRetFalse

I also wrote CompareToReturnsNormally and EqualsReturnsNormally, and I updated HashCodeReturnsNormally so that it has the same design as the other returns to normally classes.
These three classes aren't included in operation model or aren't given a test input because toCodeString() method for each one invokes assertTrue("Contract failed", x0.compareTo()); or .equals(), or .hashCode(). However, we can't evaluate an int to a boolean.
A solution would be to have an assertNoException(...), but such a thing doesn't exist.


mernst:
Can you please pull master into your branch and resolve the ensuing merge conflicts?

src/main/java/randoop/contract/CompareToAntiSym.java
```diff
@@ -0,0 +1,67 @@
+package randoop.contract;
+
+import randoop.Globals;
+
+/**
+ * The contract: Reversing a compareTo on two objects should result in equal values of opposite signs
```
mernst: I suggest explictly using the term "antisymmetric" somewhere in the Javadoc, especially since an abbreviated version of that word is in the class name.

src/main/java/randoop/contract/CompareToAntiSym.java
```diff
@@ -0,0 +1,67 @@
...
+ * <code>Math.signum(x0.compareTo(x1)) == -Math.signum(x1.compareTo(x0))</code>.
+ */
+public class CompareToAntiSym implements ObjectContract {
```
mernst: Is there a reason not to write out the word rather than using the abbreviation "Sym"?

src/main/java/randoop/contract/CompareToAntiSym.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public int getArity() {
+        return 2;
+    }
+
+    @Override
+    public String toCommentString() {
+        return "compare-to-anti-symmetry on x0 and x1";
```
mernst: Please "symmetry" be "symmetric", for consistency with other names of contracts that are adjectives (such as "reflexive").

I would also use "compareTo" instead of "compare-to", since "compareTo" is the Java method being tested.

mernst: (Do we have a standard for use of adjectives vs. nouns in contract names.  Let's check them all and make sure the naming is consistent.)

src/main/java/randoop/contract/CompareToReflexive.java
```diff
@@ -0,0 +1,66 @@
+package randoop.contract;
+
+import randoop.Globals;
+
+/**
+ * The contract: Comparing an object with itself should result in 0
```
mernst: Add the word "reflexive" to the Javadoc.  Do the same for any other classes as appropriate.

src/main/java/randoop/contract/CompareToReflexive.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public int getArity() {
+        return 1;
+    }
+
+    @Override
+    public String toCommentString() {
+        return "compare-to-reflexive on x0";
```
mernst: Change "compare-to" to "compareTo".  Likewise anywhere else it appears.

src/main/java/randoop/contract/CompareToReturnsNormally.java
```diff
@@ -0,0 +1,66 @@
...
+        return instance;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public boolean evaluate(Object... objects) {
+        assert objects != null && objects.length == 1;
+        Object o = objects[0];
+        assert o != null;
+	    if (o instanceof Comparable) {
```
mernst: Please don't use tabs in source files.

src/main/java/randoop/contract/CompareToAntiSym.java
huangwaylon: Okay! I changed it to CompareToAntiSymmetric. I was worried that it might be "too long." Is there a general guideline as to when to use abbreviations then?

src/main/java/randoop/contract/CompareToSubs.java
```diff
@@ -0,0 +1,72 @@
+package randoop.contract;
+
+import randoop.Globals;
+
+/**
+ * The contract: Checks the substitutability property of compare to for equals
```
mernst: Change "compare to" to compareTo; the former is confusing because it doesn't clearly reference a method name and on first reading I didn't realize it meant exactly that method.

src/main/java/randoop/contract/CompareToAntiSym.java
mernst: Programmers won't have to write this very often, so I think it's OK.  You are right that it is a judgment call.

mernst:
When you have independent changes (such as changing contracts into singletons), It is easier and faster to review two pull requests, each of which performs a single operation.

I created pull request #102 for the change that makes contracts singletons.  Once that is reviewed and merged, then you can pull master into your fork and this pull request will become shorter and easier to review.

mernst:
The Travis job is failing.  Can you please fix that before we continue with the code review?  Thanks!

PR #108, opened 2016-06-18
Review of first half of developer manual

mernst:
Some comments that I did not address but that should perhaps be added:

In the "Randoop classpath" section:

"It is recommended that the classpath for classes under test be placed before
Plume-lib."
Please explain why, and more importantly give an example of how to achieve this.  Should the given text appear at the end of the classpath?

The alternative of using the "-all" jar file is not mentioned; wouldn't that be an easier modification of one's classpath?


bjkeller:
Having no mention of the fat-jar is an omission.

PR #113, opened 2016-08-11
Systemtest coverage

bjkeller:
Adds a coverage check to most system tests that were already compiling and running Randoop generated tests. Includes refactoring that moves capture and management of information from different phases of a system test into other classes. Documentation for writing tests updated, with a long comment added to RandoopSystemTest class with more detail.


doc/dev.html
```diff
@@ -535,17 +526,9 @@ <h4 id="system-tests">System Tests</h4>
...
-</p>
-<p>
-  The output of the Randoop run is captured in the <code>RandoopRunDescription</code>
-  object returned by <code>generateAndCompile</code>, and the output from running
-  the tests is captured in the <code>TestRunDescription</code> returned by <code>runTests</code>.
-  This output can be used in other assertions as needed.
+  Additional information can be found in the <code>RandoopSystemTest</code> class.
 </p>
 
-
```
mernst: I personally like to put a larger break before a section header (or at other places in a file where the topic changes).

src/main/java/randoop/operation/MethodCall.java
```diff
@@ -254,8 +254,6 @@ public static TypedClassOperation parse(String signature) throws OperationParseE
       throw new OperationParseException(msg);
     }
 
-    System.out.println("Looking for: " + opname + " args: " + arguments);
```
mernst: Should this be protected by an if statement that is enabled during debugging?

bjkeller: I would prefer to add lines as needed for particular debugging scenarios, and remove them when done. Otherwise, there would either be a lot of these messages to dredge through, or very complex mechanisms for enabling relevant groups of messages.

bjkeller:
closing this and doing merge directly

PR #140, opened 2016-12-15
Adding serialization support for ExecutableSequence class.

otmarjr:
- Added marker Serializable interface to:
  * CallableOperation
  * TypedOperation
  * Sequence
  * Statement
  * TypeTuple
  * IdentityMultiMap
  * RegressionChecks
  * ParameterBound
  * Type
  * TypeArgument
  
- Added custom serialization methods in classes:
  * ExecutableSequence
  * ConstructorCall
  * MethodCall

- Just made an unserializable field transient and no side-effects observed
  * ExplicitTypeVariable

- Unit test
  * SequenceTests : Method testSerialization added


bjkeller:
I'm a little concerned that your serialization won't work across executions. The reason is that classes loaded by different class loaders are distinct, so I would expect the deserialized Class<?> object for LinkedList to be different from LinkedList.class.  Have you tried it?

otmarjr:
No, I've not tried it. Is there any infrastructure built-in to test the usage of different class loaders ?

bjkeller:
Not exactly.  Though, perhaps you could see if it does what you need it to do.

bjkeller:
As I said before, my first question is whether it is doing what you need it to do.

Otherwise, there is no testing framework, but the basics are already used in the system tests.

For testing the scheme would be the following:
1. Write a JUnit test that creates sequences and serializes them. 
2. Write a JUnit test that deserializes the sequences from the first test and checks they are the *same*  as the sequences that were written (essentially this test would create the same sequences as the first test and check against the deserialized sequences).  
3. Extend the first test so that it runs the second test in a new JVM process (this would be similar to `randoop.main.RunStatus` in the system tests source set), capturing the process output, and ensuring that the tests pass. (It would not be necessary to use use the JaCoCo agent.)

I probably won't be able to get to any of this until January, but will be able to answer questions if you want to try yourself.

otmarjr:
Great to have this heads up, Ben. I'm not familiar with those concepts and I'd decided to postpone this adventure for tomorrow, when a fresh brain would definetely help :)

From the little research I've done, I tought a new thread with a different class loader would suffice. Thanks for putting me in the right direction!

bjkeller:
That may be true, but starting up a JVM running JUnit is simple.

otmarjr:
Well, before running all this workflow, which seems somewhat complex to me, I've created two separate unit tests, one to serialize to a file and another to deserialize from a file, intented to be executed in different runs of jUnit. This way, by having myself control the execution of these tests, I expect to reproduce the class loader issues you mentioned before. After fixing any serialization issues, I'll try to create an integration test the way you've described.

So far, it seemed to work reasonably well. The only differente is that the statement

`java.util.Set<java.lang.String> set_str5 = java.util.Collections.synchronizedSet((java.util.Set<java.lang.String>)treeset_str4);`

when deserialized is converted into the following statement:

`java.util.Set<java.lang.String> set_str5 = treeset_str4.synchronizedSet();`

To me, I cannot see the issues you mentioned with LinkedList. However, the call to `Collections.synchronizedSet` is missing. Do you believe this has any relation to the issue you've mentioned ?

otmarjr:
Nevermind, that was the isStatic field I'd forgotten to read back in MethodCall class. So, after fixing the method, I execute a unit test to write to a file and stop the execution. Thereafter, I execute the second unit test to parse the file and convert to an executable sequence. So far, the code string representation of the parsed file is equal to the executable sequence I've created.

bjkeller:
I think the most sophisticated scenario of use is that sequences are serialized in one run of Randoop, and then used to generate new sequences in subsequent runs. (Though, I'm pretty sure the options don't allow this at the moment). This would require that the types of the deserialized sequences be compatible with the types constructed within Randoop, and you cannot tell this by comparing the strings.

otmarjr:
Yeah, that is my very objective currently! I intend to run Randoop for long periods in my university's computer and circumvent the eventual power interruptions, which occur sometimes. In my current research I also would like to compare sequences of different sizes, therefore, I hope that reusing subsequences obtained from previous executions would greatly accelerate the generation process.

What should I compare to know that the deserialization worked ?

bjkeller:
The issue is with the type classes. In the second half of the test (run with a new class loader), you would both deserialize and recreate the sequence directly. This creates two sequences, one with types constructed by deserialization, and the other with types created directly in Randoop. Then iterate over the statements and test that the operations (or, minimally, the return types) are equal across the two.   The equals method should sufficient to detect if there are class loader differences.

otmarjr:
Now I've updated the tests to use the equals operation and I can confirm that it fails for the addFirst operation... I'll take a look and try to solve this.

bjkeller:
Yes, we would definitely want to know that the deserialization works across runs.

PR #147, opened 2017-01-18
Move test driver logic to test classes.

alcuadrado:
TestDriver was generating gigantic methods that won't compile.

My first attempt to solve this was to split the driver's method in one method per test class. This didn't work. Sometimes the driver's class is too big, and `javac` gives a `too many constants` error.

This commits creates a static `testAll` method in each test class, so the driver only has to invoke them.

This commits leaves `src/main/java/randoop/JunitFileWriter.java` with an invalid style. If someone helps me to find its errors I'll fix it.

### DIFF
--- a/src/main/java/randoop/JunitFileWriter.java
+++ b/src/main/java/randoop/JunitFileWriter.java
@@ -216,10 +216,17 @@ public class JunitFileWriter {
         writeFixture(out, AFTER_EACH, "", AFTER_EACH_METHOD, afterEachText);
       }
 
+      final List<String> methodNames = new ArrayList<>();
+
       for (ExecutableSequence s : sequences) {
-        writeTest(out, testClassName, methodNameGen.next(), s);
+        final String methodName = methodNameGen.next();
+        writeTest(out, testClassName, methodName, s);
+        methodNames.add(methodName);
         out.println();
       }
+
+      writeTestAllMethod(testClassName, out, methodNames);
+
       out.println("}");
       classMethodCounts.put(testClassName, methodNameGen.nameCount());
     } finally {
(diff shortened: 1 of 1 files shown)


bjkeller:
Patricio,

Thanks for the PR.  

If you run ./gradlew googleJavaFormat it should take care of the formatting issue.

Ben

alcuadrado:
Thanks Ben. I fixed it.

bjkeller:
Patricio,

I could use some help with tests that help show how this issue occurs, and prevents us from regressing in the future.  Would you be able to add a test case?

Ben

mernst:
Patricio said:
> TestDriver was generating gigantic methods that won't compile.

Ben said:
> Would you be able to add a test case?

Patricio, if you could provide a test case for this problem, that would help us resolve it.  I believe the problem is real, but you can surely reproduce it, whereas doing so would take us longer.

Thanks!

alcuadrado:
Hi, sorry for the delay.

I'm running Randoop like this: 

`java -ea -cp randoop-all-3.0.9.jar:examples/build/classes/main randoop.main.Main gentests --testclass=examples.unannotated.FiniteStack --timelimit=180 --junit-reflection-allowed=false`

Depending on the timelimit I set it may or may not trigger the bug. I'm running this on an oldish MacBook Air and 180 seems to be enough to reproduce it.

In this case, it generated 9 `RegressionTest<N>`, and a `RegressionTestDriver` whose `main` method is about 30k lines of code and fails to compile.

Here's `examples.unannotated.FiniteStack`'s source:

```java
package examples.unannotated;

public class FiniteStack<T> {

  private final Object[] data;

  private int size = 0;

  public FiniteStack(final int capacity) {

    if (capacity <= 0) {
      throw new IllegalArgumentException();
    }

    data = new Object[capacity];
  }

  public FiniteStack() {
    this(5);
  }

  public void Push(T item) {
    data[size++] = item;
  }

  public T Pop() {
    return (T) data[--size];
  }

  public int getSize() {
    return size;
  }
}
```

PR #164, opened 2017-03-17
JUnit test minimizer

mernst:


build.gradle
```diff
@@ -246,6 +249,9 @@ test {
   exclude 'randoop/test/Test_SomeDuplicates.*'
   exclude 'randoop/test/Test_SomePass.*'
   exclude 'randoop/operation/OperationParserTests.*'
+  exclude 'randoop/main/TestInput1.*'
```
mernst: Why are these in `main/` rather than in `minimizer/`?

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+      String filePath,
+      String classpath,
+      String packageName,
+      String expectedOutput,
+      int timeoutLimit) {
+    Results res = compileAndRun(filePath, classpath, packageName, timeoutLimit);
+
+    return res.compOut.exitValue == 0
+        && res.runOut != null
+        && res.runOut.errout.isEmpty()
```
mernst: The Minimizer seems to require that the standard output is empty.
Why does it do so?  I don't think we care about standard error when running the test, or standard output for that matter except for the JUnit errors that are output to it.
Isn't it OK for the test's output to change, so long as the failures are in the same location and with the same stack traces?
More generally, I would expect `normalizeJUnitOutput` to return a list of JUnit errors, with each one giving a full stack trace, but ignoring all output other than stack traces.
I haven't tried it myself, but you might be able to do this by running JUnit programmatically (see http://stackoverflow.com/questions/24943839/how-to-run-entire-junit-test-suite-from-within-code and other resources).

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+      return false;
+    }
+
+    System.out.println("Obtaining expected output.");
+
+    // Find the package name of the input file if it has one.
+    String packageName = null;
+    try {
+      PackageDeclaration classPackage = compUnit.getPackage();
+      if (classPackage != null) {
```
mernst: All of the test cases are in the default package.  Please add/move at least one in a package (preferably not even in a top-level package, but in a subpackage).

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+
+        // Simplify statement by removing the left hand side.
+        replacements.add(removeLeftHandSideSimplifiation(vdExpr));
+      }
+    }
+    return replacements;
+  }
+
+  /**
+   * If {@code currStmt} is a statement that is a passing assertion, store the value associated with
```
mernst: How does this routine test that the assertion is passing?  It just checks whether the statement is an assertion and assumes it is passing.  Probably update the documentation.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+
+        // Write, compile, and run the new Java file with the new suffix "Minimized".
+        String newFilePath = writeToFile(compUnit, filePath, "Minimized");
+        if (checkCorrectlyMinimized(
+            newFilePath, classpath, packageName, expectedOutput, timeoutLimit)) {
+          // No compilation or runtime issues, obtained output is the same as the expected output.
+          // Use simplification of this statement and continue with next statement.
+          replacementFound = true;
+          // Assertions are never simplified; if this is an assertion, it's the original statement.
+          storeValueFromAssertion(currStmt, primitiveValues);
```
mernst: `storeValueFromAssertion` is called from a location where I suspect that it never finds an assertion.  Please add a test case that shows the substitution happening.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+    if (currStmt instanceof ExpressionStmt) {
+      Expression exp = ((ExpressionStmt) currStmt).getExpression();
+      if (exp instanceof MethodCallExpr) {
+        MethodCallExpr mCall = (MethodCallExpr) exp;
+        // Check if the method call is an assertTrue statement
+        if (mCall.getName().equals("assertTrue")) {
+          // Is an assertTrue statement.
+          List<Expression> mArgs = mCall.getArgs();
+          if (mArgs.size() == 1) {
+            // Retrieve and store the value associated with the variable in the assertion.
```
mernst: storeValueFromAssertion assumes that the body of assertTrue is always of the form
  `assertTrue(x == 17)`
but it could be one of these:
  `assertTrue(x != 17)`
  `assertTrue(myList.isEmpty())`
Please have storeValueFromAssertion check the syntax rather than assuming it's an equality.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+
+  /**
+   * If {@code currStmt} is a statement that is a passing assertion, store the value associated with
+   * the variable in the {@code primitiveValues} map.
+   *
+   * @param currStmt a statement
+   * @param primitiveValues a map of variable names to variable values, modified if {@code currStmt}
+   *     is a passing assertion, asserting a variable's value.
+   */
+  private static void storeValueFromAssertion(
```
mernst: Once storeValueFromAssertion is working, please ensure that the minimizer handles variable reassignments.  Please add a test such as the following (but I haven't tested it to see if it triggers problems).

```
int i = 3
j = increment(i);
i = 7;
assertTrue(i == 7);
```

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+    List<VariableDeclarator> vars = exprCopy.getVars();
+    if (vars.size() > 1) {
+      // More than one variable declared in this expression.
+      return null;
+    }
+    VariableDeclarator vd = vars.get(0);
+
+    // Based on the declared variable type, assign the variable 0, false, or null.
+    Type type = exprCopy.getType();
+    if (type instanceof PrimitiveType) {
```
mernst: `rhsAssignZeroValue` and `rhsAssignValueFromPassingAssertion` have very similar bodies, here and elsewhere.  Abstract out the similar code into a method that both can call, which will make Minimize.java smaller overall.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+  private static int getFileLength(String filepath) {
+    int lines = 0;
+    try {
+      BufferedReader reader = new BufferedReader(new FileReader(filepath));
+      while (reader.readLine() != null) {
+        lines++;
+      }
+      reader.close();
+    } catch (FileNotFoundException e) {
+      System.err.println("File length not calculated, file not found exception.");
```
mernst: When outputting an error message about a file I/O problem, always give the name/path of the file.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+    } catch (ParseException e) {
+      // Unexpected error from parsing import.
+      System.err.println("Error parsing import");
+      return;
+    }
+
+    List<ImportDeclaration> importDeclarations = compilationUnit.getImports();
+    for (ImportDeclaration im : importDeclarations) {
+      // Check if the compilation unit already includes the import.
+      if (importDeclaration.toString().equals(im.toString())) {
```
mernst: It would be good to check for wildcards too.  Please add a test case that contains
  `import java.util.*;`
and make sure that you don't add a redundant
  `import java.util.List;`

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+        return;
+      }
+    }
+
+    // Add the import to the compilation unit's list of imports.
+    importDeclarations.add(importDeclaration);
```
mernst: It might be a nice touch to sort the imports alphabetically.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+
+    // Get the name of the file without full path
+    String fileName = Paths.get(filePath).getFileName().toString();
+    // Remove .java extension from file name
+    return fileName.substring(0, fileName.indexOf('.'));
+  }
+
+  /**
+   * Rename the overall class. Visit every class and interface declaration, of which we expect there
+   * to be one in the input test suite. The overall class will be renamed to class name + suffix.
```
mernst: Why is this restriction (only one class/interface declaration) needed?  If needed, it should be mentioned in the user-visible documentation.  However, I think you should be able to eliminate the restriction.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+      }
+    }
+  }
+
+  private static class TypeVisitor extends VoidVisitorAdapter<Object> {
+    /** @param arg a set of {@code Type} objects */
+    @SuppressWarnings("unchecked")
+    @Override
+    public void visit(ClassOrInterfaceType n, Object arg) {
+      Set<Type> params = (Set<Type>) arg;
```
mernst: What does this code do?  It seems to only side-effect a local variable and have no effect on the rest of the program.  If I am wrong, please document its purpose and how it achieves that, or fix it.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+  }
+
+  /**
+   * Given the absolute file path to a Java file, return the simple class name of the file.
+   *
+   * @param filePath absolute path to a Java file
+   * @return the simple class name
+   */
+  public static String getClassName(String filePath) {
+    if (filePath == null) {
```
mernst: What is the purpose of this code?  Can you please explain it?
I would think that if the argument is null, that is a bug in Randoop and it is better to fail than to mask it.

src/test/resources/minimizer/TestInput3.java.expected
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class TestInput3Minimized {
+
+    // Test simplification of right hand side values
+    @Test
+    public void test1() throws Throwable {
+        int d = 0;
```
mernst: The test cases TestInput1 and TestInput3 are a bit unfortunate.
They do not help a programmer detect the error.
For example, suppose the input was
```
  int x = sqrt(4);
  assert x == 2;
```
but someone optimized sqrt incorrectly so that now it produces 1.999, so the assertion failed.
The minimizer would change this to
```
  int x = 0;
  assert x == 2;
```
which fails for the same reason but has nothing to do with the original error.
This suggests that minimization might not be desirable, and perhaps should not be enabled by default.
Or maybe just that our purely dynamic approach to minimization is undesirable, and I was wrong to suggest it.
Or maybe this kind of result doesn't occur very often and so it isn't a concern in practice (though just in case, we would want Randoop to output both the original and the minimized test suite).

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+    // JUnit output starts with index 1 for first failure
+    int index = 1;
+    // String to represent the result that we return
+    String res = "";
+    // Continue until we finish processing the input String
+    while (scn.hasNextLine()) {
+      String line = scn.nextLine();
+      // Look for a String of the form "i)" where i is the current failing
+      // test index.
+      if (line.startsWith(index + ") ")) {
```
mernst: We want to ensure that the whole stack trace is retained and is the same, not just the very top of it.  How does this normalization do that?

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+  }
+
+  /**
+   * Normalize the standard output obtained from running a JUnit test suite. By normalizing the
+   * {@code String} representation of the output, we remove any extraneous information such as line
+   * numbers. The relevant information that is retained includes the indices of each failing test
+   * and the respective method names.
+   *
+   * @param input the {@code String} produced from running a JUnit test suite
+   * @return a {@code String} which contains several {@code String} pairs of output from the JUnit
```
mernst: What is the format?  I'm not sure how the pairs are represented within the big string.  It seems it would be cleaner to return a collection of pairs of strings, rather than a single string that encodes everything.

src/main/java/randoop/main/Minimize.java
huangwaylon: Added a test in a subpackage.

src/main/java/randoop/main/Minimize.java
huangwaylon: Added a test which uses a value from a passing assertion.

src/main/java/randoop/main/Minimize.java
huangwaylon: Updated documentation to state that the statement is just an assertion. It is then assumed to be passing since removing the assertion did not affect the failure stack trace.

src/main/java/randoop/main/Minimize.java
huangwaylon: Refactored the two methods and combined the similar bodies.

src/main/java/randoop/main/Minimize.java
huangwaylon: I tried the above approach, however there were various issues that arose. Running the JUnit test suite programmatically required the minimizer to pass in the `Class` object for the test suite. I attempted to use Class.forName(...) and ClassLoader.loadClass(...), but these both resulted in `ClassNotFoundExceptions`. After some searching, it looked like the issue was with the ClassLoader and not being able to dynamically load the new classes that were being generated by the minimizer. When I ran the minimizer the first time, an exception would occur (after the new file was written and compiled). When I then ran it a second time, there wouldn't be an exception since the file existed. However, the minimizer writes to this file many times, but the one that gets used is apparently the cached copy of the original. The new approach is to use the existing approach of running the JUnit test suite, but parsing the standard output to produce a map from method name to failure stack trace. This time I made sure to handle line numbers.

src/main/java/randoop/main/Minimize.java
huangwaylon: The minimizer now retains the whole stack trace. Previously I had thought that only the method name and name of the failure were necessary. I had also needed a way to handle the line numbers and the initial approach ignored all lines with line numbers.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+    Set<Type> fullyQualifiedNames = new HashSet<Type>();
+    // Collect all of the type names in the compilation unit.
+    new TypeVisitor().visit(compUnit, fullyQualifiedNames);
+
+    // Iterate through the set of fully qualified names and fill the map
+    // with mappings from fully qualified names to simple type names. Also
+    // add in necessary import statements.
+    for (Type type : fullyQualifiedNames) {
+      String typeName = type.toString();
+      addImport(compUnit, typeName);
```
huangwaylon: Added a test case resembling the scenario you described. The current approach tries to make the import. If doing so causes a compilation error, it then doesn't actually add the import and make the type name simplifications. Is this approach reasonable? Or should the minimizer attempt to import a.b and for c.d and use the type names b.MyClass and d.MyClass?

src/main/java/randoop/main/Minimize.java
huangwaylon: Added a test case with wildcard imports. Redundant imports aren't added anymore.

src/main/java/randoop/main/Minimize.java
huangwaylon: This is supposed to add onto a set containing all the types that appear within the Java file. So if the type `org.apache.commons.lang3.MutablePair` appears in the program, it will be added to the set of types. This is then used for type name substitutions to simplify `org.apache.commons.lang3.MutablePair` into `MutablePair` after adding an import statement.

src/main/java/randoop/main/Minimize.java
huangwaylon: The intent is that it only renames the main class. So if the input test suite is named TestSuite, it gets renamed to TestSuiteMinimized.

src/main/java/randoop/main/Minimize.java
huangwaylon: Added code to print an error message and exit.

src/main/java/randoop/main/Minimize.java
mernst: Please don't answer here in the pull request.  This conversation is not available to programmers who are reading the source code.  Please update the documentation instead.
Thanks!

src/main/java/randoop/main/Minimize.java
mernst: If you added this in the user-visible minimizer documentation, that's OK.  It would be good to detect when it is violated (or perhaps handle that case in the future).

src/main/java/randoop/main/Minimize.java
mernst: It's possible that an incorrect import will still compile -- for example, the two different classes of the same name might have the same interface, such as when one of them is a mock.

src/main/java/randoop/main/Minimize.java
huangwaylon: Sorting import statements now - this helped resolve the one error that was causing one test to fail in two of the three checks run by Travis.

src/main/java/randoop/main/Minimize.java
huangwaylon: The format for expected output (for an entire test suite) is now a map from method name to the method's failure stack trace (excluding line numbers).

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+    }
+
+    // Collect and return the results from the standard output and error output.
+    return new Outputs(stdOutputString, errOutputString, timeLimitProcess.exitValue());
+  }
+
+  /**
+   * Normalize the standard output obtained from running a JUnit test suite. By normalizing the
+   * {@code String} representation of the output, we remove any extraneous information such as line
+   * numbers. The relevant information that is retained includes the indices of each failing test
```
huangwaylon: Previously "handled" line numbers by ignoring all lines with line numbers. The minimizer now uses the whole stack trace and removes the line numbers from the trace.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1078 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Return a variable declaration statement that replaces the right hand side by 0, false, or null,
+   * whichever is type correct. The input variable declaration statement is assumed to declare only
+   * one variable, for instance, {@code int i = expr;}. Multiple variable declarations in a single
+   * statement in the form {@code int i = expr1, j = expr2;} are not valid.
```
huangwaylon: Adding checks in clients for null return values.

src/main/java/randoop/main/Minimize.java
huangwaylon: The minimizer now "checks" syntax by checking if the expression contains the string "==", but this doesn't seem like the ideal way since it seems like an expression with (a==b) != (c==d) could appear.

PR #170, opened 2017-03-20
Call avoidance

bjkeller:
PR for the purposes of reviewing documentation changes for planned modifications to mapcall agent.

Do not merge....


agent/mapcall/src/main/java/randoop/instrument/Premain.java
```diff
@@ -44,8 +44,10 @@
    */
   public static void premain(String agentArgs, Instrumentation inst) throws IOException {
 
-    System.out.format(
-        "In premain, agentargs ='%s', " + "Instrumentation = '%s'%n", agentArgs, inst);
+    if (verbose) {
```
mernst: This seems unrelated to the rest of the changes.  How about making this change on master instead?  Then it won't show up in the diffs.

build.gradle
```diff
@@ -256,13 +257,7 @@ test {
...
 }
 
-/*
- * Configuration of agentTest task.
- * Runs JUnit with exercised-class Java agent over all classes in agentTest
- * sourceSet.
- * Requires that exercised-class jar file copied into Randoop build/libs.
- */
-task agentTest(type: Test, dependsOn: 'copyJars') {
+task exercisedClassTest(type: Test, dependsOn: 'copyJars') {
```
mernst: Can you add brief documentation, as is given for other tasks?

docs/manual/index.html
```diff
@@ -1063,66 +1063,145 @@ <h3 id="map_calls">Avoiding calls to specific methods</h3>
...
 <p>
-Alternately, you can provide a different implementation of a method that will be used
-when running under Randoop.  Randoop transforms the code under test,
-replacing each call to a undesired method by a call to a new implementation
-that you provide.
+Alternately, you can use the <code>mapcall</code> agent to provide a different
+implementation of a method that will be used when running under Randoop.
+The mapcall Java agent transforms the code under test, replacing each call to a
+undesired method by a call to a new implementation.
+Default alternate implementations are provided for <code>System.exit()</code>
```
mernst: Please avoid passive voice.  It isn't clear what a user needs to do to achieve this effect.

docs/manual/index.html
```diff
@@ -1063,66 +1063,145 @@ <h3 id="map_calls">Avoiding calls to specific methods</h3>
 </p>
 
 <p>
-Alternately, you can provide a different implementation of a method that will be used
-when running under Randoop.  Randoop transforms the code under test,
-replacing each call to a undesired method by a call to a new implementation
-that you provide.
+Alternately, you can use the <code>mapcall</code> agent to provide a different
```
mernst: I would start out with the goal, rather than introducing jargon ("the mapcall agent") right at the very beginning of the description.

bjkeller:
I'm closing this so we can switch to call-avoidance-docs-only branch

PR #171, opened 2017-03-25
JUnit Test Minimizer

huangwaylon:
Minimizer to minimize failing JUnit test suites.


src/test/java/randoop/main/minimizer/MinimizerTests.java
```diff
@@ -0,0 +1,240 @@
...
+        new StringBuilder(inputFilePath)
+            .insert(inputFilePath.lastIndexOf('.'), "Minimized")
+            .toString();
+    String expectedFilePath = inputFilePath + ".expected";
+
+    // Obtain File object references
+    File inputFile = new File(inputFilePath);
+    File outputFile = new File(outputFilePath);
+
+    // Obtain the complete path to the input, output, and expected files
```
mernst: Why is this step necessary?
More generally, there seems to be a lot of string manipulation and conversion between strings and Files.  Can it be simplified?

src/test/java/randoop/main/minimizer/MinimizerTests.java
```diff
@@ -0,0 +1,196 @@
...
+
+  @Test
+  public void test1() throws IOException {
+    testWithInput("TestInput1.java");
+  }
+
+  @Test
+  public void test2() throws IOException {
+    testWithInput("TestInput2.java");
+    // Test input contains while loop.
```
mernst: Comments like these may be better in the files themselves, where they will be seen by someone who is debugging test failures.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+    // used in a simplification later on.
+    if (currStmt instanceof ExpressionStmt) {
+      Expression exp = ((ExpressionStmt) currStmt).getExpression();
+      if (exp instanceof MethodCallExpr) {
+        MethodCallExpr mCall = (MethodCallExpr) exp;
+        // Check if the method call is an assertTrue statement
+        if (mCall.getName().equals("assertTrue")) {
+          // Is an assertTrue statement.
+          List<Expression> mArgs = mCall.getArgs();
+          if (mArgs.size() == 1) {
```
mernst: This does not accommodate the [two-argument version of assertTrue](http://junit.sourceforge.net/javadoc/org/junit/Assert.html).

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+            // Check that the expression only has two nodes.
+            if (children.size() != 2) {
+              return;
+            }
+            String mExpStr = mExp.toString();
+
+            // Take the substring without the first and second expression nodes.
+            String firstExp = children.get(0).toString();
+            int secondExpIndex = mExpStr.lastIndexOf(children.get(1).toString());
+            mExpStr = mExpStr.substring(firstExp.length(), secondExpIndex);
```
mernst: String manipulations are ugly and error-prone.  Please get the operator directly from the AST rather than doing String manipulation.  It's a bit more work to understand the API well enough to use it, but worth it in terms of robustness (and efficiency).

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+    if (type instanceof PrimitiveType) {
+      switch (((PrimitiveType) type).getType()) {
+        case Boolean:
+          return rhsAssignWithValue(vdExpr, "false");
+        case Char:
+          return rhsAssignWithValue(vdExpr, "");
+        default:
+          return rhsAssignWithValue(vdExpr, "0");
+      }
+    }
```
mernst: For wrapped primitives such as `Integer`, try the corresponding zero value as well.
That means that this routine will return a list that may be of length 0, 1, or 2.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+          break;
+        case Double:
+          vd.setInit(new DoubleLiteralExpr(value));
+          break;
+      }
+    } else {
+      // Set right hand side to null.
+      vd.setInit(new NullLiteralExpr());
+    }
+    // Create a new statement with the simplified expression.
```
mernst: Here is another comment that doesn't add much to the code immediately below it.  Variable `exprStmt` will be returned, so giving it a name containing `result` would be more informative.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+    ExpressionStmt newStmt = new ExpressionStmt(exprCopy);
+    exprCopy.setParentNode(newStmt);
```
mernst: The `setParentNode` call is conceptually part of the constructor invocation, so I would add a blank line below it, and also a blank line above the comment, to show that those two lines go together as a logical chunk.

However, looking at the source code for JavaParser, it seems that this is already done by the constructor so the extra call is not necessary.  Remove the call or write a comment explaining why it is necessary.  For example, do you only have to do this because you are using an older version of JavaParser?

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+    // with mappings from fully-qualified names to simple type names. Also
+    // add necessary import statements.
+    for (Type type : fullyQualifiedNames) {
+      String typeName = type.toString();
+
+      // If the type is a generic type, handle the main type.  The other
+      // elements are included in the list and are processed separately.
+      int leftAngleBracketIndex = typeName.indexOf('<');
+      if (leftAngleBracketIndex >= 0) {
+        typeName = typeName.substring(0, leftAngleBracketIndex);
```
mernst: Use the `Type` data structure, rather than string parsing.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+
+  /**
+   * Get the simple type name of a fully-qualified type name. For example, {@code java.lang.String}
+   * should be simplified to {@code String}.
+   *
+   * @param typeName fully-qualified type name
+   * @return simple type name of {@code typeName}
+   */
+  public static String getSimpleTypeName(String typeName) {
+    int indexOflastSeparator = typeName.lastIndexOf('.');
```
mernst: Don't work with `String`s; use datatype operations instead.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+   * @param timeoutLimit number of seconds allowed for the whole test suite to run
+   * @return the standard output from running the Java file
+   */
+  private static String runJavaFile(
+      String filePath, String classpath, String packageName, int timeoutLimit) {
+    // Obtain directory to carry out compilation and execution step.
+    String executionDir = getExecutionDirectory(filePath, packageName);
+
+    // Obtain directory path from file path.
+    Path fPath = Paths.get(filePath).getParent();
```
mernst: What does `fPath` stand for?  I would have thought it had to do with the file, but it's actually the file's directory.

It's confusing that variables named `*Path` are sometimes strings and sometimes `Path`s.  Could you clean this up throughout the code?  It would be nice to do less string manipulation throughout the code.  For example, `writeToFile` could return a `File` rather than a `String`.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1185 @@
...
+   *     Returns {@code null} if more than one variable is declared in the {@code
+   *     VariableDeclarationExpr}.
+   */
+  private static Statement rhsAssignWithValue(VariableDeclarationExpr vdExpr, String value) {
+    if (vdExpr.getVars().size() != 1) {
+      // Number of variables declared in this expression is not one.
+      return null;
+    }
+
+    // Copy the variable declaration expression.
```
mernst: This comment says what the next line does, but that's clear from reading the next line.
It would be better to say what the variable is or how it will be used.  Why is the code making a copy?
In this case, it's because the copy will be side-effected and returned.  Giving the variable a more descriptive name than `exprCopy`, which expresses that, would be helpful.

I noticed the same thing about some of the other comments in your code; please try to make your comments informative.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+    }
+
+    File originalFile = new File(filePath);
+    // Create a new string to represent the new file name.
+    String minimizedFileName =
+        new StringBuilder(filePath).insert(filePath.lastIndexOf('.'), "Minimized").toString();
+    File minimizedFile = new File(minimizedFileName);
+
+    // Rename the overall class to [old class name][suffix].
+    String orgClassName = FilenameUtils.removeExtension(originalFile.getName());
```
mernst: `org` makes me think of `organization`.  Use `orig` instead.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+    try {
+      PackageDeclaration classPackage = compUnit.getPackage();
+      if (classPackage != null) {
+        packageName = classPackage.getPackageName();
+      }
+    } catch (NoSuchElementException e) {
+      // No package declaration.
+    }
+
+    File originalFile = new File(filePath);
```
mernst: The next 5 statements are a logical block, so give them a comment at the top:
```java
    // Rename file, and class inside it, to postpend "Minimized".
```

and remove the two blank lines in the middle.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+  /**
+   * Run a Java file and return the standard output.
+   *
+   * @param file the file to be compiled and executed
+   * @param classpath dependencies and complete classpath to compile and run the Java program
+   * @param packageName the package that the Java file is in
+   * @param timeoutLimit number of seconds allowed for the whole test suite to run
+   * @return the standard output from running the Java file
+   */
+  private static String runJavaFile(
```
mernst: I found this very hard to follow.  It mixed together operations on the classpath and non-classpath, for example.
I'm also not clear why both "." and `dirPath` are needed on the classpath.  Couldn't you set `dirPath` to "." if it would otherwise be null?  If not, please explain.

Something like this seems clearer:

```java
  private static String runJavaFile(
      File file, String userClasspath, String packageName, int timeoutLimit) {
    // Obtain directory to carry out compilation and execution step.
    File executionDir = getExecutionDirectory(file, packageName);

    // Directory path for the classpath.
    String dirPath;
    if (file.getParentFile() != null) {
      dirPath = file.getParentFile().getPath();
    } else {
      dirPath = ".";
    }

    String classpath = systemClassPath + pathSeparator + dirPath;
    if (classpath != null) {
      // Add user-specified classpath.
      classpath += pathSeparator + classpath;
    }

    // Fully-qualified classname.
    String fqClassName = FilenameUtils.removeExtension(file.getName());
    if (packageName != null) {
      fqClassName = packageName + "." + fqClassName;
    }

    // Command to run the Java file.
    String command = "java -classpath " + classpath + " org.junit.runner.JUnitCore " + fqClassName;

    // Run the specified Java file.
    return runProcess(command, executionDir, timeoutLimit).stdout;
  }
```

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+      dirPath = file.getParentFile().getPath();
+    }
+
+    // Fully-qualified classname.
+    String fqClassName = FilenameUtils.removeExtension(file.getName());
+    if (packageName != null) {
+      fqClassName = packageName + "." + fqClassName;
+    }
+
+    // Command to run the Java file.
```
mernst: Comments like these are not very informative; you can cut them.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+        classPath,
+        expectedOutput,
+        timeoutLimit,
+        verboseOutput);
+
+    // Cleanup: simplify variable type names and sort the import statements.
+    compUnit =
+        simplifyVariableTypeNames(
+            compUnit, packageName, minimizedFile, classPath, expectedOutput, timeoutLimit);
+    sortImports(compUnit);
```
mernst: `simplifyVariableTypeNames` should call `sortImports`.  It's natural for `simplifyVariableTypeNames` to do this work, and this way the client does not have to remember to do so.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+      int timeoutLimit,
+      boolean verboseOutput) {
+    for (TypeDeclaration type : cu.getTypes()) {
+      for (BodyDeclaration member : type.getMembers()) {
+        if (member instanceof MethodDeclaration) {
+          MethodDeclaration method = (MethodDeclaration) member;
+
+          // Get method's annotations.
+          for (AnnotationExpr annotationExpr : method.getAnnotations()) {
+            if (annotationExpr.toString().equals("@Test")) {
```
mernst: It's more natural to abstract this out into a routine, `isTestMethod`.
Then you can drop the for loop and the break statement, and the code will be easier to read.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+
+          if (mExp instanceof BinaryExpr) {
+            BinaryExpr binaryExp = (BinaryExpr) mExp;
+            // Check that the operator is an equality operator.
+            if (binaryExp.getOperator().equals(BinaryExpr.Operator.equals)) {
+              // Retrieve and store the value associated with the
+              // variable in the assertion.
+              String var = binaryExp.getLeft().toString();
+              String val = binaryExp.getRight().toString();
+              primitiveValues.put(var, val);
```
mernst: How do you know it's primitive?
I didn't see a test that verified that.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+
+    Type type = vdExpr.getType();
+    if (type instanceof PrimitiveType) {
+      // Replacement with zero value on the right hand side.
+      resultList.add(rhsAssignWithValue(vdExpr, type, null));
+    } else {
+      // Replacement with null on the right hand side.
+      resultList.add(rhsAssignWithValue(vdExpr, type, null));
+
+      if (type instanceof ReferenceType) {
```
mernst: How can this be false?  If it cannot be, then just do the cast without a check.
If a programmer writes a check, that is a signal to a reader that it could return either true or false at run time.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+      return null;
+    }
+  }
+
+  /**
+   * Return a variable declaration statement that sets the right hand side of a variable declaration
+   * to the value that is passed in.
+   *
+   * @param vdExpr variable declaration expression representing the current statement to simplify
+   * @param value value that will be assigned to the variable being declared
```
mernst: I was very confused by the calls to this routine, which looked wrong.  There is no documentation of the (surprising!) behavior when null is passed in.

This routine needs documentation of its behavior when null is passed in -- but I think it would be better to make its contract simpler.  You can write a separate routine `getZeroExpression` that returns a LiteralExpression for false, 0, or 0.0 as appropriate.
That would abstract out complexity to a self-contained, logical place.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+        case Double:
+          if (value == null) {
+            vd.setInit(new DoubleLiteralExpr("0.0"));
+          } else {
+            vd.setInit(new DoubleLiteralExpr(value));
+          }
+          break;
+        case Long:
+          if (value == null) {
+            vd.setInit(new LongLiteralExpr("0l"));
```
mernst: Always use uppercase "L" in long literal expressions.  The lowercase letter "l" looks too much like the number "1".

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+
+    CompilationUnit result = compUnit;
+    List<ImportDeclaration> imports = result.getImports();
+    for (String type : typeNameMap.keySet()) {
+      String compUnitStr = result.toString();
+
+      // Replace all instances of the fully-qualified type name with the
+      // simple type name.
+      if (compUnitStr.contains(type)) {
+        compUnitStr = compUnitStr.replace(type, typeNameMap.get(type));
```
mernst: Please use a visitor rather than heuristic string manipulation.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+    }
+    // Add JUnitCore command.
+    command += " org.junit.runner.JUnitCore " + fqClassName;
+
+    // Run the specified Java file.
+    return runProcess(command, executionDir, timeoutLimit).stdout;
+  }
+
+  /**
+   * Get directory to execute command in given file path and package name.
```
mernst: It's good to say how the method is used -- this is important.  Thanks.  
Please also give the method's specification.  What does it return?  Example input and output could help here if you find it more concise or understandable.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1183 @@
...
+   * @param packageName package name of input Java file
+   * @return the directory to execute the commands in. Null if packageName is null.
+   */
+  private static File getExecutionDirectory(File file, String packageName) {
+    if (packageName == null) {
+      return null;
+    }
+
+    // Determine how many layers above we should be executing the process in.
+    int foldersAbove = StringUtils.countMatches(packageName, ".") + 2;
```
mernst: Why +2?  Please document magic constants.

src/test/java/randoop/main/minimizer/MinimizerTests.java
bjkeller: Has this not been fixed?  This looks odd to me.  Create a File object from String and work with that.

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1374 @@
...
+   * (as few lines of code as possible) and that fails the same way:
+   *
+   * <ol>
+   *   <li>Same failing assertions as in the original input test suite.
+   *   <li>Same stacktrace produced by failing assertions.
+   * </ol>
+   *
+   * <p>The original input Java file will be compiled and run once. The "expected output" is a map
+   * from test method name to failure stack trace. A method is included in the map only if the
+   * method contains a failing assertion. Thus, the expected output of a test suite with no failing
```
huangwaylon: I think the problem with my comment here is that I'm intending for "expected output" to mean the standard output obtained from running the input test suite.

I have changed the comment to:
`The original input Java file will be compiled and run once. The "expected output" derived from the standard output from running the input file is a map from test method name to failure stack trace. A method is included in the map only if the method contains a failing assertion. Thus, the "expected output" of running a test suite with no failing tests will be an empty map. The "expected output" will be used in subsequent runs of the modified test suite to determine whether or not the test suite still fails in the same way.`

src/main/java/randoop/main/Minimize.java
```diff
@@ -0,0 +1,1374 @@
...
+      CompilationUnit cu,
+      String packageName,
+      File file,
+      String classpath,
+      Map<String, String> expectedOutput,
+      int timeoutLimit,
+      boolean verboseOutput) {
+    for (TypeDeclaration type : cu.getTypes()) {
+      for (BodyDeclaration member : type.getMembers()) {
+        if (member instanceof MethodDeclaration) {
```
huangwaylon: Would it be reasonable to attempt to simplify variable names in the next version of the minimizer?

PR #168, opened 2017-03-19
DigDog Features for weighted constants and weighted sequences

moychal:
Contains three main features that bias Randoop's sequence selection towards higher branch and line coverage:

- Implements a static weighting scheme for constants using a modified tf-idf formula.
- Implements a dynamic weighting scheme for executed sequences based on runtime, size, and number of times run.
- Implements a weighted random selection algorithm to randomly selecting an input sequence from a weighted distribution of sequences, given the presence of any of the above weighting schemes.

We have written a report on our findings on these features. Our results indicate that these are promising.

[DigDog-Improving-Randoop.pdf](https://github.com/randoop/randoop/files/852994/DigDog-Improving-Randoop.pdf)
Our evaluation script is here: [https://github.com/jkotalik/randoopEvaluation](url)

We need to discuss:

1. Are these features suitable for Randoop? Further tuning is possible, especially to the interaction between the two weighting schemes.
2. If suitable, how we would like to merge them. The changes are made to be as stand-alone as possible. But, they would be cleaner if integrated into the existing code and/or if the existing code is refactored further.
3. Failing tests. The runNaiveCollectionsTest in `RandoopSystemTest.java` consistently fails. Example trace: 

```
randoop.main.RandoopSystemTest > runNaiveCollectionsTest STANDARD_OUT
    Randoop:
    PUBLIC MEMBERS=119
    Explorer = randoop.generation.ForwardGenerator@619a5dff

    Normal method executions: 1900401
    Exceptional method executions: 161

    Average method execution time (normal termination):      0.0398
    Average method execution time (exceptional termination): 0.165

    No error-revealing tests to output

    Regression test output:
    Regression test count: 344
    Writing JUnit tests...

    Created file: /homes/iws/shintaku/490e1/project/randoop/build/working-directories/naive-collections-test/java/foo/bar/NaiveRegression0.java
    Created file: /homes/iws/shintaku/490e1/project/randoop/build/working-directories/naive-collections-test/java/foo/bar/NaiveRegression.java

    Invalid tests generated: 0

    Ignoring java2.util2.TreeSet.$jacocoInit() in coverage checks

    Ignoring java2.util2.Collections.access$000(java.lang.Object, java.lang.Object) in coverage checks
    Ignoring java2.util2.Collections.unmodifiableList(java2.util2.List) in coverage checks
    Ignoring java2.util2.Collections.$jacocoInit() in coverage checks

    Ignoring java2.util2.ArrayList.add(int, java.lang.Object) in coverage checks
[...]

PR #175, opened 2017-03-29
DigDog: Weighted Literals and Weighted Sequences

moychal:
A continuation of this PR: #168

Here's a short summary of our team's changes:

1. When retrieving the sequences to randomly select from, Randoop instead retrieves only class-level literals with probability `--p-const`.  Otherwise, Randoop behaves as before and retrieves all valid sequences. This is by default, through `--literals-level=CLASS-OR-ALL`
2. When selecting the sequences randomly, Randoop does a weighted random selection, where each sequence's weight factors in up to 3 individual weighting schemes:
    - A static weighting scheme for sequences which are literals, using a modified tf-idf formula.
    - A dynamic weighting scheme for executed sequences based on runtime, size, and number of times run.
    - A static weighting scheme for all sequences, which is just their immutable weight field

Report: [DigDog-Improving-Randoop.pdf](https://github.com/randoop/randoop/files/877601/DigDog-Improving-Randoop.pdf)
Evaluation script: https://github.com/jkotalik/randoopEvaluation

**Design Changes**
- ~~`--literals-level` is obsolete now. The package literal-level is removed, and there are only the class and global/all literal-levels (which are always enabled).~~
- `--literals-level` is updated to support the addition of weighted literals through the new default value "CLASS-OR-ALL"
- `--small-tests` is changed as it now incorporates the weighted literals/constants but ignores the new weighting schemes.

**Failing tests**
- `ICSE07ContainersTest.java` is excluded from the test module, as it seemed all of its tests were failing due to specific coverage expectations.
- There are many coverage inconsistencies introduced, requiring many ignores by the coverage checker in `RandoopSystemTest.java`.  These inconsistencies are also dependent on the test environment (local machine vs Travis).  I'm not sure what's causing this.
- ~~in systemTest, `runLiteralsTest` and `runRandoopOutputSequenceInfo` are failing due to Randoop exiting badly, which may be due to the different coverage.  Example trace:~~
``` 
java.lang.AssertionError: Randoop exited badly, exit value = 1
	at org.junit.Assert.fail(Assert.java:88)
	at randoop.main.RandoopRunStatus.generateAndCompile(RandoopRunStatus.java:83)
	at randoop.main.RandoopSystemTest.generateAndCompile(RandoopSystemTest.java:1139)
	at randoop.main.RandoopSystemTest.generateAndTestWithCoverage(RandoopSystemTest.java:972)
	at randoop.main.RandoopSystemTest.generateAndTestWithCoverage(RandoopSystemTest.java:999)
	at randoop.main.RandoopSystemTest.runLiteralsTest(RandoopSystemTest.java:433)

```
- ~~In systemTest, `runNaiveCollectionsTest` is failing, but instead due to not covering expected methods:~~

```
    java.lang.AssertionError: Expected methods not covered:
      java2.util2.Collections.enumeration(java2.util2.Collection)
      java2.util2.Collections.synchronizedSet(java2.util2.Set)
      java2.util2.Collections.unmodifiableMap(java2.util2.Map)
      java2.util2.Collections.unmodifiableSet(java2.util2.Set)
      java2.util2.LinkedList.clone()
      java2.util2.LinkedList.remove(java.lang.Object)
      java2.util2.LinkedList.set(int, java.lang.Object)
      java2.util2.LinkedList.toArray()
        at org.junit.Assert.fail(Assert.java:88)
        at randoop.main.CoverageChecker.checkCoverage(CoverageChecker.java:124)
        at randoop.main.RandoopSystemTest.generateAndTestWithCoverage(RandoopSystemTest.java:982)
        at randoop.main.RandoopSystemTest.runNaiveCollectionsTest(RandoopSystemTest.java:304)
```


bjkeller:
For the tests:
1. Please restore `ICSE07ContainersTest.java`, but edit `build.gradle` and add an exclusion with a comment describing the reason in the `test` configuration. (There are other examples starting at line 232 in the file. You can also search for "Temporary exclusion".)

2. The failure in the tests where Randoop is "exiting badly" appears to be directly due to changes to the command-line that you make in the PR.  In particular, you say that the behavior for `--literals-level` has changed,  Also, `runRandoopOutputSequenceInfo` is introduced by the PR, and my guess is that it is using command-line arguments that you just removed.

3. For `runNaiveCollectionsTest` you'll see many lines like
```
coverageChecker.exclude("java2.util2.ArrayList.add(int, java.lang.Object)");
```
You'll need to add a line like this for each of the missing methods in the failed assertion.

moychal:
I still think `runNaiveCollectionsTest` has issues. The coverage checker seems to not exclude things on Travis despite being specifically excluded with `coverageChecker.exclude()`, but it is correctly excluded on my local machine. I reverted our team's changes in `OperationModel.java` to use the TreeSet instead of LinkedHashSet to be consistent. The test from commit #fd1f131 on Travis says:

```
randoop.main.RandoopSystemTest > runNaiveCollectionsTest FAILED
    java.lang.AssertionError: Expected methods not covered:
      java2.util2.ArrayList.get(int)
      java2.util2.ArrayList.remove(int)
      java2.util2.Collections.unmodifiableMap(java2.util2.Map)
      java2.util2.LinkedList.add(int, java.lang.Object)
      java2.util2.LinkedList.remove(java.lang.Object)
      java2.util2.LinkedList.toArray()
        at org.junit.Assert.fail(Assert.java:88)
        at randoop.main.CoverageChecker.checkCoverage(CoverageChecker.java:124)
        at randoop.main.RandoopSystemTest.generateAndTestWithCoverage(RandoopSystemTest.java:964)
        at randoop.main.RandoopSystemTest.runNaiveCollectionsTest(RandoopSystemTest.java:314)
```
Despite these being explicitly excluded with `coverageChecker.exclude()`.  On my local machine I have different results with:

```
randoop.main.RandoopSystemTest > runNaiveCollectionsTest FAILED
    java.lang.AssertionError: Expected methods not covered:
      java2.util2.LinkedList.get(int)
        at org.junit.Assert.fail(Assert.java:88)
        at randoop.main.CoverageChecker.checkCoverage(CoverageChecker.java:124)
        at randoop.main.RandoopSystemTest.generateAndTestWithCoverage(RandoopSystemTest.java:983)
        at randoop.main.RandoopSystemTest.runNaiveCollectionsTest(RandoopSystemTest.java:333)
```
At the latest commit, I have the above contradictory methods ignored.

bjkeller:
Using ordered sets for types was the change I added that seems to have resolved the non-determinism on the master branch, so we definitely want to keep that.  As far as I know, all of the reflection methods that return arrays do not guarantee the order will be the same on each run, and we've had problems with this in the past.

Unfortunately, there are still some discrepancies between Travis and local machines that we need to track down.  So, thanks for getting the exclusions/ignores setup so that the tests pass on Travis.

moychal:
Ok, I'll get to tracking down those discrepancies. I'm not sure if it was before I used the ordered sets, but there might've been discrepancies due to other methods returning collections in unstable order.

bjkeller:
Michael,  
We are going to work on trying to figure out where the non-determinism is coming from in this test. You can feel free to pause until we have a chance to look deeper.
Ben

moychal:
I'm just about done. After this current Travis build passes/fails I'll have classified the methods that have shown themselves to be inconsistent thus far between Travis and my local machine.  I was going to do one last Travis build to determine whether a set of methods ignored from the checker were failing because of my previous use of the unordered set, or whether it's a bug in the coverageChecker.

moychal:
There also seems to be non-determinism causing runCollectionsTest() to fail, by the method:

`java2.util2.Collections.max(java2.util2.Collection, java2.util2.Comparator)`

However I've only seen it on my local machine. It seems to always be covered on Travis.

bjkeller:
Yes, I had seen problems there too. 

The issue with non-determinism seems to relate to using the classes in `src/testinput/java/java2/util2/` as inputs. These classes are rawtype versions of the `java.util` classes.

moychal:
`--literals-level=CLASS-OR-ALL` with `--p-const` introduces more coverage inconsistencies than `--literals-level=CLASS`.  I'm not sure why there are coverage inconsistencies with `--literals-level=CLASS`, as the other changes outside of the `--literals-level=CLASS-OR-ALL` feature were just to bias the random selection.  Furthermore, there are still platform specific coverage inconsistencies between my local machine and Travis.

bjkeller:
The nondeterminism we've encountered in other places has been associated with using reflection methods that return an array of objects but don't guarantee the order will be consistent between calls. I would guess that is happening here. 

The solution we've used is to make sure the objects have a well-defined order, and then placed them in a TreeSet to force the ordering.

moychal:
I'm still puzzled as to why my team's changes are causing coverage inconsistencies between machines and when isolated on a single machine.  The map and set structures we used shouldn't change the forced ordering from the TreeSet, and it doesn't seem the randomness in selecting the combination of weighting schemes (via `--p-const`) affected the inconsistencies.

I'm currently checking to see if the random weighted selection algorithm we added in `Randomness.java` is affecting the inconsistencies, but I doubt it. I suspect there's some underlying side effects or timing issues of passing the structures along, like in our changes to `ClassLiteralExtractor.java`.

bjkeller:
I haven't been able to find it in this input library, but there is a chance that there is an issue with the code under test that results in nondeterministic behavior.  Because, for instance, new constants are collected as sequences are executed, methods that have different behavior across runs could result in differences. In this case, a Randoop test expecting the same methods to be covered is going to be flaky, even without your changes.

bjkeller:
It would be good if we can figure out what is going on with this before your team leaves campus for the summer and our collective memories get too faded.

What I'm wondering is whether your changes allow Randoop to capture values that can vary between executions. My proposal is that you try another experiment to see if the problem is still present when you disable any additional constant collection that you introduced.

Would you need help doing this?

moychal:
Yes, Randoop should be able to capture values that vary between executions with our changes.  But I thought Randoop already incorporated that with its default random selection of sequences?

These problems with coverage changes still existed with the constant collection disabled: [https://github.com/randoop/randoop/pull/175#issuecomment-293123204](url)
So I think that something with the incorporated 3 weighting schemes or the code under test is causing the issues.

My team hasn't been in contact since the end of last quarter, so help would be greatly appreciated.

mernst:
Closing in favor of new `digdog` branch that doesn't suffer merge conflicts and reduces differences in tests.

PR #181, opened 2017-05-17
JUnit Test Suite Minimizer Fix

huangwaylon:
Fixes issue of the minimizer where import statements were not being added for classes from the same package. Added test case to demonstrate that the minimizer handles this case.


bjkeller:
Have you done a pull and merge of master recently?

PR #188, opened 2017-06-27
Selection log

mernst:


src/main/java/randoop/util/Randomness.java
```diff
@@ -36,21 +38,54 @@ public static int nextRandomInt(int i) {
     if (Log.isLoggingOn()) {
       Log.logLine("randoop.util.Randomness: " + totalCallsToRandom + " calls so far.");
     }
-    return random.nextInt(i);
+    int value = random.nextInt(i);
+    if (GenInputsAbstract.selection_log != null) {
+      StackTraceElement[] trace = Thread.currentThread().getStackTrace();
```
mernst: This code is repeated almost identically multiple times in the file.  Could you please abstract it out into a method?  That will also make it easier to call from elsewhere.

src/main/java/randoop/util/Randomness.java
```diff
@@ -36,21 +38,54 @@ public static int nextRandomInt(int i) {
...
+      try {
+        String msg = String.format("nextRandomInt: %d called from %s%n", value, trace[2]);
+        GenInputsAbstract.selection_log.write(msg);
+        GenInputsAbstract.selection_log.flush();
+      } catch (IOException e) {
+        throw new RandoopLoggingError("Error writing to selection-log: " + e.getMessage());
```
mernst: Please include the name of the file in the message.  (Unless the file name is always in the IOException, but I don't think it is.)

PR #189, opened 2017-06-29
Omitmethods

bjkeller:
This branch changes the implementation of `--omitmethods` to
1. match against the signature of the method instead of `java.lang.reflect.Method.toString()`
2. omit methods that are either exact matches or that are inherited from a class or interface where the method is an exact match.

Most of the files changed are due to moving the pattern matching out of the `DefaultReflectionPredicate` and into `OperationExtractor`. The notable changes are in `OperationExtractor` where it is necessary to search for methods that match the patterns; and `TypedClassOperation` and several classes in `randoop.types` where methods were introduced to form the signature that is matched against.

There is search code in `OperationExtractor` that could be generalized to a method in the `Types` hierarchy. After starting to work on it, I think it would be better to do a careful refactor later.


src/docs/manual/index.html
```diff
@@ -794,21 +794,18 @@ <h2 id="command-line-options">Command-line options</h2>
...
-       method.
- </ul>
+             A regex that indicates methods that should not be called directly in generated tests. This does
+ not prevent indirect calls to such methods from other, allowed methods.
+
+ <p>Randoop will not directly call a method whose fully-qualified signature matches the regular
+ expression, or a method inherited from a superclass or interface for which the method matches
+ the regular expression.
+
+ <p>The signature has the form <code>package-name.classname.method-name(argument-list)</code> where
```
mernst: It's weird that --methodlist and --omitmethods use two different formats.  (At the least, the prefix and spacing is different; the --omitmethods documentation should say how a constructor is named.)  I suggest that these use the same format.

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -41,24 +50,51 @@
   /** The class type of the declaring class for the collected operations */
   private ClassOrInterfaceType classType;
 
+  /** The list of {@code Pattern} objects to omit matching operations */
+  private List<Pattern> omitPatterns;
+
   /**
    * Creates a visitor object that collects the {@link TypedOperation} objects corresponding to
-   * members of the class type and satisfying the given predicate.
+   * members of the class type and satisfying the given predicates.
```
mernst: OperationExtractor documentation doesn't mention omitmethods, but the class does use omitmethods as a filter

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -174,4 +174,32 @@ public boolean hasWildcardTypes() {
     }
     return new ArrayList<>(paramSet);
   }
+
+  /**
+   * Returns the raw type signature for this operation. This signature consists of the classname as
+   * a fully-qualified raw type, the method name, and the argument types as fully-qualified raw
+   * types.
+   *
+   * <p>The raw type signature for a constructor <code>C()</code> has has this form instead of the
```
mernst: It's unclear what "this form is".  Is a constructor name the name of the class?  It would be good to be explicit.

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -174,4 +174,32 @@ public boolean hasWildcardTypes() {
...
+   * reflection form <code>C.&lt;init&gt;</code>.
+   *
+   * @return the signature for this operation using raw types for the class and argument types
+   */
+  public String getRawSignature() {
```
mernst: getRawSignature is called frequently.  How about caching its result?

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -185,4 +222,195 @@ public void visit(Enum<?> e) {
...
+  /**
+   * Indicates whether some omit pattern matches the {@link TypedClassOperation#getRawSignature()}
+   * for the given operation.
+   *
+   * @param operation the constructor or method call to match against the omit patterns of this
+   *     extractor
+   * @return true if the signature matches an omit pattern, and false otherwise.
+   */
+  private boolean omit(TypedClassOperation operation) {
+    if (omitPatterns.isEmpty()) {
```
mernst: If you cache the result of getRawSignature as suggested elsewhere, you can remove this special-case optimization and reduce code size/clutter a bit.

src/docs/manual/index.html
bjkeller: I'll agree the `--methodlist` format is odd, but there is a difference between giving a loadable name and a regex.  There is a parser for the names that can be more tolerant of spacing differences, while to allow variation in regex input seems like trouble.

Incidentally, the replacement file format in the call-avoidance branch is more like what --methodlist should allow.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -106,28 +107,30 @@ private OperationModel() {
...
-   * @return the operation model for the parameters
-   * @throws OperationParseException if a method signature is ill-formed
+   * @return the {@link OperationModel} constructed with the given arguments
+   * @throws SignatureParseException if a method signature is ill-formed
    * @throws NoSuchMethodException if an attempt is made to load a non-existent method
    */
   public static OperationModel createModel(
       VisibilityPredicate visibility,
       ReflectionPredicate reflectionPredicate,
+      OmitMethodsPredicate omitPredicate,
```
mernst: Why does OperationModel.createModel take an OmitMethodsPredicate rather than a list of patterns?
OmitMethodsPredicate is only ever created and immediately passed into OperationModel.createModel.
How about making OperationModel.createModel take as input a List<Pattern> instead?  That can move the creation of the OmitMethodsPredicate into OperationModel, so it will appear once in the source code instead of many times.

src/main/java/randoop/main/GenTests.java
```diff
@@ -551,12 +546,15 @@ private static SimpleLog getSimpleLog(String filename) {
...
+   * Adds patterns for methods to be omitted to the given list and returns an {@link
+   * OmitMethodsPredicate} using the pattern list. Reads from the {@link
    * GenInputsAbstract#omitmethods_list} file.
    *
    * @param omitmethods the list of {@code Pattern} objects to add new patterns to, must not be null
+   * @return the {@link OmitMethodsPredicate} for the omit patterns from both {@link
+   *     GenInputsAbstract#omitmethods} and {@link GenInputsAbstract#omitmethods_list}.
    */
-  private void extendOmitMethods(List<Pattern> omitmethods) {
+  private OmitMethodsPredicate createOmissionPredicate(List<Pattern> omitmethods) {
```
mernst: The code in createOmissionPredicate feels out of place, because it is doing two distinct things.  First, it is doing command-line option processing, by setting omitmethods correctly, according to other command-line arguments.  Second, it is creating an OmitMethodsPredicate.  I feel that the first part should appear in GenTests, just after the call to checkOptionsValid(), where other variables are set.  The code would update omitmethods from other variables, so from that point onward omitmethods can be used without referring to the other variables.  This will also accommodate the simplification of OmitMethodsPredicate suggested above.
(Hmm, I wrote that suggestion while reading the new version of the code.  Now that I see the diffs, they indicate that the code used to work that way.  There was a method named extendOmitMethods that you changed into createOmissionPredicate by adding a single line to its end.  I suggest that you reverse the change.)

src/main/java/randoop/reflection/DefaultReflectionPredicate.java
```diff
@@ -20,32 +18,21 @@
  */
 public class DefaultReflectionPredicate implements ReflectionPredicate {
 
-  private List<Pattern> omitMethods = null;
   private Set<String> omitFields;
```
mernst: Please document the field, and in particular the format of these strings.

src/main/java/randoop/reflection/DefaultReflectionPredicate.java
```diff
@@ -20,32 +18,21 @@
...
-    this(null, new HashSet<String>());
+    this(new HashSet<String>());
   }
 
   /**
    * DefaultReflectionFilter creates a filter object that uses default criteria for inclusion of
    * reflection objects.
    *
-   * @param omitMethods pattern for methods to omit, if null then no methods omitted
    * @param omitFields set of field names to omit
```
mernst: Please document the format:  are they simple names or fully-qualified?

src/docs/manual/method_list_example.txt
```diff
@@ -1,54 +1,54 @@
...
-method : java.util.TreeSet.subSet(java.lang.Object,java.lang.Object)
+java.util.TreeSet()
+java.util.TreeSet(java.util.Collection)
+java.util.TreeSet(java.util.Comparator)
+java.util.TreeSet(java.util.SortedSet)
+java.lang.Object()
+java.util.AbstractCollection.containsAll(java.util.Collection)
+java.util.AbstractCollection.retainAll(java.util.Collection)
+java.util.AbstractCollection.toArray()
+java.util.AbstractCollection.toArray([Ljava.lang.Object;)
```
mernst: The documentation of a fully-qualified signature does not indicate how to specify array types.  Please update the documentation.  It looks like you are using either the Class.getName format or field descriptiors, but not fully-qualified types which suggests use of the Java source code format; calling them fully-qualified names is misleading.  (I would guess you are using Class.getName.)

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -118,18 +144,20 @@ public void visit(Method method) {
...
-                operation.getOperation(),
-                classType,
-                operation.getInputTypes(),
-                operation.getOutputType());
+        operation = operation.getOperationForType(classType);
       }
     }
-    addOperation(operation);
+    if (operation == null) {
+      System.out.println("null operation for " + classType + " method " + method.getName());
```
mernst: Should this code issue a more prominent error, such as crashing Randoop?
If not, make this message more specific about whether this is a problem or just normal operation.

src/main/java/randoop/reflection/OperationModel.java
bjkeller: The constructor is only called once in `src/main`. 

Are you talking about the effect that changing the interface has on diffs of test classes?

bjkeller:
Ready for next review.

There are a couple of new things:
1. fixed a bug where Randoop would halt if no `--testclass` or `--classfile` is given but there is a `--methodlist`
2. Uses the `randoop.reflection.RawSignature` class from `condition-specification` in `randoop.operation.TypedClassOperation.getRawSignature()`  

There is a design problem made obvious by the `randoop.operation.TypedClassOperation.getRawSignature()` method. The method is only defined over operations that are Methods or Constructors and not fields (which show up as synthetic setter/getter operations). This relates somewhat to a comment you had made previously about being able to indicate both fields and methods for omission (i.e., have an `--omitmembers`). Given the rest of what needs to be done, I'm not trying to fix this.

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -116,7 +117,7 @@ public TypedClassOperation applyCaptureConversion() {
   }
 
   /**
-   * Produces a Java source code representation of this statement and append it to the given
+   * Produces a Java source code representation of this statement and appends it to the given
```
mernst: What is "this statement"?

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -68,12 +70,11 @@ public int hashCode() {
    */
   @Override
   public int compareTo(TypedOperation op) {
-    int result = 0;
```
mernst: It's confusing that `compareTo` is defined in TypedClassOperation, in TypedOperation, and in TypedTermOperation, and all three implementations have hard-coded knowledge of the complete type hierarchy.
The implementations will be hard to keep consistent.

Please move all the logic into the superclass, TypedOperation.  (It can do run-time tests and casts as needed.)  Remove the definitions of `compareTo` from TypedClassOperation and TypedTermOperation.

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -68,12 +70,11 @@ public int hashCode() {
...
     if (op instanceof TypedTermOperation) {
       return 1;
     }
     TypedClassOperation other = (TypedClassOperation) op;
-    result = declaringType.compareTo(other.declaringType);
+    int result = declaringType.compareTo(other.declaringType);
```
mernst: On line 85 below, it's confusing that the overview differs from the `@return` clause.  What does it mean for an operation to represent a value, and how can a TypedClassOperation do so?

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -116,7 +117,7 @@ public TypedClassOperation applyCaptureConversion() {
   }
 
   /**
-   * Produces a Java source code representation of this statement and append it to the given
```
mernst: In `apply()` on line 101/102 above, please improve the Javadoc.  It wasn't clear to me whether this was related somehow to the type variables in the declaringClass, if the declaringClass is generic.  The documentation for the overridden method is clearer, but then I'm not sure exactly how this overriding implementation is different.

src/main/java/randoop/reflection/SignatureParser.java
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Parses a fully-qualified signature and returns the corresponding {@code
+   * java.lang.reflect.AccessibleObject}.
+   *
+   * <p>A signature is expected to have the form
+   *
+   * <ul>
+   *   <li>{@code package-name.classname.method-name(argument-list)} for a method,
+   *   <li>{@code package-name.classname.&lt;init&gt;(argument-list)} or {@code
```
mernst: Inside `{@code ...}`, use `<` rather than `&lt;`.  (That's an advantage of `{@code ...}`.)

src/main/java/randoop/reflection/SignatureParser.java
```diff
@@ -0,0 +1,152 @@
...
+   *       package-name.classname(argument-list)} for a constructor.
+   * </ul>
+   *
+   * <p>where {@code package-name} is a period-separated list of identifiers, and <code>
```
mernst: Use `{@code ...}` instead of `<code>...</code>`.  It's shorter, is usually easier to read, and suffers fewer formatting artifacts (such as extra spaces before or after) in the resulting HTML.

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -174,4 +186,49 @@ public boolean hasWildcardTypes() {
...
+
+  /**
+   * Creates an operation with the same name, input types, and output type as this operation, but
+   * having the given type as the owning class.
+   *
+   * <p>Note: this is only a valid object if {@code type} has the method. This is definitely the
+   * case if {@code type} is a subtype of the declaring type of the operation, but this method does
+   * not force that check because we sometimes want to create the operation for superclasses.
+   *
+   * @param type a subtype of the declaring class of this operation to substitute into the
```
mernst: This requires that the argument is a subtype, but the previous sentence said it wasn't necessarily one.

src/main/java/randoop/reflection/OmitMethodsPredicate.java
```diff
@@ -0,0 +1,229 @@
...
+      if (result) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Indicates whether an omit pattern matches the raw signature of the method in either the
+   * declaring class of the method or a supertype.
```
mernst: What is the relationship between "declaring class of the method" and "the type to which the method belongs"?  Please clarify, and use consistent terminology when possible.

src/main/java/randoop/reflection/OmitMethodsPredicate.java
```diff
@@ -0,0 +1,229 @@
...
+   * declaring class that have the method.
+   *
+   * @param classType the type to which the method belongs
+   * @param operation the operation for the method
+   * @param method the reflection object for the method
+   * @return true if the signature of the method in the current class or a super class matches an
+   *     omit pattern, false otherwise
+   */
+  boolean shouldOmit(
+      ClassOrInterfaceType classType, final TypedClassOperation operation, final Method method) {
```
mernst: Why are all three arguments needed?

The TypedClassOperation has a class associated with it already.  Is that guaranteed to be the same as `classType`, or are they guaranteed to be different, or are they sometimes the same and sometimes different?

I have the same questions about the relationship to `method`.

Please document the relationship among the arguments and what each one is used for.

src/main/java/randoop/reflection/SignatureParser.java
```diff
@@ -0,0 +1,154 @@
+package randoop.reflection;
+
+import java.lang.reflect.AccessibleObject;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/** Parses type signature strings used to identify methods and constructors in input. */
+// TODO: This duplicates should be factored into a separate source set (aka, module) so that it
```
mernst: How much work would it be to do this?
Because I made improvements to this section, they are already out of synch with one another.

src/main/java/randoop/reflection/SignatureParser.java
```diff
@@ -0,0 +1,154 @@
...
+   *
+   * <p>where {@code package-name} is a period-separated list of identifiers, and <code>
+   * argument-list</code> is a comma-separated (spaces-allowed) list of fully-qualified Java raw
+   * types. Array types have the format <code>element-type[]</code>.
+   *
+   * @param signature the string to parse: a signature string for a method or constructor, in the
+   *     above format
+   * @param visibility the predicate for determining whether the method or constructor is visible
+   * @param reflectionPredicate the predicate for checking reflection policy
+   * @return the {@code AccessibleObject} for the method or constructor represented by the string
```
mernst: It can also return null.  Document that fact, and under what circumstances null is returned.

src/main/java/randoop/reflection/SignatureParser.java
bjkeller: it is probably ~1/2 day, but it would be easier to do once everything is merged, because it will affect the build.

bjkeller:
omitmethods branch is updated -- ready for re-review

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -38,55 +39,117 @@
   /** The collection of operations */
   private final Collection<TypedOperation> operations;
```
mernst: The Javadoc could emphasize that this is the key result of the visitor.

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -38,55 +39,117 @@
...
       ClassOrInterfaceType classType,
       Collection<TypedOperation> operations,
       ReflectionPredicate predicate,
+      OmitMethodsPredicate omitPredicate,
       VisibilityPredicate visibilityPredicate) {
     this.classType = classType;
     this.operations = operations;
     this.predicate = predicate;
     this.visibilityPredicate = visibilityPredicate;
+    this.omitPredicate = omitPredicate;
```
mernst: Manipulate fields in the same order they are declared and the same order they are declared in the class.  (Otherwise it's too hard to keep track of them in your head.)  I suggest that `operations` should be last, because it is an output whereas all the others are inputs.  But, you might eliminate `operations` per my comment in email.

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -38,55 +39,117 @@
...
   /**
-   * Adds an operation to the collection of this extractor. If the declaring class type is an {@link
-   * InstantiatedType}, then the substitution for that class is applied to the types of the
-   * operation, and this instantiated operation is returned.
+   * Adds an operation to the collection of this extractor.
    *
    * @param operation the {@link TypedClassOperation}
    */
   private void addOperation(TypedClassOperation operation) {
     if (operation != null) {
```
mernst: At every call site, the argument is non-null.  So, make this an assertion, or remove it.  (Alternately, add documentation that the formal parameter can be null and what happens in that case, but I don't see the point of doing so.)

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -102,34 +165,44 @@ public void visit(Constructor<?> constructor) {
...
+    if (!omitPredicate.shouldOmit(operation)) {
+      addOperation(operation);
+    }
   }
 
   /**
    * Creates a {@link MethodCall} object for the {@link Method}.
    *
+   * <p>The created operation has the declaring class of {@code method} as the declaring type. An
+   * exception is a static method for which the declaring class is not public.
```
mernst: Document what is done in this case.  (It's currently in code comments.)

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -102,34 +165,44 @@ public void visit(Constructor<?> constructor) {
...
     if (!predicate.test(method)) {
       return;
     }
-    TypedClassOperation operation = TypedOperation.forMethod(method);
-    if (classType.isSubtypeOf(operation.getDeclaringType()) && operation.isStatic()) {
+    TypedClassOperation operation = instantiateTypes(TypedOperation.forMethod(method));
+    checkSubTypes(operation);
+
+    if (operation.isStatic()) {
+      // If this classType inherits this static method, but declaring class is not public, then
```
mernst: How can this happen?  static methods are never inherited.

Oh, you don't mean inheritance.  I think you are talking about how the compiler statically resolves which method gets called at a call site, which can be from a superclass.  Please use different terminology than that for inheritance, which is a different language mechanism.

src/main/java/randoop/reflection/SignatureParser.java
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Parses a fully-qualified signature and returns the corresponding {@code
+   * java.lang.reflect.AccessibleObject}.
+   *
+   * <p>A signature is expected to have the form
+   *
+   * <ul>
+   *   <li>{@code package-name.classname.method-name(argument-list)} for a method,
+   *   <li>{@code package-name.classname.<init>(argument-list)} or {@code
```
bjkeller: It is not that it is discouraged, it is just that for simplicity the documentation only uses the non-`<init>` form.

src/main/java/randoop/reflection/OperationExtractor.java
bjkeller: The JLS says static methods may be inherited https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8

mernst: OK, thanks for the correction.

bjkeller:
squash and merge performed and pushed to master

Issue #136, opened 2016-12-15
Detecting flaky tests on command line


bjkeller:
Generated tests can be flaky between executions inside Randoop, between Randoop execution and command-line execution, and between command-line executions. The last two need to be tested for, and, if possible, the failures need to be incorporated into test generation.

bjkeller:
It would be difficult to incorporate this filtering back into the test generation, so #219 only applies the filter when the regression tests are output.

PR #219, opened 2017-07-18
Flaky filter

bjkeller:
This pull-request implements a fix to issue #136.


src/main/java/randoop/output/MinimizerWriter.java
```diff
@@ -0,0 +1,30 @@
+package randoop.output;
+
+import java.io.File;
+import randoop.main.Minimize;
+
+/** Created by bjkeller on 7/3/17. */
```
mernst: Please write proper comments for all files/methods/fields.  Thanks!

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,232 @@
...
+      "\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*";
+
+  /** The {@link randoop.execution.TestEnvironment} for running the test classes. */
+  private final TestEnvironment testEnvironment;
+
+  /** The underlying {@link randoop.output.JavaFileWriter} for writing a test class. */
+  private final JavaFileWriter javaFileWriter;
+
+  /**
+   * Create a {@link FailingTestFilter} run in the environment and using the given {@link
```
mernst: Does run mean "that will be run", or "that was run"?

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,232 @@
...
+      Path workingDirectory = createWorkingDirectory(classname, pass);
+
+      boolean success = compileTestClass(sourceList, workingDirectory);
+      assert success;
+
+      RunStatus status = null;
+      try {
+        status = testEnvironment.runTest(qualifiedClassname, workingDirectory.toFile());
+      } catch (ProcessException e) {
+        throw new BugInRandoopException("Error filtering regression tests: " + e.getMessage());
```
mernst: As a rule, it is better to use the two-argument exception constructor:  pass in `e` as the `cause` argument.  This will give more information.  (When running in a non-verbose or non-debugging mode, and the handler for BugInRandoopException can always just use the cause's message and no more, if you want to avoid the extra output that the cause engenders.)

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+
+    int pass = 0;
+    boolean passing = false;
+
+    while (!passing) {
+
+      Path workingDirectory = createWorkingDirectory(classname, pass);
+
+      boolean success = compileTestClass(sourceList, workingDirectory);
+      assert success;
```
mernst: If this is important, why not write it as `if (!success) throw ...` so that the (cheap) test is always done and a useful message is printed?

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+  @Override
+  public File writeClass(String packageName, String classname, String classString) {
+
+    List<File> sourceList = new ArrayList<>();
+    File testFile = javaFileWriter.writeClass(packageName, classname, classString);
+    sourceList.add(testFile);
+
+    String qualifiedClassname = (packageName.isEmpty() ? "" : packageName + ".") + classname;
+
+    int pass = 0;
```
mernst: Document what this is for:  to create unique directory names.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+      } catch (ProcessException e) {
+        throw new BugInRandoopException("Error filtering regression tests: " + e.getMessage());
+      }
+      if (status.exitStatus == 0) {
+        passing = true;
+        continue;
+      }
+
+      /*
+       * Read the standard output from running the test class using JUnit4.
```
mernst: This comment is confusing:  it says "read the standard output", but the code reads the error output, and the assertion message mentions both standard out and error output.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+       */
+      assert status.errorOutputLines.isEmpty()
+          : "JUnit writes to standard out, error output unexpected";
```
mernst: Make assertion output more verbose.  I have changed some already.  It's worth clearly explaining what is going on -- many, including this one, are obscure even in the context of the code.  Say that there was error output and it was unexpected; currently, it's unclear both what happened and whether than was a problem.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+        status = testEnvironment.runTest(qualifiedClassname, workingDirectory.toFile());
+      } catch (ProcessException e) {
+        throw new BugInRandoopException("Error filtering regression tests: " + e.getMessage());
+      }
+      if (status.exitStatus == 0) {
+        passing = true;
+        continue;
+      }
+
+      /*
```
mernst: This method body is long, with no overview of what is happening.  I suggest that you abstract out the code from here to line 173 into a separate method.  With a useful name and documentation, this will make both it and the code that calls it more comprehensible.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+       * Read the standard output from running the test class using JUnit4.
+       */
+      assert status.errorOutputLines.isEmpty()
+          : "JUnit writes to standard out, error output unexpected";
+
+      /*
+       * First, find the message that indicates the number of failures in the run.
+       */
+      int totalFailures = -1;
+      Iterator<String> lineIterator = status.standardOutputLines.iterator();
```
mernst: It's not clear that lineIterator will be used for more than just this one bit of the code.  In fact, its declaration under the "First, find ..." comment suggests that it's only used here.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+      while (lineIterator.hasNext() && totalFailures < 0) {
```
mernst: This way of writing a while loop is confusing.  I would put a `break;` statement after the assignment to `totalFailures`, which clarifies the exit condition at the relevant place.  Then you don't need the second clause at all.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+        String line = lineIterator.next();
+        Matcher messageMatcher = FAILURE_MESSAGE_PATTERN.matcher(line);
+        if (messageMatcher.matches()) {
+          totalFailures = Integer.parseInt(messageMatcher.group(1));
+        }
+      }
+      assert lineIterator.hasNext()
```
mernst: It's unclear why you are testing lineIterator rather than totalFailures.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+                      + classname
+                      + "\\.java:(\\d+)\\)");
+          lineNumber = -1;
+          while (lineIterator.hasNext() && lineNumber < 0) {
+            line = lineIterator.next();
+            Matcher lineMatcher = linePattern.matcher(line);
+            if (lineMatcher.matches()) {
+              lineNumber = Integer.parseInt(lineMatcher.group(1));
+            }
+          }
```
mernst: lineNumber might be -1 at this point.  The other 3 while loops checked for the possibility of no match, but this one does not.  Why not?

Also, you could abstract out all the while loops into a readUntilRegexMatch routine that would be called at each of the places that currently has a while loop.  This would simplify and clarify the code, and guarantee consistent error handling.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+   *
+   * @param sourceList the list of source files
+   * @param outputDirectory the directory for class file output
+   * @return true if the classes compile without error
+   */
+  private boolean compileTestClass(List<File> sourceList, Path outputDirectory) {
+    boolean success = false;
+    FileCompiler compiler = new FileCompiler();
+    try {
+      success = compiler.compile(sourceList, outputDirectory);
```
mernst: Change `success =` to `return` and you can delete lines 199 and 207.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+   * Compiles the Java files in the list of files and writes the resulting class files to the
+   * directory.
+   *
+   * <p>Calls {@code System.exit()} if the file does not compile.
+   *
+   * @param sourceList the list of source files
+   * @param outputDirectory the directory for class file output
+   * @return true if the classes compile without error
+   */
+  private boolean compileTestClass(List<File> sourceList, Path outputDirectory) {
```
mernst: It's confusing that the name is singular -- "compile test class" -- but the first parameter is a list.  However, the list is always a singleton, so make it a `File` instead of a `List<File>`, for simplicity.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,235 @@
...
+            if (lineMatcher.matches()) {
+              lineNumber = Integer.parseInt(lineMatcher.group(1));
+            }
+          }
+          assert lineNumber - 1 < classLines.length;
+          classLines[lineNumber - 1] = "// flaky: " + classLines[lineNumber - 1];
+        }
+      }
+
+      classString = UtilMDE.join(classLines, Globals.lineSep);
```
mernst: It's inefficient to create a string just for the purpose of passing it to `writeClass`.  Could you create a version of writeClass that takes `List<String>` rather than `String` for the file text?

Even more efficient would be to avoid writing the .java source code to a file at all.  This shows how to create a SimpleJavaFileObject that lets javac be invoked on code stored in a String:
http://docs.oracle.com/javase/7/docs/api/javax/tools/JavaCompiler.html
Wouldn't that be more efficient than writing the .java source code to a file?
And it could save the overhead of starting a new JVM to run each instance of javac.

src/main/java/randoop/execution/ProcessException.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.List;
+
+/**
+ * Exception representing an error that occurs while running a process with {@link
+ * RunEnvironment#run(List, File, long)}.
+ */
+public class ProcessException extends Throwable {
+  private static final long serialVersionUID = 736230736083495268L;
+
+  public ProcessException(String s, Throwable e) {
```
bjkeller: The Java 8 javadoc doesn't have this flag, and I'm not seeing it any documentation. 

There is the accepted answer to this question on SO:
https://stackoverflow.com/questions/29519085/how-to-fail-gradle-build-on-javadoc-warnings
but it seems potentially fragile

src/main/java/randoop/main/GenTests.java
```diff
@@ -696,76 +738,46 @@ private void handleFlakySequenceException(AbstractGenerator explorer, SequenceEx
...
-        outputTestDriver(
-            GenInputsAbstract.regression_test_basename, junitCreator, testClassNames, jfw));
-
-    if (!GenInputsAbstract.noprogressdisplay) {
-      logTestFiles(testFiles);
+    // create and write driver
+    String driverName = namePrefix;
+    String classSource;
+    if (GenInputsAbstract.junit_reflection_allowed) {
+      classSource = jUnitCreator.createSuiteClass(driverName, testMap.keySet());
```
bjkeller: I made this mistake the other day. The methods are different.

src/main/java/randoop/main/GenTests.java
```diff
@@ -788,33 +800,10 @@ private void logOutputStart(int sequenceCount, String testKind) {
...
-      Set<String> testClassNames,
-      JavaFileWriter jfw) {
-    String classSource;
-    String driverName = junitPrefix;
-    if (GenInputsAbstract.junit_reflection_allowed) {
-      classSource = junitCreator.createSuiteClass(driverName, testClassNames);
-    } else {
-      driverName = junitPrefix + "Driver";
-      classSource = junitCreator.createTestDriver(driverName, testClassNames);
+      if (f != null) {
```
bjkeller: For my edification, why is the method responsible for documenting the behavior of a client?

src/main/java/randoop/output/FailingTestFilter.java
bjkeller: The class `randoop.compile.SequenceCompiler` already provides a compiler that will work on source code as a `String`. (This was written for the compilability-predicate.) Unfortunately, some of the mechanics to manage the source text as "files" seems to short-circuit the option of writing the class file to disk.

bjkeller: The `compileTestClass` method has comments, and I've created issue #220

src/main/java/randoop/output/CodeWriter.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Writes the given class using this {@link CodeWriter}. Does not modify the class text.
+   *
+   * @param packageName the package name of the class
+   * @param classname the name of the class
+   * @param sourceLines the lines of the source text of the class to be written, must be compilable
+   * @return the {@code File} object for the Java file written
+   */
+  public File writeUnmodifiedClassLines(String packageName, String classname, String[] sourceLines);
```
mernst: This version does not seem to ever be used.
If you retain it, then I suggest that you make this class an abstract class instead of an interface, so you can provide a default implementation of one of the two `writeUnmodified*` methods in terms of the other one.
Should there also be a version that takes a `Stream` as its input?

src/main/java/randoop/output/JavaFileWriter.java
```diff
@@ -73,6 +106,12 @@ private File getDir(String packageName) {
     return dir;
   }
 
+  /**
+   * Returns the output stream to print to the given file.
+   *
+   * @param file the file to be written to
+   * @return the {@code PrintStream} for writing to the file
+   */
   private static PrintStream createTextOutputStream(File file) {
```
mernst: `PrintStream` seems inappropriate here.
Per https://docs.oracle.com/javase/7/docs/api/java/io/PrintStream.html:

> The PrintWriter class should be used in situations that require writing characters rather than bytes.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,316 @@
...
+   * @param classCode the source code for the test class, each assertion must be on its own line
+   * @param status the {@link RunCommand.Status} for running the test with JUnit
+   * @return the class source edited so that failing assertions are replaced by line comments
+   * @throws BugInRandoopException if {@code status} contains output for a failure not involving a
+   *     Randoop generated test method
+   */
+  private String commentFailingAssertions(
+      String classname, String classCode, RunCommand.Status status) {
+    // JUnit4 writes to standard out, check this doesn't change.
+    assert status.errorOutputLines.isEmpty()
```
mernst: Here and elsewhere, I'm worried about silently ignoring errors when assertions are turned off.
Wouldn't it be better to throw the error unconditionally so that we get bug reports from users and we fix the problems?

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,316 @@
...
+
+    Pair<String, String> failureCountMatch;
+    try {
+      failureCountMatch = readUntilMatch(lineIterator, FAILURE_MESSAGE_PATTERN);
+    } catch (JUnitReaderException e) {
+      throw new BugInRandoopException(
+          "Error reading JUnit output: no match for failure count message");
+    }
+    int totalFailures = Integer.parseInt(failureCountMatch.b);
+    assert totalFailures > 0 : "JUnit has non-zero exit status, but no failure found";
```
mernst: Why isn't this exception thrown unconditionally?

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,316 @@
...
+    /*
+     * First, find the message that indicates the number of failures in the run.
+     */
+
+    Pair<String, String> failureCountMatch;
+    try {
+      failureCountMatch = readUntilMatch(lineIterator, FAILURE_MESSAGE_PATTERN);
+    } catch (JUnitReaderException e) {
+      throw new BugInRandoopException(
+          "Error reading JUnit output: no match for failure count message");
```
mernst: Add `e` as a cause argument.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,277 @@
...
+   * then returns a pair containing the line and the text matching the first group of the pattern.
+   * Assumes that there is a match, and that the pattern has at least one group.
+   *
+   * @param lineIterator the iterator for reading from the JUnit output
+   * @param pattern the pattern for a regex with at least one group
+   * @return the pair containing the line and the text matching the first group
+   * @throws BugInRandoopException if the iterator has no more lines, but the pattern hasn't been
+   *     matched
+   */
+  private Pair<String, String> readUntilMatch(Iterator<String> lineIterator, Pattern pattern) {
```
mernst: The use of fields `a` and `b` makes client code *very* hard to read.
Create a new struct class (it just has to have two public fields with clear names) and use it instead of `Pair`.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,277 @@
...
+        Pattern.compile("\\d+\\)\\s+(" + ID_STRING + ")\\(" + ID_STRING + "\\)");
+
+    /*
+     * Split Java code text so that we can match the line number for the assertion with the code.
+     * Use same line break as used to write test class file.
+     */
+    String[] javaCodeLines = javaCode.split(Globals.lineSep);
+
+    int failureCount = 0;
+    while (failureCount < totalFailures) {
```
mernst: Why isn't this a `for` loop?  That brings all the control structure into 1 statement; currently it's on 2 statements here and on 1 more statement that is 43 lines away from these.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,277 @@
...
+       * number.
+       */
+      Pattern linePattern =
+          Pattern.compile(
+              String.format(
+                  "\\s+at\\s+%s\\.%s\\(%s\\.java:(\\d+)\\)", classname, methodName, classname));
+
+      Pair<String, String> failureLineMatch = readUntilMatch(lineIterator, linePattern);
+      int lineNumber = Integer.parseInt(failureLineMatch.b);
+      if (lineNumber < 1 && lineNumber < javaCodeLines.length + 1) {
```
mernst: This looks wrong.  It should be `||` instead of `&&`, it should be `>` instead of `<`, and why is the `+1` needed?

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -0,0 +1,277 @@
...
+   * directory name.
+   *
+   * @param classname the class name
+   * @param pass the pass count
+   * @return the {@code Path} for the directory created
+   */
+  private Path createWorkingDirectory(String classname, int pass) {
+    try {
+      Path workingDirectory = Files.createTempDirectory("check" + classname + pass);
+      //workingDirectory.toFile().deleteOnExit();
```
mernst: Why is this commented out?  Should it be uncommented before the branch is merged into master?

src/main/java/randoop/execution/RunCommand.java
```diff
@@ -0,0 +1,126 @@
...
+import java.io.IOException;
+import java.util.List;
+import plume.TimeLimitProcess;
+import randoop.util.StreamUtils;
+
+/**
+ * Class providing the {@link #run(List, File, long)} method to run a command in a separate process
+ * with a timeout.
+ */
+public class RunCommand {
```
bjkeller: Just a note for the future: we discussed this and decided it could be done later.

bjkeller:
flaky-filter changes have all been made -- ready for another pass

src/main/java/randoop/main/GenTests.java
```diff
@@ -496,9 +505,79 @@ public boolean handle(String[] args) throws RandoopTextuiException, RandoopInput
...
+      }
+    } else {
+      if (GenInputsAbstract.progressdisplay) {
+        System.out.printf("%nNo " + testKind.toLowerCase() + " tests to output%n");
+      }
+    }
+  }
+
+  /**
+   * Creates a {@code plume.SimpleLog} that writes to standard output if {@code filename} is "-" (a
```
mernst: It looks like there was a bad merge here.  Please remove this text and reinstate the proper Javadoc for this method.

src/main/java/randoop/main/GenTests.java
```diff
@@ -685,144 +764,28 @@ private void handleFlakySequenceException(AbstractGenerator explorer, SequenceEx
...
+    LinkedHashMap<String, CompilationUnit> testMap = new LinkedHashMap<>();
     for (int i = 0; i < sequencePartition.size(); i++) {
       List<ExecutableSequence> partition = sequencePartition.get(i);
-      String testClassName = String.format(classNameFormat, i);
+      String testClassName = junitPrefix + i;
       CompilationUnit classAST =
-          junitCreator.createTestClass(testClassName, methodNamePrefix, partition);
-      testMap.put(testClassName, classAST);
+          junitCreator.createTestClass(testClassName, TEST_METHOD_NAME_PREFIX, partition);
+      if (classAST != null) {
```
mernst: How can this happen?  If it is expected and no error, document why.  Otherwise, throw an error.

bjkeller:
merged on command-line

PR #224, opened 2017-08-01
Condition specification

bjkeller:
This PR introduces classes for handling specifications of conditions on operations.


src/main/java/randoop/condition/ConditionMethodCreator.java
```diff
@@ -0,0 +1,204 @@
...
+
+  /** The name generator used to generate class names. */
+  private static final NameGenerator nameGenerator = new NameGenerator(CONDITION_CLASS_BASENAME);
+
+  /**
+   * Creates the {@code java.lang.reflect.Method} to test the condition in the condition code.
+   *
+   * <p>Generates the Java source for a class with the method and then
+   *
+   * @param signature the signature string for the condition method
```
mernst: Only the package and simple name are used.

src/main/java/randoop/condition/ConditionMethodCreator.java
```diff
@@ -0,0 +1,204 @@
...
+   * @param parameterDeclaration the parameter declaration string, includes parentheses
```
mernst: What is the relationship of `parameterDeclaration` to `signature`?  How is this different than the formal parameters in `signature`?

src/main/java/randoop/condition/ConditionMethodCreator.java
```diff
@@ -0,0 +1,204 @@
...
+   * @param conditionSource the source code for the condition
+   * @param compiler the compiler used to compile the condition class
```
mernst: "used" => "to use"

src/main/java/randoop/condition/RandoopConditionError.java
```diff
@@ -0,0 +1,33 @@
+package randoop.condition;
+
+/**
+ * Exception to represent errors that occur during creation or evaluation of a {@link Condition}.
```
mernst: Are these exceptions all bugs in Randoop, all user errors, or a combination of the two?  Please document.
If it's a combination, I suggest differentiating them so that all user errors are shown to the user and halt Randoop.

src/main/java/randoop/condition/ConditionMethodCreator.java
```diff
@@ -0,0 +1,204 @@
...
+   * @param packageName the package name for the condition class
+   * @param receiverType the declaring class of the method or constructor, used as receiver type if
+   *     {@code shift == 1}
+   * @param parameterTypes the parameter types for the original method
+   * @param shift starting position of {@code parameterTypes} in condition method parameter list,
+   *     either 0 or 1
+   * @return the constructed pre-condition method signature
+   */
+  static RawSignature getPreconditionSignature(
+      String packageName, Class<?> receiverType, Class<?>[] parameterTypes, int shift) {
```
mernst: I think it would be easier to understand the method specification (and less internal checking would be necessary) if this were changed to a boolean.

src/main/java/randoop/condition/ConditionMethodCreator.java
```diff
@@ -0,0 +1,204 @@
...
+   * @param shift starting position of {@code parameterTypes} in condition method parameter list,
+   *     either 0 or 1
+   * @return the constructed post-condition method signature
+   */
+  static RawSignature getPostconditionSignature(
+      String packageName,
+      Class<?> receiverType,
+      Class<?>[] parameterTypes,
+      Class<?> returnType,
+      int shift) {
```
mernst: Change this to a boolean, too.

src/main/java/randoop/condition/NameReplacementMap.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+   * Replace occurrences of names from this map with corresponding replacements.
+   *
+   * @param text the text to search for occurrences of names in this map
+   * @return the text modified by replacing original names with replacement names
+   */
+  String replaceNames(String text) {
+    // make sure that we are replacing from longer to shorter strings to avoid mangled replacement
```
mernst: This doesn't guarantee that there is no mangled replacement.
Consider changing "a"=>"b" and "b"=>"c".
That will change all "a" to "c".
To avoid mangled replacement, you need to find each match to `\b(name1|name2|name3|...)\b` and replace it by whatever it maps to, then proceed to the next one.

src/main/java/randoop/condition/OutcomeTable.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Set;
+import randoop.test.ExpectedExceptionGenerator;
+import randoop.test.ExtendGenerator;
+import randoop.test.InvalidCheckGenerator;
+import randoop.test.PostConditionCheckGenerator;
+import randoop.test.TestCheckGenerator;
+
+/** Records the outcome of checking all of the conditions for a method. */
+public class OutcomeTable {
+  private boolean isEmpty = true;
```
mernst: Please add Javadoc for the next 5 members.

src/main/java/randoop/condition/OutcomeTable.java
```diff
@@ -0,0 +1,99 @@
...
+  public OutcomeTable() {
+    exceptionSets = new ArrayList<>();
+    postConditions = new ArrayList<>();
+  }
+
+  /**
+   * Adds the outcome of checking the conditions of a specification.
+   *
+   * @param preconditionsSatisfied boolean value indicating whether all preconditions satisfied
+   * @param throwsClauses set of exceptions expected in post-state
```
mernst: These are not exceptions, but pairs.

src/main/java/randoop/condition/OutcomeTable.java
```diff
@@ -0,0 +1,99 @@
...
+   * @param postCondition post-condition that must be true in post-state, null if none
```
mernst: They don't have to be true in the post-state in general.  They only have to be true after a normal execution, when no exception is thrown.

src/main/java/randoop/condition/OperationConditions.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Represents all conditions on an operation including pre-, post-, and throws-conditions defined on
+ * this operation and inherited from supertypes. Does not represent the operation itself.
+ *
+ * <p>For an operation that is a method, the {@link OperationConditions} form an arbitrary directed
```
mernst: This paragraph is inconsistent with the previous paragraph.  The previous paragraph indicates that the abstraction represented is the conditions, so how they are represented is an implementation detail.  This suggests differently.  Please make them consistent, so the documentation says how clients should view this code.

src/main/java/randoop/condition/ConditionMethodCreator.java
bjkeller: The signature only carries the types and not the parameter names.

mernst: Thanks for the clarification.  A comment about that would be helpful.

src/main/java/randoop/condition/NameReplacementMap.java
bjkeller: OK, I implemented this, but this solution doesn't work either.

If we have the pattern `\b(a|b)\b` replacing "a" first gives the same bad outcome. The method would have to know to pick "b"=>"c" first. In general, there could be cyclic replacements that could not be avoided.

However, in this application the replacements look like "a"=>"x1", "b"=>"x2". So, it is only a problem if the original identifiers from the condition are of the form "xi", for i some number.

src/main/java/randoop/condition/RandoopConditionError.java
bjkeller: There were a couple throws that would be condition class configuration bugs, I changed these to BugInRandoopException.

src/main/java/randoop/condition/NameReplacementMap.java
mernst: Why doesn't it work?  By "proceed to the next one" I meant proceed to the next match, which appears lexically after the first one.  My proposed algorithm never re-examines any character.

bjkeller: OK, I got it. You have to use substring and concatenation instead of using replace methods.

mernst: Right.  I didn't say that explicitly, but I'm glad it is clear now.

src/main/java/randoop/condition/OperationConditions.java
```diff
@@ -0,0 +1,175 @@
+package randoop.condition;
+
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Represents the collection of all conditions on an operation including pre-, post- and
```
mernst: This comment is not correct:  it isn't just a collection of conditions, but other data structures.
This means that the name of the class needs to be changed (it's misleading to name it Operation**Conditions**).
Also incorrect is the use of "conditions" in the Javadoc `returnConditions`, `throwsConditions`, and `parentList` (and the former two fields should probably be renamed).
The `ConditionPair` class name is not informative; it serves little purpose, so replace it by two classes with descriptive names.

src/main/java/randoop/condition/OutcomeTable.java
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @return true if preconditions of all specifications are unsatisfied, and there are no expected
+   *     exceptions; false, otherwise
+   */
+  public boolean isInvalid() {
+    return !isEmpty && !hasValid && exceptionSets.isEmpty();
+  }
+
+  /**
+   * Constructs the {@link TestCheckGenerator} to apply post call by extending the given generator.
```
mernst: Please make the specification say what the goal or purpose of the method is, from a client point of view.  Saying it creates an object of a particular class is not helpful; a higher-level description is needed.  Here, the goal is to check whether an invocation satisfies the specification.
The phrase "apply post call" is not clear.

src/main/java/randoop/condition/OutcomeTable.java
```diff
@@ -0,0 +1,109 @@
...
+import randoop.test.PostConditionCheckGenerator;
+import randoop.test.TestCheckGenerator;
+
+/** Records the outcome of checking all of the conditions for a method. */
+public class OutcomeTable {
+
+  /** Indicates whether this table is empty. */
+  private boolean isEmpty = true;
+
+  /** Indicates whether this table has a valid entry. */
```
mernst: What is the definition of "valid"?  Uses later in the class suggest that this class does not use a standard definition, which is confusing (and especially confusing when not documented, but it's better to use a standard definition *and* document it).

src/main/java/randoop/condition/OutcomeTable.java
```diff
@@ -0,0 +1,109 @@
...
+  private boolean hasValid = false;
+
+  /** The list of throws clauses for this table. */
```
mernst: This Javadoc comment doesn't say much:  it gives the type, and says it's "for this table", which is obvious.  It would be more helpful to say, for example, that their preconditions were all satisfied.

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * @return true if preconditions of all specifications are unsatisfied, and there are no expected
+   *     exceptions; false, otherwise
+   */
+  public boolean isInvalidPrestate() {
+    return !isEmpty && !hasValid && exceptionSets.isEmpty();
+  }
+
+  /**
+   * Constructs the {@link TestCheckGenerator} to apply post call by extending the given generator.
```
mernst: Please make the specification say what the goal or purpose of the method is, from a client point of view. Saying it creates an object of a particular class is not helpful; a higher-level description is needed. Here, the goal is to check whether an invocation satisfies the specification.
The phrase "apply post call" is not clear.

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.List;
+import java.util.Set;
+import randoop.test.ExpectedExceptionGenerator;
+import randoop.test.ExtendGenerator;
+import randoop.test.InvalidCheckGenerator;
+import randoop.test.PostConditionCheckGenerator;
+import randoop.test.TestCheckGenerator;
+
+/**
+ * Records the outcome of checking all of the preconditions for a method. Each table entry records:
```
mernst: There was only one line of documentation for this class.  There wasn't even a cross-reference to where documentation could be found.  Please document each class.  Do avoid duplication, but it's better to cross-reference from elsewhere to class documentation than to put (partial) class documentation elsewhere where it may not be discovered.

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+ *   <li>The expected postcondition, given the observed values (that is, which return-clauses'
+ *       guards are satisfied)
+ * </ul>
+ */
+public class PreconditionOutcomeTable {
+
+  /** Indicates whether this table is empty. */
+  private boolean isEmpty = true;
+
+  /** Indicates whether this table has a valid entry. */
```
mernst: What is the definition of "valid"? Uses later in the class suggest that this class does not use a standard definition, which is confusing (and especially confusing when not documented, but it's better to use a standard definition and document it).

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+  private boolean hasValid = false;
+
+  /** The list of throws clauses for this table. */
+  private final List<Set<ThrowsClause>> exceptionSets;
```
mernst: This Javadoc comment doesn't say much: it gives the type, and says it's "for this table", which is obvious.  The comment does not give semantics. It would be more helpful to say, for example, that their preconditions were all satisfied.
(This is related to the more general lack of documentation for this class.)

The comment is also incorrect about the type:  it's not a list, but a list of sets.  What is the semantics (such as why it cannot be just a list, and how the different sets are used)?

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /** The list of post-conditions for this table. */
+  private final List<PostCondition> postConditions;
```
mernst: In documentation, normal-case behavior comes first.  (This is the reason that the Javadoc documentation states that `@return` clauses should precede `@throws` clauses.)
Please make the order of the fields, and the way they are used in the class, consistent with this by swapping the order of the fields (and of uses and formal parameters where possible).
Consistency with standards helps to make code easier to read, and it highlights places where the programmer has a good reason for a different order.

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+   * @param preconditionsSatisfied boolean value indicating whether all preconditions are satisfied
+   * @param throwsClauses set of exception type-comment pairs for exceptions expected in post-state
+   * @param postCondition post-condition that must be true in post-state if no exception is thrown,
+   *     null if none
+   */
+  void add(
+      boolean preconditionsSatisfied,
+      Set<ThrowsClause> throwsClauses,
+      PostCondition postCondition) {
+    isEmpty = false;
```
mernst: A method call can set this field even if the method call has no other effect (if the precondition is not satisfied).
So, what is the semantics of this field?  Or, are there undocumented requirements on the method formal parameters that guarantee that every call makes a change?

src/main/java/randoop/condition/PreconditionOutcomeTable.java
```diff
@@ -0,0 +1,138 @@
...
+      hasValid = true;
+    }
+    if (!throwsClauses.isEmpty()) {
+      exceptionSets.add(throwsClauses);
+    }
+  }
+
+  /**
+   * Indicate whether this set of results indicates a definitively invalid pre-state. Occurs when
+   * all preconditions fail and there are no expected exceptions.
```
mernst: This is not the standard definition of an illegal call.  A call is illegal if any precondition fails.  This is independent of whether throws-preconditions or return-guards are satisfied.  A call may be guaranteed to throw an exception but still be legal.  
"invalid" is commonly used as a synonym for "illegal", so it's not good to make a distinction between the two terms.
Clarify what this method (and the associated field) mean, and explicitly compare to illegal calls.
Is this method about "is normal prestate" rather than illegality or invalidity?

src/main/java/randoop/condition/OperationConditions.java
bjkeller: Actually, it does represent the collection of all preconditions, pre-post-condition pairs, and pre-throws-condition pairs.

I've changed the use of "condition" in the text to be more specific, but I'm leaving the class name, because I don't understand what you want it to be.

bjkeller:
I've pushed what I was able to address.

src/main/java/randoop/condition/OperationConditions.java
mernst: There is a class named `randoop/condition/Condition.java`.  It is reasonable for a user to assume that any use of the word "condition" refers to that.  If you want "condition" to refer to something else, then you need to rename that class.  Otherwise, change occurrences of "condition" in text where it is not referring to `Condition`.

Another way to put this is that the Javadoc says:
> Represents the collection of all conditions 

and you agreed that is incorrect and misleading, because your comment itself says:
> it does represent the collection of all preconditions, pre-post-condition pairs, and pre-throws-condition pairs.

(ie, not a **"collection of conditions"**, but a collection of other things like preconditions and pairs).

src/main/java/randoop/condition/SpecificationTranslator.java
```diff
@@ -0,0 +1,398 @@
...
+   * @param executable the method or constructor to which the precondition belongs
+   * @return the {@link RawSignature} for a pre-condition method of {@code executable}
+   */
+  private static RawSignature getPreConditionSignature(Executable executable) {
+    boolean isMethod = executable instanceof Method;
+    Class<?> declaringClass = executable.getDeclaringClass();
+    Class<?>[] parameterTypes = executable.getParameterTypes();
+    String packageName = getPackageName(declaringClass.getPackage());
+    return ConditionMethodCreator.getRawSignature(
+        packageName, declaringClass, parameterTypes, null, isMethod);
```
mernst: The `isMethod` argument passed to the `firstArgumentIsReceiver` formal parameter looks wrong here and for `getPostConditionSignature`.  (The two methods differ by just one line; I would unify them to avoid code duplication.)
An inner class constructor has a receiver.
If the current code is correct, then document why my reaction was incorrect.

src/main/java/randoop/condition/OperationConditions.java
```diff
@@ -0,0 +1,174 @@
...
+  private final List<PreThrowsConditionPair> preThrowsConditionPairs;
+
+  /**
+   * Mirrors the overrides/implements relation among methods. If this OperationConditions is the
+   * local specification for method m, the {@code parentList} contains one element for each method
+   * that m overrides or implements.
+   *
+   * <p>For an operation that is a method, the {@link OperationConditions} form an arbitrary
+   * directed acyclic graph consisting of conditions for methods of supertypes, each of which has
+   * attached specifications.
```
mernst: What is the type of "attached specifications", and where do they appear?

src/main/java/randoop/condition/OperationConditions.java
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Check the pre-conditions for this operation against the arguments. Constructs an {@link
+   * ExpectedOutcomeTable} for this operation, and for this operation in all supertypes.
+   *
+   * @param args the argument values to test the preconditions
+   * @return the table with entries for this operation
+   * @see #check(Object[], ExpectedOutcomeTable)
+   */
+  public ExpectedOutcomeTable check(Object[] args) {
```
mernst: This doesn't check the entire OperationConditions, so I think it should be named `checkPrestate` instead.  You could make the same naming change to other methods in this class that start with `check`.

src/main/java/randoop/condition/OperationConditions.java
```diff
@@ -0,0 +1,174 @@
...
+      List<Condition> preconditions,
+      List<PrePostConditionPair> prePostConditionPairs,
+      List<PreThrowsConditionPair> preThrowsConditionPairs) {
+    this.preconditions = preconditions;
+    this.prePostConditionPairs = prePostConditionPairs;
+    this.preThrowsConditionPairs = preThrowsConditionPairs;
+  }
+
+  /**
+   * Check the pre-conditions for this operation against the arguments. Constructs an {@link
```
mernst: This documentation is confusing.  It says it checks the pre-conditions, but it uses the term "pre-conditions" differently than elsewhere in the same class.  For example, there is a field `preconditions` that is document as "the pre-conditions", but that is *not* what this class checks!  Please change terminology throughout the codebase so that the term "pre-conditions" has only one, clear meaning.  Here, I think it's better to say that all prestate parts of the spec are checked.

src/main/java/randoop/condition/OperationConditions.java
mernst: This doesn't check the entire specification.  I suggest that this and other methods starting with `check` should instead be named `checkPrestate...`.

src/main/java/randoop/condition/OperationConditions.java
mernst: The documentation here is confusing and inconsistent.  The documentation says that it checks the pre-conditions, but it is using that term differently than elsewhere in the class.  For example, there is a field `preconditions` that is documented as "the pre-conditions", and that is *not* what this method checks!
Please change here and throughout the code so that the term "pre-condition" is used unambiguously, with only one meaning.  Here, I think it's better to say that it checks all the prestate parts of the specification.

src/main/java/randoop/condition/Condition.java
```diff
@@ -1,33 +1,107 @@
...
+import java.util.Objects;
+import randoop.util.Log;
+
 /**
- * Represents a condition on an operation. Kind of like a predicate, but with names changed to
- * protect the innocent.
+ * A {@code Condition} represents a Boolean expression.
+ *
+ * <p>Used to represent a pre- or post-condition of an operation where the expression is given as an
+ * expression in Java source code and converted using {@link ConditionMethodCreator}.
```
mernst: Converted into what?

bjkeller:
Could you provide me with the terminology you want to use for the user-provided specifications?

mernst:
Can you be more specific in your question?  "user-provided specifications" seems OK to refer to user-provided specifications.  Or can you provide a proposal?

bjkeller:
If there is a standard nomenclature that you want to adhere to in order to avoid confusing users, it would be useful to know what it is before I start choosing names.

mernst:
The main thing is consistency and having each term mean one thing rather than meaning different things in different contexts.  Given that, I think the code will be much clearer.  Thanks!

src/main/java/randoop/condition/OutcomeTable.java
bjkeller: not `GenInputsAbstract.BehaviorType.INVALID`

src/main/java/randoop/condition/SpecificationTranslator.java
bjkeller: Looking at `ConstructorCall`, I'm pretty sure that reflection gives the receiver for the inner class constructor as part of the parameter types.

bjkeller:
I think I've resolved everything here.

The `condition/package-info` may still be a bit verbose for your tastes. I think it might be a little hard to pull the algorithm together just from documentation in the supporting classes.

src/main/java/randoop/condition/ExpectedOutcomeTable.java
```diff
@@ -0,0 +1,156 @@
...
+ *       post-condition property, if one exists. If any such property fails, then classify as {@link
+ *       randoop.main.GenInputsAbstract.BehaviorType#ERROR}.
+ * </ol>
+ */
+public class ExpectedOutcomeTable {
+
+  /** Indicates whether this table is empty. */
+  private boolean isEmpty = true;
+
+  /** Indicates whether a precondition was satisfied. */
```
mernst: Document how the rows of the table map to these variables; that is, document the correspondence between the abstraction and the concrete state.

src/main/java/randoop/condition/ExpectedOutcomeTable.java
```diff
@@ -0,0 +1,156 @@
...
+  /** Creates an empty {@link ExpectedOutcomeTable}. */
+  public ExpectedOutcomeTable() {
+    postConditions = new ArrayList<>();
+    exceptionSets = new ArrayList<>();
+  }
+
+  /**
+   * Adds the outcome of checking the prestate parts of an operation's specification.
+   *
+   * @param guardIsSatisfied boolean value indicating whether all guard expressions are satisfied
```
mernst: "All guard expressions" is not defined.  Is it only for preconditions, or all guards including for postconditions and throws-conditions?

src/main/java/randoop/condition/ExpectedOutcomeTable.java
```diff
@@ -0,0 +1,156 @@
...
+   * @param propertyExpression property expression that must be true in post-state if no exception
+   *     is thrown
+   * @param throwsClauses set of exception type-comment pairs for exceptions expected in post-state
+   */
+  void add(
+      boolean guardIsSatisfied,
+      PropertyExpression propertyExpression,
+      Set<ThrowsClause> throwsClauses) {
+    // An empty table cannot represent a pre-state for which the call is invalid, so setting isEmpty
```
mernst: This suggests that the `isEmpty` documentation, which states "Indicates whether this table is empty.", is incorrect or at least misleading and needs to be updated.  Please do so there (don't put documentation of the field here, at only one of the places that it is used).  Better yet, is the field necessary?  Could you change `hasSatisfiedGuardExpression` to `hasViolatedPrecondition` and do away with the `isEmpty` variable?

src/main/java/randoop/condition/ExpectedOutcomeTable.java
```diff
@@ -0,0 +1,156 @@
...
+      exceptionSets.add(throwsClauses);
+    }
+  }
+
+  /**
+   * Indicate whether this set of results indicates a definitively invalid pre-state.
+   * (<i>Invalid</i> meaning that the call should be classified as {@link
+   * randoop.main.GenInputsAbstract.BehaviorType#INVALID}.)
+   *
+   * <p>Occurs when all guard expressions fail and there are no expected exceptions.
```
mernst: As elsewhere, I'm confused about "guard expressions", which I think means "preconditions".

src/main/java/randoop/condition/SpecificationTranslator.java
```diff
@@ -0,0 +1,388 @@
...
+   * Creates a {@link SpecificationTranslator} object to translate the {@link
+   * OperationSpecification} of {@code executable}.
+   *
+   * @param executable the {@code java.lang.reflect.AccessibleObject} for the operation with {@link
+   *     OperationSpecification} to translate
+   * @param identifiers the {@link Identifiers} from the specification to be translated
+   * @param compiler the sequence compiler to use to create expression methods
+   * @return the translator object to convert the specifications for {@code executable}
+   */
+  static SpecificationTranslator createTranslator(
```
mernst: Why isn't this a constructor?
Factory methods are usually used to create a value of a different type, to return a possibly-existing value, or to disambiguate if there would be too many constructors with confusingly-similar signatures.  What is the reason this isn't a factory method?

src/main/java/randoop/condition/SpecificationTranslator.java
```diff
@@ -0,0 +1,388 @@
...
+import randoop.condition.specification.Postcondition;
+import randoop.condition.specification.Precondition;
+import randoop.condition.specification.Property;
+import randoop.condition.specification.ThrowsCondition;
+import randoop.main.GenInputsAbstract;
+import randoop.reflection.RawSignature;
+import randoop.types.ClassOrInterfaceType;
+import randoop.util.Log;
+
+/** Translates an {@link OperationSpecification} object to an {@link OperationConditions} object. */
```
mernst: Is a `SpecificationTranslator` only good for one `OperationSpecification`, or can it be re-used?  Please make this comment more explicit.

src/main/java/randoop/condition/SpecificationTranslator.java
```diff
@@ -0,0 +1,388 @@
...
+public class SpecificationTranslator {
+
+  /** The name of dummy variables used by {@link randoop.contract.ObjectContract}. */
+  private static final String DUMMY_VARIABLE_NAME = "x";
+
+  /** The {@link RawSignature} for a guard expression method. */
+  private RawSignature guardExpressionSignature;
```
mernst: This isn't final like all the other instance fields are.  Is it meant to be re-assigned?  Same question for `propertyExpressionSignature`.

mernst: The wording "a" guard expression is confusing.  Which one?  Can it be re-used for multiple expression methods?

src/main/java/randoop/condition/SpecificationTranslator.java
```diff
@@ -0,0 +1,388 @@
...
+
+  /** The parameter declaration string, with parentheses, for a guard expression method. */
+  private final String guardExpressionDeclaration;
```
mernst: State the relationship between the RawSignature and this string.  Does one contain variable names?
(Then comments also apply to the `property*` variables.)

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -152,6 +182,13 @@ public void visit(Constructor<?> constructor) {
     TypedClassOperation operation = instantiateTypes(TypedOperation.forConstructor(constructor));
     checkSubTypes(operation);
     if (!omitPredicate.shouldOmit(operation)) {
+      if (operationSpecifications != null) {
+        OperationConditions conditions =
+            operationSpecifications.getOperationConditions(constructor);
```
mernst: Because the type hierarchy is a DAG, it looks like this may add duplicates to `operationSpecifications` when a superclass is visited more than once.  In a comment, discuss whether this is true, and what the consequences of it are if it is true.

src/main/java/randoop/reflection/OperationExtractor.java
```diff
@@ -152,6 +182,13 @@ public void visit(Constructor<?> constructor) {
...
+        if (!conditions.isEmpty()) {
```
mernst: I think you should do this unconditionally, to avoid expensive re-computation of empty conditions.

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+    this.conditionMap = new HashMap<>();
+    SequenceClassLoader sequenceClassLoader = new SequenceClassLoader(getClass().getClassLoader());
+    List<String> options = new ArrayList<>();
+    this.compiler = new SequenceCompiler(sequenceClassLoader, options);
+  }
+
+  /**
+   * Creates a {@link SpecificationCollection} from the list of files of serialized specifications.
+   *
+   * @param specificationFiles the files of serialized specifications
```
mernst: What format do these have?  How were they created?  What do they represent (eg, all specifications from an entire project, or less than that)?

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+        OperationSignature operation = specification.getOperation();
+
+        // Check for bad input
+        if (operation == null) { // deserialization could result in null operation
+          continue;
+        }
+        if (specification.getIdentifiers().hasDuplicatedName()) {
+          String msg =
+              String.format(
+                  "Ignoring specification with identifier name conflict: %s",
```
mernst: Since this is illegal user input, I would halt Randoop rather than leaving a user to wonder why a spec was ignored (if, as seems likely, the user didn't notice this in the Randoop output.

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+          System.out.println(msg);
+          if (Log.isLoggingOn()) {
+            Log.logLine(msg);
+          }
+          continue;
+        }
+
+        AccessibleObject accessibleObject = getAccessibleObject(operation);
+        specificationMap.put(accessibleObject, specification);
+        if (accessibleObject instanceof Method) {
```
mernst: Comment why constructors are ignored.

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+import randoop.compile.SequenceClassLoader;
+import randoop.compile.SequenceCompiler;
+import randoop.condition.specification.OperationSignature;
+import randoop.condition.specification.OperationSpecification;
+import randoop.reflection.TypeNames;
+import randoop.util.Log;
+import randoop.util.MultiMap;
+
+/**
+ * A collection of {@link OperationSpecification} objects. Also a map from an {@link
```
mernst: How are these two parts of the data structure related?

Also say how this is intended to be used.  Does it represent all the specifications for all methods in an entire project?

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+ * AccessibleObject} reflection object to the {@link OperationConditions} for the corresponding
+ * {@link randoop.operation.TypedClassOperation}.
+ *
+ * <p>The {@link SpecificationCollection} should be constructed from the specification input before
```
mernst: This paragraph is confusing.  It uses a definite modifier in "the specification input" and "the OperationModel", but this is the first mention of those things.

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+ * <p>This class stores the {@link OperationSpecification} objects, and only constructs the
+ * corresponding {@link OperationConditions} on demand. This lazy-strategy avoids building
+ * conditions methods for specifications that are not used.
+ */
+public class SpecificationCollection {
+
+  /** The map from reflection objects to the corresponding {@link OperationSpecification} */
+  private final Map<AccessibleObject, OperationSpecification> specificationMap;
+
+  /** The map from method signatures to methods with that signature and specifications */
```
mernst: What does "and specifications" mean?  Do you mean that the method must have a specification?  Clarify the wording in any event.

src/main/java/randoop/condition/SpecificationCollection.java
```diff
@@ -0,0 +1,312 @@
...
+  private final MultiMap<OperationSignature, Method> signatureMap;
+
+  /** The map from reflection object to overridden method with specification */
+  private final Map<AccessibleObject, Set<Method>> parentMap;
+
+  /** The compiler for creating conditionMethods */
+  private final SequenceCompiler compiler;
+
+  /** Map for memoizing conditions for specifications converted by parent search */
```
mernst: What is parent search?  Is it related to `specificationMap`.
(Later edit: more likely related to `parentMap`?  But, be concrete about this or cross-reference to a place with better documentation.)

src/main/java/randoop/condition/BooleanExpression.java
```diff
@@ -0,0 +1,312 @@
...
+   * @param expressionSource the source code for a Java expression to be used as the body of the
+   *     expression method
+   * @param contractSource a Java expression that is the source code for the expression, in the
+   *     format of {@link BooleanExpression#getContractSource()}
+   * @param comment the comment describing the expression
+   * @param compiler the compiler to used to compile the expression method
+   * @return the {@link BooleanExpression} that evaluates the given expression source on parameters
+   *     described by the declaration string
+   */
+  static BooleanExpression createBooleanExpression(
```
mernst: Why is this a factory method instead of a constructor?  The latter seems clearer, and would be more concise at call sites.

src/main/java/randoop/condition/BooleanExpression.java
```diff
@@ -0,0 +1,312 @@
...
+      String expressionSource,
+      String contractSource,
+      String comment,
+      SequenceCompiler compiler) {
+    Method expressionMethod = createMethod(signature, declarations, expressionSource, compiler);
+    return new BooleanExpression(expressionMethod, comment, contractSource);
+  }
+
+  /**
+   * Returns the {@link BooleanExpression} that checks the expression with the given argument values
```
mernst: I don't understand the semantics and intended use of this method.  Can you write a more complete specification?

As one example of an ambiguity, are the given argument values a subset of all the arguments?  Or must the size of `args` be the same as the current size of the arguments?

src/main/java/randoop/condition/OperationConditions.java
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * The {@link GuardThrowsPair} objects for the {@link
+   * randoop.condition.specification.ThrowsCondition}s of the operation.
+   */
+  private final List<GuardThrowsPair> guardThrowsPairs;
+
+  /**
+   * Mirrors the overrides/implements relation among methods. If this OperationConditions is the
+   * local specification for method declaration m, the {@code parentList} contains one element for
+   * each method that m overrides or implements (and has specifications).
```
mernst: Immediately or transitively overrides/implements?

PR #235, opened 2017-08-18
Add mailing list info on homepage also

thadguidry:
...not just buried down in a help section in the manual...because many folks might want to know more and its a great advertising venue to expand contributions !

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -71,6 +71,8 @@ Randoop continues to be used in industry, for example at ABB corporation.
 
 <p><a href="manual/index.html">Randoop user manual</a></p>
 
+<p>Mailing list: email either randoop-discuss@googlegroups.com or randoop-developers@googlegroups.com with your question.</p>
+
 <p>For Randoop developers/contributors:</p>
 
 <ul>


mernst:
My apologies for not noticing this pull request earlier, and thanks for the suggestion.

Can you explain the rationale for wanting this information on the homepage?  My thinking is that the information is easy to find in the manual, and no one should be sending email until after reading the manual.

thadguidry:
Many times folks want to contribute or ask about technical questions of the architecture, etc.  Remember, your audience in Open Source is not always just users...but contributors and students with free time to fix bugs as well.  Having a quick method of contact information for any Open Source project, front and center, is a best practice Michael.  (that's why you don't and should not bury contact info in a manual)  How many manuals have you read completely end to end BEFORE you started forming questions ? :)

mernst:
Thanks for the background about your suggestion; that is helpful.

I'm surprised that your colleage thought the project might be dead given the commit history at https://github.com/randoop/randoop/commits/master , given the release history at https://github.com/randoop/randoop/releases , and given the release date in the manual at https://randoop.github.io/randoop/manual/ .  I'm not sure how to handle someone who wasn't able to find that information, but I don't think that adding a mailing list link is the right approach.

I'm sorry it took you 7 minutes to find the mailing lists in the manual.  Can you tell me where you looked?  The Randoop manual is the most prominent link from the Randoop homepage at https://randoop.github.io/randoop/, and then searching for "mailing list" in the manual leads you to it right away.

> How many manuals have you read completely end to end BEFORE you started forming questions ? :)

I don't read every single word, but I do know how to open the manual and I always look for information relevant to my question.  No one should ever ask a question without first searching the manual.

Thanks again for the suggestion, and if you have any other feedback about Randoop, please share it.

PR #249, opened 2017-10-12
 Simpler way to specify coverage goals

mernst:
Also avoid using randomVariableForTypeLastStatement which can fail;
this decision needs to be revisited in the future.


mernst:
This branch does more than the pull request says; in particular, it makes a design change.  I'm closing this pull request; a different one could be opened.

PR #252, opened 2017-10-13
Remove redundant, compile-blocking @param from doc

SnowOnion:
The redundant `@param message the error message` in the method doc of constructor `randoop.output.RandoopOutputException(IOException cause)` is regarded as an error by `javadoc`.

It blocks compiling randoop from a clean clone.


mernst:
Thanks for the fix!

PR #281, opened 2017-10-31
add a File utility to delete a directory and its contents

markro49:


mernst:
This change looks fine.
However, rather than creating a new file and method, why not use [UtilMDE.deleteDir](https://types.cs.washington.edu/plume-lib/api/plume/UtilMDE.html#deleteDir-java.io.File-)?

I suggest that version only because Randoop already uses plume-lib; the method also exists in Apache and Google libraries.

markro49:
Sorry - not the familiar with contents of plume-lib.  Happy to make change.  Am I reading code for UtilMDE.deleteDir correctly in that it will fail on nested directories?  Is that intentional?  Even if so, will work fine for this randoop usage.  Also, UtilMDE.deleteDir will exit false on empty directory rather than delete it?

mernst:
It does work correctly on empty directories, but you are right that it's a bug that it only deletes immediate children rather than working recursively.  Thanks for pointing that out!
I have fixed this upstream, but if the current implementation is good enough let's go with it.
(I notice that Guava will soon remove their version of this file because it does not properly detect symbolic links.  That's not an issue for us here.)

markro49:
Could you merge please? (or give me write access as you wish)  Thanks

PR #282, opened 2017-11-01
parser fixes

markro49:
terminate on parse error if debug-checks is set.  my limited testing showed no significant performance cost for --debug-checks.  We could use for in-house testing and default behavior would be unchanged.

correct parse error with embedded quotes (")

### DIFF
--- a/src/main/java/randoop/output/JUnitCreator.java
+++ b/src/main/java/randoop/output/JUnitCreator.java
@@ -46,7 +46,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


markro49:
When we met yesterday, I thought we had decided to have different behavior for our internal testing vs. release.  No problem to change to always throw if that is what you prefer.

mernst:
I misunderstood and thought this was code that would only run during testing.  If it's cheap, I would have it on all the time.  Thanks!

markro49:
do not count timeout exceeded as a compile failure was minor nit I fixed - it has nothing to due with parse errors.

mernst:
Thanks!

PR #290, opened 2017-11-23
Fix randoop error on encountering an unparameterized type that extend…

unshorn:
I have an older Java library that extends java.util.ArrayList<E> resulting in a non-generic type. Randoop bails out on encountering the operation java.util.ArrayList<E>.add(E) complaining that the receiver is not a subtype of java.util.ArrayList<E>

### DIFF
--- a/src/main/java/randoop/reflection/OperationExtractor.java
+++ b/src/main/java/randoop/reflection/OperationExtractor.java
@@ -157,6 +157,11 @@ public class OperationExtractor extends DefaultClassVisitor {
    */
   // TODO: poor name
   private void checkSubTypes(TypedClassOperation operation) {
+
+    if (operation.getDeclaringType().isGeneric() && !classType.hasWildcard()) {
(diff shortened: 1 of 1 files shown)


mernst:
Thanks for the pull request.

Is it possible for you to provide a test case?
That would clarify the issue, and would ensure that future changes do not revert the behavior.

For example, I'm not sure whether you have
```java
class MyClass extends ArrayList { ... }
```
or
```java
class MyClass extends ArrayList<String> {
```

(I think it's the former.)

Also, is your change guaranteed correct, or is it a heuristic that makes Randoop behave better on some code but might degrade Randoop's performance on other code?

mernst:
@unshorn A ping on this pull request.  Can you answer my questions?  Thanks!

unshorn:
Hi Michael,

It indeed is the former. However, the change does break existing assumptions. I will have a closer look and post it as an issue.

Apologies for the delayed response as I was away.

mernst:
Sounds good.  Thanks!

PR #296, opened 2017-12-13
add print-file-system-state option and cleanup error output

markro49:


src/main/java/randoop/main/Main.java
```diff
@@ -80,10 +81,15 @@ public void nonStaticMain(String[] args) {
...
       System.out.println("Randoop failed in an unexpected way.");
-      System.out.println("Please report at https://github.com/randoop/randoop/issues .");
-      e.printStackTrace();
-      System.err.flush();
-      success = false;
+      System.out.println("Please report at https://github.com/randoop/randoop/issues.");
+
+      // Calls to flush() do not untangle System.out and System.err;
+      // probably an OS issue, not Java.  So we do the following
+      // to send printStackTrace() to System.out not System.err.
```
mernst: `printStackTrace` takes an optional `PrintStream` argument, so I think you could do
```
  e.printStackTrace(System.out);
```
instead of the next three lines.
Also, should `System.out` have been flushed instead of `System.err`?

src/main/java/randoop/main/Main.java
```diff
@@ -80,10 +81,15 @@ public void nonStaticMain(String[] args) {
     } catch (BugInRandoopException e) {
       System.out.println();
       System.out.println("Randoop failed in an unexpected way.");
-      System.out.println("Please report at https://github.com/randoop/randoop/issues .");
-      e.printStackTrace();
-      System.err.flush();
-      success = false;
+      System.out.println("Please report at https://github.com/randoop/randoop/issues.");
```
mernst: The space before the period was intentional, so that people can click on the URL without their tools interpreting the "." as part of the URL.

PR #317, opened 2018-03-13
replace TimeLimitProcess with Commons Exec

markro49:
This replaces usage in src/systemtest/java/randoop/main/ProcessStatus.java; there is still one more to go: src/main/java/randoop/main/Minimize.java.

### DIFF
--- a/src/systemtest/java/randoop/main/ProcessStatus.java
+++ b/src/systemtest/java/randoop/main/ProcessStatus.java
@@ -1,15 +1,17 @@
 package randoop.main;
 
-import static org.apache.commons.codec.CharEncoding.UTF_8;
 import static org.junit.Assert.fail;
 
-import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
-import plume.TimeLimitProcess;
+import org.apache.commons.exec.CommandLine;
+import org.apache.commons.exec.DefaultExecuteResultHandler;
+import org.apache.commons.exec.DefaultExecutor;
+import org.apache.commons.exec.ExecuteWatchdog;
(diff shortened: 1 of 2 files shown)


src/systemtest/java/randoop/main/ProcessStatus.java
```diff
@@ -64,48 +66,59 @@ public String dump() {
...
-        new BufferedReader(new InputStreamReader(p.getInputStream(), UTF_8))) {
-      String line = rdr.readLine();
-      while (line != null) {
-        outputLines.add(line);
-        line = rdr.readLine();
+    try {
+      String buf = outStream.toString();
+      // Don't create a list with a single, empty element.
+      if (buf.length() > 0) {
+        outputLines = Arrays.asList(buf.split(lineSep));
```
mernst: If the only use of outputLines is to print it, why split into lines at all?

markro49:
About half the time we just print outputLines and about half the time we parse them.  So we could simplify the former but would need to add something like Arrays.asList(outputLines.split(lineSep)) to each of the latter.  Your call.

mernst:
Would removing the call here entail changes anywhere else?
I was really asking about this particular occurrence.  I'm fine with splitting into lines when parsing, but I wouldn't split into lines everywhere just to make all the code textually consistent.  (Doing so might confuse programmers who wonder why, and incidentally it's unnecessary computation.)

markro49:
I assumed you were suggesting changing the definition of outputLines from List<String> to just String.  In which case all clients need to change, some get simpler, some get slightly more complex.

mernst:
I think we may be talking at cross purposes.
`outputLines` is a local variable in Randoop.  There are no clients of it.

markro49:
but the local outputLines is passed to the ProcessStatus constructor to become the outputLines member of the returned ProcessStatus object.

mernst:
Thanks for the clarification.  I thought I had looked and didn't see any other use, but I obviously overlooked that.  Sorry about that!
Either way is fine with me, and in particular the current way is fine.

PR #340, opened 2018-04-12
Replacecalls constructors

markro49:
replacement of calls to <init> methods are working in my simple test cases.


agent/replacecall/src/main/resources/default-replacements.txt
```diff
@@ -7,9 +7,8 @@ java.lang randoop.mock.java.lang
...
-// Replacements to instance method calls of java.util.Random.
-java.util.Random randoop.mock.java.util.MockRandom
-// Currently, the replacecalls agent cannot replace constructors.
-// This can be reinstated, and the replacements for all the methods of
-// Random removed, when replacecalls can handle constructors.
-// java.util.Random.<init>() randoop.mock.java.util.RandomReplace.randomWithSeedZero()
+// Replacements for java.util.Random.
+// Make Random deterministic by always using a random seed of 0.
+java.util.Random.<init>() randoop.mock.java.util.RandomReplace.randomWithSeedZero()
+java.util.Random.<init>(long) randoop.mock.java.util.RandomReplace.randomWithSeedZero(long)
```
mernst: I suppose you could leave the methods defined and these two lines commented out, with a disclaimer about changing program behavior.  But I wonder whether they are needed at all.

agent/replacecall/src/main/java/randoop/instrument/CallReplacementTransformer.java
```diff
@@ -345,44 +338,112 @@ private InvokeInstruction getReplacementInstruction(
...
+    //
+    // The object created by the new instruction has not been initialized
+    // until ithe constructor completes.  If we simply replace the constructor
+    // call with a call to a replacement method, the JVM will fault on the
+    // uninitialized object reference.  So we must delete the new and dup
+    // instructions and have the replacement constructor supply a new object
+    // reference.
+    //
+    // The following code searches backwards from the invokespecial in the
+    // instruction list until it finds a new/dup pair and then deletes them.
```
mernst: What if one of the arguments to the constructor is itself the result of `new MyOtherClass()`, as in
`new MyClass(new MyOtherClass())`?
Should it search only for new/dup pairs of the correct type?

markro49: I included that case in my test and fixed CallReplaceTransformer to support it.

PR #351, opened 2018-04-30
 Make randomMemberWeighted more efficient

mernst:
This pull request should not affect behavior, only performance.


src/main/java/randoop/util/Randomness.java
```diff
@@ -160,71 +176,88 @@ public static int nextRandomInt(int i) {
...
-      logSelection(
-          index,
-          "randomMemberWeighted(List,Map)",
-          String.format(
-              "%n << %s%n    (class %s),%n    %s%n    (class %s, size %s)>>",
-              list, list.getClass(), weights, weights.getClass(), weights.size()));
+    double currentPoint = 0;
+    for (int i = 0; i < list.size(); i++) {
+      currentPoint += getWeight(list.get(i), weights);
+      if (currentPoint >= chosenPoint) {
```
huangwaylon: Hi Mike,
I'm wondering about the case where we have elements that have weights of value zero.
I was thinking that Randomness should never select an element that has a weight of zero? However, if the list's first element were to have a weight of zero and chosenPoint were to be zero, this code would select that element.

mernst: This is possible.  There is an approximately 1/2^64 chance of this happening.
We could change `>=` to `<` (which, it occurs to me, might be the correct algorithm anyway), or add an explicit check for two zeroes, or ignore the situation.

PR #352, opened 2018-04-30
Fix active flags for observers and add system tests for --observers flag.

caseyxing:
Observers should clear the active flags for all parameters and not just the receiver.
A system test was also added to make sure that observers are not called to try to side-effect any of the inputs.
getClass was omitted due to randoop generating many wildcard tests from .getClass().


caseyxing:
Ready for review.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -229,8 +229,7 @@ private void determineActiveIndices(ExecutableSequence seq) {
       Log.logPrintf("isObserver => %s for %s%n", isObserver, stmt);
       if (isObserver) {
         List<Integer> inputVars = stmts.getInputsAsAbsoluteIndices(i);
-        int receiver = inputVars.get(0);
-        seq.sequence.clearActiveFlag(receiver);
+        seq.sequence.clearAllActiveFlags();
```
mernst: This looks suspicious to me.
I would think that the active flag would stay set for the observer call itself.
Can you confirm that the observer's result is put in the pool and is used in subsequent tests?

caseyxing: The tests generated are:

import org.junit.FixMethodOrder;
import org.junit.Test;
import org.junit.runners.MethodSorters;

@FixMethodOrder(MethodSorters.NAME_ASCENDING)
public class SideEffectObserver0 {

    public static boolean debug = false;

    @Test
    public void test1() throws Throwable {
        if (debug)
            System.out.format("%n%s%n", "SideEffectObserver0.test1");
        observers.Box box0 = new observers.Box();
        int int1 = box0.getB();
        org.junit.Assert.assertTrue(box0.getA() == 2);
        org.junit.Assert.assertTrue(box0.getB() == 3);
        org.junit.Assert.assertTrue("'" + int1 + "' != '" + 3 + "'", int1 == 3);
    }

    @Test
    public void test2() throws Throwable {
        if (debug)
            System.out.format("%n%s%n", "SideEffectObserver0.test2");
        observers.Box box0 = new observers.Box();
        int int1 = box0.getA();
        org.junit.Assert.assertTrue(box0.getA() == 2);
        org.junit.Assert.assertTrue(box0.getB() == 3);
        org.junit.Assert.assertTrue("'" + int1 + "' != '" + 2 + "'", int1 == 2);
    }
}

caseyxing: Hmmm, I agree that it doesn't look right. I'll try and see if those flags are properly set and that the results are used.

mernst:
Merge when tests pass.

PR #379, opened 2018-06-08
Use InputSequenceSelector to select inputs; abstract out small tests heuristic

mernst:


src/main/java/randoop/generation/SmallTestsSequenceSelection.java
```diff
@@ -0,0 +1,50 @@
...
+ * Select input sequences, favoring shorter sequences. This makes Randoop produce smaller JUnit
+ * tests.
+ */
+public class SmallTestsSequenceSelection implements InputSequenceSelector {
+
+  /** Map from a sequence to its weight. */
+  private final Map<Sequence, Double> weightMap = new HashMap<>();
+
+  /**
+   * Pick a sequence from the candidate list using the member's natural weight.
```
huangwaylon: If `Sequence` has been modified so that it no longer implements `WeightedElement` should we use a term other than 'natural weight' since it does not have a getWeight() method anymore?

PR #441, opened 2019-03-03
Non-Multi-Run Deterministic Method Blacklist (+if-idf Categorization of Flaky Methods) and Fully Qualified Array Type Parsing

caseyxing:


src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -341,6 +355,11 @@ private ExecutableSequence createNewUniqueSequence() {
     TypedOperation operation = operationSelector.selectOperation();
     Log.logPrintf("Selected operation: %s%n", operation.toString());
 
+    // TODO: cxing - CR request: Is this the best way to do this?
```
caseyxing: Request for feedback - this is reactive rather than proactive.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +556,60 @@ public boolean handle(String[] args) {
...
+
+      if (flakyTests.size() > 0) {
+        HashMap<TypedOperation, Integer> testOccurrences = new HashMap<>();
+        HashMap<TypedOperation, Integer> flakyOccurrences = new HashMap<>();
+
+        // Tally occurrences of all methods for if-idf
+        // Each method is counted once if it appears in a sequence test.
+        tallyOperationsInSequences(testOccurrences, explorer.getRegressionSequences());
+
+        // TODO: cxing handle Error Test Sequence tallying.
```
caseyxing: We don't currently re-run error revealing tests for flakiness (unless I missed something).

PR #442, opened 2019-03-03
Array type fully qualified signature parsing

caseyxing:


caseyxing:
Thank you for the feedback! I've updated the code and this should be ready for another round of review.

src/main/java/randoop/types/Type.java
```diff
@@ -72,6 +73,82 @@ public static Type forName(@ClassGetName String typeName) throws ClassNotFoundEx
...
+      return fullyQualifiedBaseType;
+    }
+  }
+
+  /**
+   * Returns the Class for a fully qualified name. Does not support arrays.
+   *
+   * @param fullyQualifiedName the fully qualified name of a non-array type
+   * @return the type object for the type with the name, null if none is found
+   * @throws ClassNotFoundException if name is not a recognized type
```
mernst: The above two lines seem to be in conflict.  Does it return null or throw an exception?  Or do both in different situations (which would be a bit confusing)?

mernst:
When you make the last change, I can squash-and-merge this.

PR #443, opened 2019-03-03
Flaky test method categorization with if-idf

caseyxing:


src/main/java/randoop/main/GenTests.java
```diff
@@ -537,6 +597,34 @@ public boolean handle(String[] args) {
     return true;
   }
 
+  /**
+   * Tallies the number of sequences an operation occurs in.
```
mernst: Can you make this comment a bit more descriptive?

src/main/java/randoop/main/GenTests.java
```diff
@@ -537,6 +597,34 @@ public boolean handle(String[] args) {
...
+   *
+   * @param tallyMap the count map to increment for each operation
```
mernst: This, too, can indicate what it maps from and to.

src/main/java/randoop/main/GenTests.java
```diff
@@ -537,6 +597,34 @@ public boolean handle(String[] args) {
...
+   * @param sequences sequences to process for operations
+   */
+  private void tallyOperationsInSequences(
```
mernst: Why does this method take an out-parameter and return void?
Methods are usually clearer if they do not modify their arguments, and I think it would be nicer for this to return a freshly-constructed map, unless there is a reason to prefer an out-parameter.

src/main/java/randoop/main/GenTests.java
```diff
@@ -537,6 +597,34 @@ public boolean handle(String[] args) {
...
+      Map<TypedOperation, Integer> tallyMap, List<ExecutableSequence> sequences) {
+    for (ExecutableSequence es : sequences) {
+      HashSet<TypedOperation> ops = new HashSet<>();
+      SimpleList<Statement> statements = es.sequence.statements;
+      for (int i = 0; i < statements.size(); i++) {
```
mernst: Why is the ops set needed?
Maybe it's to ensure that each operation contributes only once to the result.  That is, maybe the result is the number of tests that the operation appears in, not the number of occurrences of the operation (as I had assumed).
The specification is unclear on what is being computed.  Could you please document that?  And document the `testOccurrences` and `flakyOccurrences` variables too?

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,60 @@ public boolean handle(String[] args) {
...
+        tallyOperationsInSequences(testOccurrences, explorer.getRegressionSequences());
+
+        // TODO: cxing handle Error Test Sequence tallying.
+        // tallyOperationsInSequences(testOccurrences, explorer.getErrorTestSequences());
+
+        List<ExecutableSequence> regressionSequences = explorer.getRegressionSequences();
+        for (String flakyTestNum : flakyTests) {
+          int testNum = Integer.parseInt(flakyTestNum.substring(4)); // length of test
+          // Tests start at 001
+          ExecutableSequence flakyTest = regressionSequences.get(testNum - 1);
```
mernst: The following code seems duplicative with `tallyOperationsInSequences`.  Could you make a list of flaky tests, then pass them into that method to avoid code duplication?  If that is not possible or not desirable, please document why.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,60 @@ public boolean handle(String[] args) {
...
+          for (TypedOperation to : ops) {
+            if (flakyOccurrences.containsKey(to)) {
+              flakyOccurrences.put(to, flakyOccurrences.get(to) + 1);
+            } else {
+              flakyOccurrences.put(to, 1);
+            }
+          }
+        }
+
+        Map<String, Double> methodFlakyPercentage = new HashMap<>();
```
mernst: There is no documentation of what computation is being done.  A few words about that would make the code easier to understand, and then the specification and the implementation can be reviewed independently rather than reverse-engineering the code to determine what it does, then determining whether what it does is desirable.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,60 @@ public boolean handle(String[] args) {
...
+        List<Entry<String, Double>> sortedMethodsbyFlakiness;
```
mernst: Declare a variable at its point of initialization when possible.  You can move this declaration to [line 570](https://github.com/randoop/randoop/pull/443/files#diff-dc322f142726487312b7b1299f931aebR570).  That helps to emphasize to readers that it isn't used in the `for` loop.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,60 @@ public boolean handle(String[] args) {
...
+          methodFlakyPercentage.put(
+              to.toParsableString(), flakyMethodOccurrences / testOccurrences.get(to));
+        }
+
+        sortedMethodsbyFlakiness = new ArrayList<>(methodFlakyPercentage.entrySet());
+        sortedMethodsbyFlakiness.sort(reverseOrder(Entry.comparingByValue()));
+
+        System.out.println();
+        System.out.println("Percentage of method flakiness in tests.");
+        for (Entry<String, Double> e : sortedMethodsbyFlakiness) {
```
mernst: Should this only output the first few methods?  Otherwise it could be a huge amount of output.
Also, I think it needs a clearer explanation than "Percentage of method flakiness".
For example, what should a user do with this information?

caseyxing:
Thank you for the feedback! I've updated the code and this should be ready for another round of review.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -171,6 +171,13 @@ public GenInputsAbstract(
   @Option("What to do if a flaky test is generated")
   public static FlakyTestAction flaky_test_behavior = FlakyTestAction.HALT;
 
+  /**
+   * How many suspected flaky methods to print out (sorted by likelihood) upon flaky test
+   * generation. A flaky test is one that behaves differently on different executions.
+   */
+  @Option("Maximum number of suspected flaky methods to print")
+  public static int num_suspected_flaky_test_to_output = 20;
```
mernst: Should `_test_` be `_method_`?  It's not tests, but methods from user code, that are being output, right?
Also, 20 seems like a large number.  Are the tool's guesses so poor that a large number are usually needed?

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+
+        // Tally occurrences of operations in flaky tests.
+        // Each method is counted exactly once if it appears in a flaky test.
+        flakyOccurrences = tallyOperationsInSequences(flakySequences);
+
+        // if-idf metric
+        // Our heuristic for probability of whether a method is flaky
+        //  is equal to the ratio of the number of flaky tests a method M occurs
+        //  in divided by the number of total tests M occurs in.
+        Map<String, Double> methodFlakyPercentage = new HashMap<>();
```
mernst: Why is this called "percentage" if it's a TF-IDF metric?  What are the keys of the map?

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+
+      if (flakyTests.size() > 0) {
+        List<ExecutableSequence> regressionSequences = explorer.getRegressionSequences();
+
+        // How many tests an operation occurs in.
+        Map<TypedOperation, Integer> testOccurrences;
+
+        // Tally occurrences of operations in all tests.
+        // Each method is counted exactly once if it appears in a sequence test.
+        testOccurrences = tallyOperationsInSequences(regressionSequences);
```
mernst: Combine the variable declaration and initialization into a single statement.

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -67,6 +71,17 @@
   public FailingTestFilter(TestEnvironment testEnvironment, JavaFileWriter javaFileWriter) {
     this.testEnvironment = testEnvironment;
     this.javaFileWriter = javaFileWriter;
+    this.flakyTests = new HashSet<>();
+  }
+
+  /**
+   * Retrieves a shallow copy of the flaky test collection.
```
mernst: This documentation doesn't say what the collection is.  It's a collection of names (as opposed to the full text of the test or something else).

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
           codeWriter,
           GenInputsAbstract.regression_test_basename,
           "Regression");
+
+      HashSet<String> flakyTests = codeWriter.getFlakyTests();
```
mernst: This is a sef of test names, not tests themselves.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+        testOccurrences = tallyOperationsInSequences(regressionSequences);
+
+        // How many flaky tests an operation occurs in.
+        Map<TypedOperation, Integer> flakyOccurrences;
+
+        // TODO: cxing handle Error Test Sequence tallying.
+        // tallyOperationsInSequences(testOccurrences, explorer.getErrorTestSequences());
+
+        List<ExecutableSequence> flakySequences = new ArrayList<>();
+        for (String flakyTestNum : flakyTests) {
```
mernst: This string is a test name.  Why is it called `flakyTestNum`?

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+          int testNum = Integer.parseInt(flakyTestNum.substring(4)); // length of "test"
+          // Tests start at 001
+          ExecutableSequence flakySequence = regressionSequences.get(testNum - 1);
+          flakySequences.add(flakySequence);
+        }
+
+        // Tally occurrences of operations in flaky tests.
+        // Each method is counted exactly once if it appears in a flaky test.
+        flakyOccurrences = tallyOperationsInSequences(flakySequences);
```
mernst: Declare `flakyOccurrences` here, not 15 line above where it is first used.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+
+        System.out.println();
+        System.out.println("Flaky tests were generated. The following methods,");
+        System.out.println("in decreasing order of likelihood, are the most likely culprits.");
+        System.out.println("Please determine whether the following methods may exhibit");
+        System.out.println("different behavior on separate runs of the same code");
+        System.out.println(
+            "and add those methods to the non-multi-run deterministic method blacklist.");
+        System.out.println("If you are unsure, consider adding one method at a time to the");
+        System.out.println("non-multi-run deterministic method blacklist and re-running Randoop");
```
mernst: What is this?  Give a command-line option or a reference into the manual.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+        System.out.println("until no flaky tests are generated.");
+        System.out.println();
+
+        for (int i = 0;
+            i < GenInputsAbstract.num_suspected_flaky_test_to_output
+                && i < sortedMethodsByFlakiness.size();
+            i++) {
+          Entry<String, Double> method = sortedMethodsByFlakiness.get(i);
+          System.out.println(method.getKey() + ":\t " + (method.getValue() * 100.0) + "%");
```
mernst: Writing this as a percentage seems misleading.  It's not the case that a method with a 100% score is 100% likely to be the cause of the nondeterminism.  And it's not a percentage of anything.  I would state it just as a number for the ranking (without "%" or units).
Also, round it or truncate it (casting to int does truncation) since users are unlikely to care about any digits after the decimal point.

src/main/java/randoop/main/GenTests.java
```diff
@@ -537,6 +612,42 @@ public boolean handle(String[] args) {
     return true;
   }
 
+  /**
+   * Counts the number of sequences an operation occurs in. A sequence contributes 1 to the tally of
+   * operation Op if Op appears at least once in the sequence. Further occurrences in the same
+   * sequence (past the initial occurrence) do not contribute to the tally.
+   *
+   * @param sequences sequences to process for operations
+   * @return the output count map to increment for each operation For each operation, the map maps
```
mernst: Missing punctuation

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -67,6 +71,17 @@
...
+   *
+   * @return a shallow copy of the flaky test collection
+   */
+  @SuppressWarnings("unchecked")
+  public HashSet<String> getFlakyTests() {
+    return (HashSet<String>) flakyTests.clone();
```
mernst: Use `return new HashSet<>(flakyTests);` because then you don't need to suppress any warnings.  (It's also slightly shorter, for what that's worth.)

src/main/java/randoop/output/FailingTestFilter.java
```diff
@@ -229,12 +245,18 @@ private void compilationError(
    * @param classname the simple (unqualified) name of the test class
    * @param javaCode the source code for the test class; each assertion must be on its own line
    * @param status the {@link randoop.execution.RunCommand.Status} from running the test with JUnit
+   * @param flakyTests names of flaky tests, e.g., "test005". This is an output parameter that is
+   *     augmented by this methodx
```
mernst: Typo

src/main/java/randoop/main/GenTests.java
mernst: You didn't address this comment.

src/main/java/randoop/main/GenTests.java
mernst: You didn't address this comment.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+          ExecutableSequence flakySequence = regressionSequences.get(testNum - 1);
+          flakySequences.add(flakySequence);
+        }
+
+        // Tally occurrences of operations in flaky tests.
+        // Each method is counted exactly once if it appears in a flaky test.
+        flakyOccurrences = tallyOperationsInSequences(flakySequences);
+
+        // tf-idf metric
+        // Our heuristic for probability of whether a method is flaky
```
mernst: This is not a probability.  (It is a percentage, but it is used purely for ranking.)

src/main/java/randoop/main/GenTests.java
mernst: Putting values in this map, only to sort them later, feels like a waste.
You could use a priority queue instead, and then you wouldn't have to sort later.
(I would suggest a TreeMap from Double to String, but you want to permit two names to have the same ranking without losing them.)

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+        //  in divided by the number of total tests M occurs in.
+        // The result is multiplied by 100.0 to obtain a percentage.
+
+        // Maps from a fully-qualified method signature to the above
+        //  percentage.
+        Map<String, Double> methodFlakyPercentage = new HashMap<>();
+
+        for (TypedOperation to : testOccurrences.keySet()) {
+          double flakyMethodOccurrences =
+              flakyOccurrences.containsKey(to) ? flakyOccurrences.get(to) : 0;
```
mernst: If the ranking is 0, then there is no need to even put the method in the ranking.  Printing it out as a potentially nondeterministic method is misleading.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+              to.toParsableString(), flakyMethodOccurrences / testOccurrences.get(to));
+        }
+
+        // Sort in descending order by value.
+        List<Entry<String, Double>> sortedMethodsByFlakiness =
+            new ArrayList<>(methodFlakyPercentage.entrySet());
+        sortedMethodsByFlakiness.sort(reverseOrder(Entry.comparingByValue()));
+
+        System.out.println();
+        System.out.println("Flaky tests were generated. The following methods,");
```
mernst: This speaks about a culprit; it would be helpful to give a simple explanation such as "This means that your program contains side-effecting or nondeterministic methods."
Rather than "Please determine", give guidance about what to do.  I think it is adequate to give a link to the manual, such as https://randoop.github.io/randoop/manual/index.html#nondeterminism .  Be sure the manual is adequate.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+        System.out.println("in decreasing order of likelihood, are the most likely culprits.");
+        System.out.println("Please determine whether the following methods may exhibit");
+        System.out.println("different behavior on separate runs of the same code.");
+        // TODO cxing: add nmrd-blacklist comment suggestion for user.
+        System.out.println();
+
+        // Output methods by ranking from most likely to least likely to be flaky.
+        for (int i = 0;
+            i < GenInputsAbstract.nondeterministic_methods_to_output
```
mernst: This would be clearer if you put the max in a variable.  Or you just set `nondeterministic_methods_to_output`, if it is guaranteed not to be used or output again.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,75 @@ public boolean handle(String[] args) {
...
+                && i < sortedMethodsByFlakiness.size();
+            i++) {
+          Entry<String, Double> method = sortedMethodsByFlakiness.get(i);
```
mernst: It's misleading to have a variable name `method` that is not a method nor a method name.

src/main/java/randoop/main/GenTests.java
```diff
@@ -537,6 +612,42 @@ public boolean handle(String[] args) {
...
+   * @param sequences sequences to process for operations
+   * @return the output count map to increment for each operation. For each operation, the map maps
+   *     to the number of sequences in which the operation occurs at least once.
+   */
+  private Map<TypedOperation, Integer> tallyOperationsInSequences(
+      List<ExecutableSequence> sequences) {
+    Map<TypedOperation, Integer> tallyMap = new HashMap<>();
+
+    for (ExecutableSequence es : sequences) {
+      // Ensure that each operation is only counted once per sequence.
```
mernst: I suggest abstracting out the first half of the body of this loop into a method named something like `operationsInSequence`.  That will make the rest of the logic clearer.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,71 @@ public boolean handle(String[] args) {
...
+
+      if (flakyTestNames.size() > 0) {
+        List<ExecutableSequence> regressionSequences = explorer.getRegressionSequences();
+
+        // How many tests an operation occurs in (regardless of how many times it appears in that
+        // test).
+        Map<TypedOperation, Integer> testOccurrences =
+            tallyOperationsInSequences(regressionSequences);
+
+        // TODO: cxing handle Error Test Sequence tallying.
```
mernst: This TODO should be addressed.

src/main/java/randoop/main/GenTests.java
```diff
@@ -516,6 +522,71 @@ public boolean handle(String[] args) {
...
+        // TODO cxing: add nmrd-blacklist comment suggestion for user.
+        System.out.println();
+
+        // Output methods by ranking from most likely to least likely to be flaky.
+        for (int i = 0;
+            i < GenInputsAbstract.nondeterministic_methods_to_output
+                && i < sortedMethodsByFlakiness.size();
+            i++) {
+          Entry<String, Double> method = sortedMethodsByFlakiness.get(i);
+          System.out.println((i + 1) + ".\t" + method.getKey());
```
mernst: There is no need to print the index.  Just print the method name.

src/main/java/randoop/main/GenTests.java
caseyxing: At present, error revealing tests aren't reran after they are generated (and thus we can't determine whether they are flaky). This might be a substantial change - should it go in another pull request? I should definitely indicate that in the comment at least.

mernst: I see.  Just expanding the comment here is adequate.  Thanks!

src/main/java/randoop/main/GenTests.java
```diff
@@ -159,6 +165,32 @@
   /** The count of sequences that failed to compile. */
   private int sequenceCompileFailureCount = 0;
 
+  /** TypedOperationHeuristic is a wrapper around a TypedOperation and a heuristic. */
+  private static class TypedOperationHeuristic {
```
caseyxing: This seems like the stylistically correct Java way to create a generic type for a PriorityQueue, but it seems a bit heavy-handed.

caseyxing:
Needs changes - working out issues with testNum corresponding with sequence when uncompilable sequences are present.

caseyxing:
Test number mismatch was fixed upstream. This should be ready for review again.

src/main/java/randoop/main/GenTests.java
```diff
@@ -510,7 +517,69 @@ public boolean handle(String[] args) {
           codeWriter,
           GenInputsAbstract.regression_test_basename,
           "Regression");
-    }
+
+      Set<String> flakyTestNames = codeWriter.getFlakyTestNames();
```
mernst: Can you abstract this out into a method?
Method handle() is already obscenely long, but let's not make it worse.

src/main/java/randoop/main/GenTests.java
```diff
@@ -510,7 +517,69 @@ public boolean handle(String[] args) {
...
+            System.out.println("  " + rankedMethod.operation.toParsableString());
+          }
+        }
+
+        System.out.println(
+            "To prevent the generation of flaky tests, see section 'Nondeterministic program");
+        System.out.println(
+            "under test' at https://randoop.github.io/randoop/manual/#nondeterminism .");
+        System.out.println();
+        // TODO cxing: add nmrd-blacklist comment suggestion for user and edit the manual
```
mernst: What needs to be done here?  Or is this related to a later pull request?

PR #454, opened 2019-03-28
Rename source directories to match Gradle source sets.

smillst:
coveredtest -> coveredTest
replacecalltest -> replacecallTest
systemtest -> systemTest
testinput -> testInput


mernst:
Thanks!

PR #487, opened 2019-05-04
don't break the build in case of htmlSanityCheck errors

gernotstarke:
just a suggestion: if you (@mernst) don't want to break the build in case of htmlSanityCheck errors, use this configuration...

best regards, Gernot

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -865,7 +865,7 @@ htmlSanityCheck {
     sourceDocuments = fileTree(sourceDir) {
         include "index.html", "projectideas.html", "publications.html", "manual/dev.html", "manual/index.html"
     }
-    failOnErrors = true
+    failOnErrors = false //false means: don't break the build in case of errors
 }
 
 build.dependsOn manual


mernst:
Gernot, thanks for the suggestion.

Because I value the feedback that htmlSanityCheck is providing, I don't want to accidentally ignore it, and I don't want to have to visually scan the output (which is tedious and error-prone, and I could forget to do so).

I could imagine having two different tasks in my buildfile -- one each with `failOnErrors` as true and false -- but I'm not sure how to do that.  I would usually do something like the following, but it does not work with `htmlSanityCheck` and I'm not enough of a Gradle expert to see how to make it work.

```
task htmlSanityCheckNoFail(type: htmlSanityCheck) {
    // sourceDir = new File( "${projectDir}/src/docs/manual" )
    sourceDir = new File( "${projectDir}/src/docs" )
    sourceDocuments = fileTree(sourceDir) {
        include "index.html", "projectideas.html", "publications.html", "manual/dev.html", "manual/index.html"
    }
    failOnErrors = false
}
```

mernst:
I'm going to close this because it requires manual examination of the output and thus is not appropriate for continuous integration and for git hooks.

PR #500, opened 2019-05-15
Support for assertions that call side-effect-free methods

caseyxing:


caseyxing:
Wondering what a good system test would be for this (if desired?).

src/main/java/randoop/main/GenTests.java
```diff
@@ -1074,13 +1154,15 @@ private void printSequenceExceptionError(AbstractGenerator explorer, SequenceExc
    *
    * @param visibility the visibility predicate
    * @param contracts the contract checks
-   * @param observerMap the map from types to observer methods
+   * @param omitMethodsPredicate the predicate for filtering out omit methods
+   * @param sideEffectFreeMethods the map from types to side-effect-free methods
```
mernst: This name `sideEffectFreeMethods` is better than `sideEffectFreeMap` which is used elsewhere.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -297,24 +303,56 @@ public void addClassLiterals(
...
         String message = String.format("Error while reading file %s: %s%n", file, e.getMessage());
         throw new RandoopUsageError(message, e);
       }
     }
-    return observerMap;
+    return new MultiMap<>();
+  }
+
+  /**
+   * Returns operations read from the given EntryReader.
```
mernst: In what format?  This should be explained in the Javadoc of all three `readOperations` methods.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -297,24 +303,56 @@ public void addClassLiterals(
...
+   *
+   * @param er the EntryReader to read from.
+   * @param onlyMethods if true, throw an exception if a constructor is read
+   * @return contents of the file, as a map of operations
```
mernst: I would document that this may throw an IOException.
Though, could you fold the RandoopUsageError into this in order to reduce code in both of the clients?

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -297,24 +303,56 @@ public void addClassLiterals(
...
+   */
+  public static MultiMap<Type, TypedClassOperation> readOperations(
```
mernst: Why is this public?  I think it is only called twice in this class, unless I missed a call.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -297,24 +303,56 @@ public void addClassLiterals(
...
+
+  /**
+   * Returns operations read from the given stream.
+   *
+   * @param is the stream from which to read
+   * @param filename the file name to use in diagnostic messages
+   * @param onlyMethods if true, throw an exception if a constructor is read
+   * @return contents of the file, as a map of operations
+   */
+  public static MultiMap<Type, TypedClassOperation> readOperationsFromStream(
```
mernst: The other two versions are called `readOperations`.  It's inconsistent that just this one has a distinct name.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -187,21 +196,27 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
...
-              for (TypedOperation m : observers) {
-
-                // When outputting checks, ignore observers that don't take a single argument.
+            Set<TypedClassOperation> sideEffectFreeMethods =
+                sideEffectFreeMap.getValues(var0.getType());
+            if (sideEffectFreeMethods != null) {
+              for (TypedClassOperation m : sideEffectFreeMethods) {
+                // When outputting checks, ignore side-effect-free methods that don't take a single
+                // argument.
                 if (m.getInputTypes().size() != 1) {
```
mernst: There is code duplication with code in GenTests.  Abstract it out for clarity, brevity, and to reduce the chance of errors.

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -89,83 +99,84 @@ public static void setupClass() {
...
-   *
-   *     By default, coverage is checked against all methods returned by Class.getDeclaredMethods()
-   *     for an input class. Some tests need to specifically exclude methods that Randoop should not
-   *     generate, or need to ignore methods.  These can be indicated by creating a
-   *     CoverageChecker object and adding these method names using either the exclude() or ignore()
-   *     methods, and then giving the CoverageChecker as the last argument to the alternate version
-   *     of generateAndTest(). When excluded methods are given, these methods may not be
-   *     covered, and, unless ignored, any method not excluded is expected to be covered.
-   */
+  * WRITING TEST METHODS:
```
mernst: Why is there a huge diff here?

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -89,83 +99,84 @@ public static void setupClass() {
...
+  *    of any test already in this class).
+  *    The variable systemTestEnvironmentManager refers to the global environment for a run of the
+  *    system tests, and contains information about the classpath, and directories needed while
+  *    running all of the system tests.
+  *
+  * 2. Set the options for Randoop.
+  *
+  *    The method that executes Randoop takes the command-line arguments as a RandoopOptions object,
+  *    which can be constructed by the line
+  *      RandoopOptions options = RandoopOptions.createOptions(testEnvironment);
```
mernst: I suggest that you create a single method that encapsulates both of these lines.  That will make clients a bit shorter, and more importantly it will make them less error-prone.

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -71,6 +71,16 @@ public static void setupClass() {
         SystemTestEnvironmentManager.createSystemTestEnvironmentManager(classpath, buildDir);
   }
 
+  /**
+   * Sets common system test command line options.
+   *
+   * @param options RandoopOptions to set
+   */
+  public static void setSystemTestOptions(RandoopOptions options) {
+    options.setOption("side-effect-free-methods", "resources/systemTest/JDK-sfe-methods.txt");
```
mernst: Aren't these the default?  Why do they have to be specified?  And why are they in resources/systemTest/ rather than resources/ ?

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -394,18 +409,25 @@ public void runJDKTest() {
             options,
             "java7.util7.ArrayList.addAll(int, java7.util7.Collection) ignore",
             "java7.util7.ArrayList.addAll(java7.util7.Collection) ignore",
-            "java7.util7.ArrayList.fastRemove(int) exclude",
+            "java7.util7.ArrayList.fastRemove(int) ignore",
```
mernst: Try not to change to "ignore".  Doing so makes the tests weaker, and makes it harder for subsequent programmers to see whether they have improved the situation or not.  Deleting the line makes it required.  (Is that a bad idea in this case?)

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -602,10 +627,12 @@ public void runContractsTest() {
     SystemTestEnvironment testEnvironment =
         systemTestEnvironmentManager.createTestEnvironment("contracts-test"); // temp directory
     RandoopOptions options = RandoopOptions.createOptions(testEnvironment);
+    setSystemTestOptions(options);
     options.setErrorBasename("BuggyTest");
 
     options.setFlag("no-regression-tests");
     options.setOption("generated_limit", "1000");
+    setSystemTestOptions(options);
```
mernst: This appeared 5 lines earlier, so this one is probably a bug.

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -617,6 +644,7 @@ public void runContractsTest() {
         new CoverageChecker(
             options,
             "examples.Buggy.BuggyCompareToTransitive.getTwo() ignore",
+            "examples.Buggy.BuggyEqualsTransitive.getTwo() ignore",
```
mernst: I'm concerned about this change.  Do you understand the reason for it?

src/systemTest/java/randoop/main/RandoopSystemTest.java
caseyxing: These are the system-test specific ones (java7.util7.*). Perhaps a better file name is warranted.

mernst: Thanks for the explanation.  Changing `JDK` to `JDK7` in the file name would be adequate, I think.

src/main/java/randoop/main/GenTests.java
```diff
@@ -567,26 +586,65 @@ public boolean handle(String[] args) {
...
 
+    // Filter out methods that should be omitted.
+    MultiMap<Type, TypedClassOperation> filteredSideEffectFreeMap = new MultiMap<>();
+    for (Type t : sideEffectFreeMap.keySet()) {
+      Set<TypedClassOperation> typeOperations = sideEffectFreeMap.getValues(t);
+      for (TypedClassOperation tco : typeOperations) {
+        if (omitMethodsPredicate.shouldOmit(tco)) {
+          continue;
+        }
+        if (tco.getInputTypes().size() == 1) {
```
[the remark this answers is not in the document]
mernst: I was thinking of creating an `isObserver` method that returns true if:
 * the method has one formal paramter (which might be the receiver, or might be a non-receiver for a static method),
 * has a return type of primitive, boxed primitive, String, or an enum

src/coveredTest/java/randoop/instrument/CoveredClassTest.java
```diff
@@ -224,26 +226,34 @@ private ForwardGenerator getGeneratorForTest() {
...
     }
-    assert observerMap != null;
-    Set<TypedOperation> observers = new LinkedHashSet<>();
-    for (Type keyType : observerMap.keySet()) {
-      observers.addAll(observerMap.getValues(keyType));
+    sideEffectFreeMap.addAll(sideEffectFreeJDKMap);
+    sideEffectFreeMap.addAll(sideEffectFreeUserMap);
+
+    Set<TypedOperation> sideEffectFreeMethodSet = new LinkedHashSet<>();
+    for (Type keyType : sideEffectFreeMap.keySet()) {
```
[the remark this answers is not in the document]
caseyxing: MultiMap doesn't seem to contain a values() method - Sorry for the bad variable name!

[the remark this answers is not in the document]
mernst: OK, thanks for letting me know.  At some point, we should get rid of Randoop's homegrown MultiMap and use Guava's Multimap instead.  That is not urgent, though.

src/main/java/randoop/main/GenTests.java
```diff
@@ -1074,13 +1149,15 @@ private void printSequenceExceptionError(AbstractGenerator explorer, SequenceExc
    *
    * @param visibility the visibility predicate
    * @param contracts the contract checks
-   * @param sideEffectFreeMap the map from types to side-effect-free methods
+   * @param omitMethodsPredicate the predicate for filtering out omit methods
```
mernst: This is another example of a content-free comment.  "filtering out" and "omitting" are the same thing, so you don't need both here.  But more seriously, the comment content is the same as the variable name, which is the same as the variable's type.  Describe what it stands for, where it came from, or how it is used rather than repeating information that is already present in the type.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -93,6 +94,9 @@
   /** For debugging only. */
   private List<Pattern> omitMethods;
 
+  /** Omit methods predicate. */
```
mernst: This comment is content-free.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -162,6 +166,8 @@ public static OperationModel createModel(
 
     OmitMethodsPredicate omitPredicate = new OmitMethodsPredicate(omitMethods);
 
+    model.omitMethodsPredicate = omitPredicate;
+
     model.addOperationsFromClasses(
         model.classTypes, visibility, reflectionPredicate, omitPredicate, operationSpecifications);
```
mernst: The model now contains a field with this object in it.  So, does this need to be passed in any more?

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -290,31 +296,65 @@ public void addClassLiterals(
...
-          TypedClassOperation operation =
-              signatureToOperation(
-                  sig, VisibilityPredicate.IS_ANY, new EverythingAllowedPredicate());
-          sideEffectFreeMap.add(operation.getDeclaringType(), operation);
-        }
+        return OperationModel.readOperations(er);
+
       } catch (IOException e) {
         String message = String.format("Error while reading file %s: %s%n", file, e.getMessage());
         throw new RandoopUsageError(message, e);
```
mernst: Could this be a bug in Randoop, if it is reading a built-in file?

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -365,6 +405,15 @@ public ContractSet getContracts() {
     return contracts;
   }
 
+  /**
+   * Returns the omit methods predicate associated with this operation model.
```
mernst: Here is another place that I didn't find the comments helpful.  Are these methods that the user has specified should not be called?  If so, say that.  Otherwise, say what it is.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -43,30 +46,38 @@
   private ExpectedExceptionCheckGen exceptionExpectation;
 
   /** The map from a type to the set of side-effect-free operations for the type. */
-  private MultiMap<Type, TypedOperation> sideEffectFreeMap;
+  private MultiMap<Type, TypedClassOperation> sideEffectFreeMethodsByType;
 
   /** The visibility predicate. */
   private final VisibilityPredicate isVisible;
 
+  /** The omit methods predicate to filter side effect method assertions. */
```
mernst: Should "side effect" be "side-effect-free"?
Does "filter" mean to "remove" or to "retain"?  Avoid the ambiguous term "filter".
If this always filters side-effect-free methods, why not use a test for whether a method is side-effect-free?  (Or do you mean that this filtering is *applied* to the side-effect-free methods?  That is not clear.)

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -43,30 +46,38 @@
...
   /**
    * Create a RegressionCaptureGenerator.
    *
    * @param exceptionExpectation the generator for expected exceptions
-   * @param sideEffectFreeMap the map from a type to the side-effect-free operations for the type
+   * @param sideEffectFreeMethodsByType the map from a type to the side-effect-free operations for
+   *     the type
    * @param isVisible the visibility predicate
    * @param includeAssertions whether to include regression assertions
+   * @param omitMethodsPredicate the omit methods predicate used to filter {@code
```
mernst: Again, don't use the ambiguous term "filter" (throughout).

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -253,4 +277,44 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
...
+   * Returns true if the method:
+   *
+   * <ul>
+   *   <li>has one formal parameter
+   *   <li>has a return type of primitive, boxed primitive, String, or enum
+   *   <li>is side-effect-free
+   * </ul>
+   *
+   * @param tco side-effect-free method as a TypedClassOperation
+   * @return whether we can use this method as a side-effect-free assertion
```
mernst: This is better than the main method comment above.  Make the two consistent.
In several places the code uses "as an assertion", but you really mean "in an assertion".  For a method (call) to be as an assertion, its return type would have to be boolean.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -253,4 +277,44 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
...
+  /**
+   * Returns true if the method:
+   *
+   * <ul>
+   *   <li>has one formal parameter
+   *   <li>has a return type of primitive, boxed primitive, String, or enum
+   *   <li>is side-effect-free
+   * </ul>
+   *
+   * @param tco side-effect-free method as a TypedClassOperation
```
mernst: Comments should not repeat the type, which is already apparent.  So, cut "as a TypedClassOperation".

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -253,4 +277,44 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
...
     return checks;
   }
+
+  /**
+   * Returns true if the method:
+   *
+   * <ul>
+   *   <li>has one formal parameter
+   *   <li>has a return type of primitive, boxed primitive, String, or enum
+   *   <li>is side-effect-free
```
mernst: This does not seem to be true; at least, I don't see an implementation of this in the method body.  The method seems to already assume that any routine passed to it will be side-effect-free.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -253,4 +277,44 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
     }
     return checks;
   }
+
+  /**
+   * Returns true if the method:
+   *
+   * <ul>
+   *   <li>has one formal parameter
```
mernst: These implementation details can be converted to // comments in the method body.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -253,4 +277,44 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
     }
     return checks;
   }
+
+  /**
+   * Returns true if the method:
```
mernst: A comment should first say what a method means or is intended to do.
This comment dives right into implementation details.  From that, it is difficult to understand the intention.

Once you write a proper specification for the method (which you should write based on what clients will find useful, not its current implementation), you will have a better feel regarding whether more tests, like a visibility test, should be folded into this method.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -253,4 +277,44 @@ public RegressionChecks generateTestChecks(ExecutableSequence sequence) {
...
+      return false;
+    }
+
+    if (tco.getOutputType().isVoid()) {
+      return false;
+    }
+
+    Class<?> outputClass = tco.getOutputType().getRuntimeClass();
+
+    if (outputClass == null) {
```
mernst: How can this be true?  You already tested that the output type is not void.

mernst:
One other comment:
Try to make each pull request be a single change.
Your changes treat  enum constants as literals, which is great.
If that change is made separately, it is easier to review and it's easier to understand the impact of the change.
I abstracted this one out and made it separately, but be on the lookout for others.
Thanks!

src/main/java/randoop/test/RegressionCaptureGenerator.java
caseyxing: I was wondering about that as well - this if statement does actually catch a few cases (which is why I added it). From the documentation it looks like the outputType could be a NullReferenceType and calling getRuntimeClass() on it could return null while still not have isVoid() be true? I'm not sure how exactly it is generated though.

src/systemTest/java/randoop/main/RandoopSystemTest.java
caseyxing: We decided in a call that this was most likely an outcome of having deterministic random testing.

src/main/java/randoop/reflection/OperationModel.java
caseyxing: Now checked in caller in readSideEffectFreeMethods().

src/main/java/randoop/test/RegressionCaptureGenerator.java
mernst: OK, thanks for adding the comment to explain.

mernst:
Thanks for this feature!

PR #513, opened 2019-06-25
changes to support Java 9

markro49:


src/main/java/randoop/reflection/RawSignature.java
```diff
@@ -34,7 +34,7 @@
   /**
    * Create a {@link RawSignature} object with the name and parameterTypes.
    *
-   * @param packageName the package name of the class; null for the unnamed package
+   * @param packageName the package name of the class
```
mernst: I would still document what is used for the unnamed package (even if it differs in Java 8 and 9).

src/main/java/randoop/reflection/RawSignature.java
```diff
@@ -44,7 +44,9 @@ public RawSignature(
       String classname,
       String name,
       Class<?>[] parameterTypes) {
-    assert !Objects.equals(packageName, "");
+    // In Java 8 the unnamed package name is null.
+    // In Java 9 the unnamed package name is "".
+    // assert !Objects.equals(packageName, "");
```
mernst: I would delete this rather than commenting it out, and move the documentation to the method header.
You might also leave it in, but java-version-specific.

build.gradle
```diff
@@ -417,7 +417,7 @@ task replacecallTest(type: Test, dependsOn: 'copyJars') {
   testClassesDirs = sourceSets.replacecallTest.output.classesDirs
   classpath = sourceSets.replacecallTest.runtimeClasspath + files("$buildDir/libs/replacecall-${version}.jar")
   // use the replacecall agent using the exclusions file from replacecallTest/resources
-  jvmArgs "-javaagent:$buildDir/libs/replacecall-${version}.jar=--dont-transform=replacecall-exclusions.txt,--debug"
+  jvmArgs "-javaagent:$buildDir/libs/replacecall-${version}.jar=--dont-transform=replacecall-exclusions.txt"
```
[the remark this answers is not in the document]
markro49: It was unneeded as far as running the test was concerned and I added additional debugging output that made the build log much more verbose.  If you prefer, we can put it back.

PR #531, opened 2019-07-06
Human Readable Flaky Method Diagnostics

caseyxing:


src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -1063,7 +1064,7 @@ public void runFlakyTest() {
             "flaky.FlakyClass.multiply(int, int) include");
 
     List<String> expectedFlakyMethodsInOrder = new ArrayList<>();
-    expectedFlakyMethodsInOrder.add("flaky.FlakyClass.flakyDefaultHashCode()");
+    expectedFlakyMethodsInOrder.add("flaky.FlakyClass.flakyDefaultHashCode(flaky.FlakyClass)");
```
mernst: This change doesn't seem right to me.
The method `flakyDefaultHashCode` takes no arguments.  Nowhere else does Randoop require explicitly giving the receiver type.  I don't think that the receiver type should be listed in this output.

src/main/java/randoop/operation/TypedClassOperation.java
```diff
@@ -158,6 +159,42 @@ public boolean hasWildcardTypes() {
     return new ArrayList<>(paramSet);
   }
 
+  /**
+   * Returns the fully-qualified signature for this operation if it is a method or constructor call.
+   *
+   * @return this operation's fully qualified signature if it is a method or constructor call, null
+   *     otherwise.
+   */
+  public String getFullyQualifiedSignature() {
```
mernst: I see a lot of similarities between the code of this method and that of getRawSignature which appears directly below it.
Could you use the RawSignature object to avoid some code duplication (at the cost of slightly less efficient creation of the string, but that isn't a huge deal)?

caseyxing: Hmm it's not immediately obvious to me how we should proceed (only four lines are duplicated and each field with the exception of classPackaged is referenced somewhere following the four lines).

PR #629, opened 2019-11-20
WIP: Method list gen

caseyxing:
CR purposes only.

Documentation and javadoc WIP.


src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,173 @@
...
+
+      // Sort in alphabetical order
+      Collections.sort(nonDetMethods);
+      Collections.sort(sideEffectFreeMethods);
+
+      // Write out the captured methods.
+      System.out.println("Nondeterministic methods count: " + nonDetMethods.size());
+      if (nonDetMethods.size() > 0) {
+        try (BufferedWriter nonDetMethodWriter =
+            Files.newBufferedWriter(nonDetFile.toFile().toPath(), UTF_8)) {
```
mernst: nonDetFile is already a Path.  Why convert it to a file and then back to a Path?

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,173 @@
...
+          for (String method : nonDetMethods) {
+            nonDetMethodWriter.write("^" + method.replace(".", "\\.").replace("(", "\\(") + "$");
```
mernst: Why isn't close parenthesis quoted?
Should this use Pattern.quote?

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,173 @@
...
+    return annotatedMethods;
+  }
+
+  private static String getFullyQualifiedName(AClass aclass, String JVMLmethodSignature) {
+    // Parse out the arguments to feed into plume.
+    int arglistStartIndex = JVMLmethodSignature.indexOf("(");
+    int arglistEndIndex = JVMLmethodSignature.indexOf(")") + 1;
+    String methodName = JVMLmethodSignature.substring(0, arglistStartIndex);
+
+    String fullyQualifiedMethodName = aclass.className;
```
mernst: Why is this called `fullyQualified**Method**Name`?  It seems to be a class name.

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,252 @@
...
+   */
+  private static List<String> readClassFile(
+      InputStream classInputStream,
+      AnnotationCategory annotationCategory,
+      Collection<String> qualifyingAnnotations)
+      throws IOException {
+    List<String> annotatedMethods = new ArrayList<>();
+
+    // Invoke the parser for the specified class file
+    AScene scene = new AScene();
```
mernst: The scene library is poor code that we are trying to get away from.  I would like to not add new dependencies on it.  Furthermore, I don't want to make the Checker Framework a dependency of Randoop.

There are better ways to parse a .class file.  The most popular library is ASM, but Randoop already uses BCEL, so I suggest that you use it.  That avoids adding a dependency.  Mark is an expert on BCEL in case you need any advice.

build.gradle
```diff
@@ -174,8 +174,11 @@ dependencies {
   implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.9'
   implementation group: 'commons-io', name: 'commons-io', version: '2.6'
   implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
+  implementation files("${System.properties['java.home']}/../lib/tools.jar")
   implementation configurations.junit.dependencies
   implementation files(project(':replacecall').sourceSets.main.output)
+  def cfLocalHome = String.valueOf(System.getenv("CHECKERFRAMEWORK"))
```
mernst: Let's not make Randoop depend on the Checker Framework.  See comment below about the Scene Library vs BCEL.

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,252 @@
...
+        // Check annotations for the method
+        Collection<Annotation> annotationLocation =
+            annotationCategory.equals(AnnotationCategory.NON_DET)
+                ? m.getValue().returnType.tlAnnotationsHere
+                : m.getValue().tlAnnotationsHere;
+        for (Annotation a : annotationLocation) {
+          // Relocation changes the constant annotation comparisons.
+          // Thus, we need to append the 'randoop.' prefix to the annotation
+          // parsed from the class files.
+          if (qualifyingAnnotations.contains("randoop." + a.def.name.trim())) {
```
mernst: Is the call to `trim()` necessary?  (This is moot if you use BCEL instead of the Scene Library.)

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  /**
+   * Returns the fully qualified signature from the given class and signature in JVML format.
+   *
+   * @param aclass AClass that the method belongs to
+   * @param JVMLmethodSignature method signature in JVML
+   * @return fully qualified method signature
+   */
+  private static String getFullyQualifiedSignatures(AClass aclass, String JVMLmethodSignature) {
```
mernst: The idiomatic way to capitalize the formal parameter name is `jvmlMethodSignature`.

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,238 @@
...
+      Collections.singletonList("org.checkerframework.checker.determinism.qual.NonDet");
+  /** The method annotations indicating a side-effect-free method. */
+  private static final Collection<String> SEF_ANNOTATIONS =
+      Arrays.asList(
+          "org.checkerframework.dataflow.qual.SideEffectFree",
+          "org.checkerframework.dataflow.qual.Pure");
+
+  /** Text to place at the beginning of file {@code omitmethods-defaults.txt}. */
+  private static final String OMIT_METHODS_FILE_HEADER =
+      "# Long-running.  With sufficiently small arguments, can be fast.\n"
```
mernst: Never use `\n` in source code.  A program should respect the newline convention of the host operating system.  Use `System.lineSeparator()` (possibly together with String.join).

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,238 @@
...
+      System.exit(1);
+    }
+  }
+
+  /**
+   * Returns the annotated methods from the the given jar.
+   *
+   * @param jarFile JDK .jar file path
+   * @param annotationCategory type of annotation we are parsing for
+   * @param annotations list of annotations
```
mernst: This is not descriptive of what the formal parameter is, means, or is used for.  This just restates the type, which is already explicit in the source code.

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,238 @@
...
+        // Check annotations for the method
+        Collection<Annotation> annotationLocation =
+            annotationCategory.equals(AnnotationCategory.NON_DET)
+                ? m.getValue().returnType.tlAnnotationsHere
+                : m.getValue().tlAnnotationsHere;
+        for (Annotation a : annotationLocation) {
+          // Relocation changes the constant annotation comparisons.
+          // Thus, we need to append the 'randoop.' prefix to the annotation
+          // parsed from the class files.
+          if (annotations.contains("randoop." + a.def.name.trim())) {
```
mernst: Is the call to trim() necessary?

src/main/java/randoop/resource/generator/MethodListGen.java
caseyxing: Thanks! I will definitely try to remember that.

src/main/java/randoop/resource/generator/MethodListGen.java
```diff
@@ -0,0 +1,252 @@
...
+            nonDetMethodWriter.write("^" + Pattern.quote(method) + "$");
+            nonDetMethodWriter.newLine();
+          }
+          nonDetMethodWriter.flush();
+        }
+      } else {
+        System.out.println("No nondeterministic methods found. Not writing to file.");
+      }
+
+      // There are some fully qualified signatures that Randoop cannot parse.
```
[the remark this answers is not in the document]
caseyxing: Not parsable: Could not parse signature java.lang.Enum.clone()
Not parsable: Could not parse signature java.lang.Object.clone()
Not parsable: Could not parse signature java.lang.String.indexOf(char[], int, int, char[], int, int, int)
Not parsable: Could not parse signature java.lang.String.indexOf(char[], int, int, java.lang.String, int)
Not parsable: Could not parse signature java.lang.String.lastIndexOf(char[], int, int, char[], int, int, int)
Not parsable: Could not parse signature java.lang.String.lastIndexOf(char[], int, int, java.lang.String, int)
Not parsable: Could not parse signature java.lang.String.lastIndexOfSupplementary(int, int)
Not parsable: Could not parse signature java.util.AbstractMap.clone()
Not parsable: Could not parse signature java.util.Hashtable$Entry.clone()
Not parsable: Could not parse signature java.util.logging.Logger.doLog(java.util.logging.LogRecord)
Not parsable: Could not parse signature java.util.logging.Logger.doLog(java.util.logging.LogRecord, java.lang.String)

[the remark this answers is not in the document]
caseyxing: I'm suspecting something related to visibility - for instance for the java.lang.String.lastIndexOf(char[], int, int, java.lang.String, int) method,
https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/String.java#l1829

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+ * This tool generates the resource files JDK-sef-methods.txt and omitmethods-defaults.txt, which
+ * contain a list of side effect free methods and a list of methods to omit during Randoop
+ * execution, respectively.
+ */
+public class MethodListGen {
+  /**
+   * The extension for a Java .class file. These files contain annotations that this tool will
+   * parse.
+   */
+  private static final String CLASS_EXT = ".class";
```
mernst: This doesn't serve any purpose.  It will never change, and it is used only once.  Use the string literal there and you can delete lines 38-43.

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+          System.lineSeparator(),
+          "# Long-running.  With sufficiently small arguments, can be fast.",
+          "# org.apache.commons.math3.analysis.differentiation.DSCompiler.getCompiler\\(int,int\\)",
+          "^org.apache.commons.math3.analysis.differentiation.",
+          "^org.apache.commons.math3.analysis.integration.",
+          "# Nondeterministic.");
+
+  /** An AnnotationLocation describes where to look for desired annotations. */
+  enum AnnotationLocation {
+    RETURN_TYPE,
```
mernst: Add Javadoc.  It can be short and obvious, but it should be present.

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+    METHOD
+  }
+
+  /**
+   * Main entry point to generate Nondeterministic and Side Effect Free Method lists.
+   *
+   * @param args command line arguments
+   *     <ul>
+   *       <li>args[0] - JDK .jar location as a path
```
mernst: Does this work with JDK 11, or only with JDK 8?

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+   *       <li>args[1] - output directory
+   *     </ul>
+   */
+  public static void main(String[] args) {
+    Path annotatedJar = Paths.get(args[0]);
+
+    Path nonDetFile = Paths.get(args[1], "omitmethods-defaults.txt");
```
mernst: Extract `args[1]` into a named variable.

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+    Path sideEffectFreeFile = Paths.get(args[1], "JDK-sef-methods.txt");
+    Path unparsableSideEffectFreeFile = Paths.get(args[1], "JDK-sef-methods-unparsable.txt");
+
+    try {
+      List<String> nonDetMethods =
+          getAnnotatedMethodsFromJar(
+              annotatedJar, AnnotationLocation.RETURN_TYPE, NONDET_ANNOTATIONS);
+      List<String> sideEffectFreeMethods =
```
mernst: Define variables as near their first use as possible.  This can come after the following if statement (near its own if statement).

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+    } catch (IOException e) {
+      e.printStackTrace();
+      System.exit(1);
+    }
+  }
+
+  /**
+   * Returns the annotated methods from the the given jar.
+   *
+   * @param jarFile JDK .jar file path
```
mernst: It doesn't have to be a JDK .jar file, right?  This should work with any .jar file.

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,241 @@
...
+        // Check annotations for the method
+        Collection<Annotation> annotationsAtLocation =
+            annotationLocation.equals(AnnotationLocation.RETURN_TYPE)
+                ? m.getValue().returnType.tlAnnotationsHere
+                : m.getValue().tlAnnotationsHere;
+        for (Annotation a : annotationsAtLocation) {
+          // Relocation changes the constant annotation comparisons.
+          // Thus, we need to append the 'randoop.' prefix to the annotation
+          // parsed from the class files.
+          if (annotations.contains("randoop." + a.def.name.trim())) {
```
mernst: This behavior is not documented.

PR #649, opened 2020-02-02
Implement --omit-classes and --omit-classes-file command-line options; fixes #599

### LINKED ISSUE #599: Implement --omit-classes and --omit-classes-file command-line arguments
I am trying to omit some classes. It didn't work. I checked the code and apparently the attributes `omit_classes_file` and `randoop.main.GenInputsAbstract#omit_classes` are never used. Do you have any other alternative to omit a class?
Thank you in advance for your time and support

mernst:


src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -980,6 +996,17 @@ public void checkOptionsValid() {
...
     }
+
+    // This does not exclude explicitly-specified methods.  In other words, if the user specified a
+    // method explicitly, this does not exclude it even if it is in an excluded class.
+    for (Iterator<String> itor = classnames.iterator(); itor.hasNext(); ) {
+      String classname = itor.next();
+      if (shouldOmitClass(classname)) {
+        itor.remove();
+        break;
+      }
```
Ahmedfir: I don't understand why there's the `break;` ? why don't you continue looping for the rest of the classnames?

mernst: You are right.  That was left over from logic before `shouldOmitClass` was refactored into its own method.  Thanks for noticing!  I appreciate your review.

PR #660, opened 2020-02-05
Side Effect Free and Nondeterministic Method List Generator w/ ASM

caseyxing:
This tool generates the SideEffectFree and Nondeterministic method lists from checker framework and Rashmi's work on determinism.


caseyxing:
Will take a look at pipeline results when I have a moment.

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,182 @@
...
+                  VisibilityPredicate.IS_ANY,
+                  new EverythingAllowedPredicate());
+              sideEffectMethodWriter.write(fullyQualifiedMethodSignature);
+              sideEffectMethodWriter.newLine();
+            } catch (RandoopUsageError | FailedPredicateException e) {
+              System.err.println("Not parsable: " + e.getMessage());
+              unparsableSideEffectMethodWriter.write(fullyQualifiedMethodSignature);
+              unparsableSideEffectMethodWriter.newLine();
+              e.printStackTrace(new PrintWriter(unparsableSideEffectMethodWriter));
+              unparsableSideEffectMethodWriter.newLine();
```
caseyxing: See the included unparseable methods file. It looks like the bulk of these either have the protected modifier or are static.

src/main/resources/JDK-sef-methods.txt
```diff
@@ -658,18 +978,62 @@ java.util.Date.equals(java.lang.Object)
 java.util.Date.hashCode()
 java.util.Date.toString()
 java.util.Deque.contains(java.lang.Object)
-java.util.Deque.isEmpty()
```
caseyxing: This change is unexpected. I'll find out.

caseyxing: Turns out this is an abstract function with no method body - I think this means that we can ignore these as anything that subclasses them will have their isEmpty marked as pure alongside their implementation.

caseyxing: Added context - @Pure is in included in the checker-framework source files.

build.gradle
```diff
@@ -171,10 +171,14 @@ dependencies {
 
   implementation 'org.apache.commons:commons-exec:1.3'
   implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.9'
+  implementation group: 'org.ow2.asm', name: 'asm', version: '7.3.1'
   implementation group: 'commons-io', name: 'commons-io', version: '2.6'
   implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
+  implementation files("${System.properties['java.home']}/../lib/tools.jar")
   implementation configurations.junit.dependencies
   implementation files(project(':replacecall').sourceSets.main.output)
+  def cfLocalHome = String.valueOf(System.getenv("CHECKERFRAMEWORK"))
```
caseyxing: I should remove this - this no longer depends on the checker framework.

src/main/java/randoop/resource/MethodListGen.java
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Returns the annotated methods from the the given jar.
+   *
+   * @param jarFile .jar file path
+   * @param annotations which annotations to capture
+   * @return list of annotated methods in fully-qualified signature format, in alphabetical order
+   * @throws IOException if Randoop cannot parse a method
+   */
+  private static List<String> getAnnotatedMethodsFromJar(
+      Path jarFile, Collection<String> annotations) throws IOException {
```
mernst: `throws IOException` does not seem accurate.  The code catches `IOException`, and code above catches `RandoopUsageError | FailedPredicateException`.

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+   * @param argumentSignature signature of the method's arguments, e.g.
+   *     "(char[],int,int,java.lang.String,int)"
+   * @return Randoop fully qualified signature
+   */
+  private String getFullyQualifiedMethodSignature(String method, String argumentSignature) {
+    return fullyQualifiedClassName + "." + method + argumentSignature;
+  }
+
+  /**
+   * Adds a method to the set of captured methods if any annotation on or within the method
```
mernst: "any annotation" here in the comment seems inaccurate, given the `annotationName` formal parameter.

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+   * signature is in {@code desiredAnnotations}. Ignores constructors and static constructors.
```
mernst: Why does it ignore constructors and static constructors?  Please justify the design decision.

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+   *
+   * @param annotationName annotation name, e.g.
+   *     "org.checkerframework.checker.determinism.qual.NonDet"
+   * @param method method name
```
mernst: In what format?

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+   */
+  private String getFullyQualifiedMethodSignature(String method, String argumentSignature) {
+    return fullyQualifiedClassName + "." + method + argumentSignature;
+  }
+
+  /**
+   * Adds a method to the set of captured methods if any annotation on or within the method
+   * signature is in {@code desiredAnnotations}. Ignores constructors and static constructors.
+   *
+   * @param annotationName annotation name, e.g.
```
mernst: Where does this come from and how is it used?

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+   * @param argumentSignature signature of the method's arguments, e.g.
+   *     "(char[],int,int,java.lang.String,int)"
+   */
+  private void addMethodIfMatching(String annotationName, String method, String argumentSignature) {
+    if (method.contains("<init>") || method.contains("<clinit>")) {
+      return; // Ignore constructors and static constructors
+    }
+
+    if (desiredAnnotations.contains(
+        Signatures.fieldDescriptorToBinaryName(annotationName.trim()))) {
```
mernst: Why is the call to `trim()` necessary?  If it's necessary, I would move it to the client.

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+  private static final int ASM_OPCODE = Opcodes.ASM7;
+
+  /** The desired annotations that serve as the criteria for methods we want to capture. */
+  private final Collection<String> desiredAnnotations;
+
+  /** The Randoop fully qualified class name of the class being visited. */
+  private String fullyQualifiedClassName = null;
+
+  /** The accumulated set of captured fully qualified method signatures. */
+  private final Set<String> matchingFullyQualifiedMethodSignatures = new HashSet<>();
```
mernst: Why is this a set rather than a list?  Is there a possibility of duplicates?  (Maybe if a method contains both `@SideEffectFree` and `@Pure`?)

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,148 @@
...
+import org.objectweb.asm.TypePath;
+import org.plumelib.reflection.Signatures;
+
+/**
+ * ClassAnnotationScanner is a ClassVisitor that collects methods that are annotated with specific
+ * annotations.
+ */
+public class ClassAnnotationScanner extends ClassVisitor {
+  /** OpCode for ASM used throughout this class. */
+  private static final int ASM_OPCODE = Opcodes.ASM7;
```
mernst: Is this an opcode or a version number?  I think it's the latter and should be documented as such.  It's only used once, so you could move its use to there.

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @Override
+  public void visit(
+      int version,
+      int access,
+      java.lang.String name,
+      java.lang.String signature,
+      java.lang.String superName,
+      java.lang.String[] interfaces) {
+    className = name.replace('/', '.');
```
mernst: What is the format of `name`?  It's better to use a method from the [`Signatures`](http://plumelib.org/reflection-util/api/org/plumelib/reflection/Signatures.html) class rather than doing your own string replacement.

mernst: This code looks wrong in the presence of nested classes.
Consider code in the outer class that comes lexically after the nested class.  Its `className` field will be incorrectly set.  

You could fix that by using a stack here, but I think it's cleaner to create a new scanner for each nested class, and never reset this field.  You could guard against that happening by throwing an error if the field is not null when it is encountered.  Or, pass the class name into the constructor and you can make the field final.

mernst (changes requested):

src/main/java/randoop/resource/ClassAnnotationScanner.java
mernst: Before fixing this, please make a test case (with nested classes) that fails for the current code, so we can make sure that the fix actually works.

src/main/java/randoop/resource/ClassAnnotationScanner.java
caseyxing: Yes, that is the intent for that very reason (which I'm not sure actually arises, but to my knowledge I didn't see any guarantees that `@Pure` and `@SideEffectFree` were mutually exclusive.

src/main/java/randoop/resource/ClassAnnotationScanner.java
caseyxing: That's an interesting point, let me take a closer look at this.

mernst (changes requested):
Here are some code review comments.

src/main/java/randoop/resource/ClassAnnotationScanner.java
```diff
@@ -0,0 +1,156 @@
...
+  public void visit(
+      int version,
+      int access,
+      String name,
+      String signature,
+      String superName,
+      String[] interfaces) {
+
+    @SuppressWarnings("signature")
+    @InternalForm String internalFormName = name;
```
mernst: I suggested putting the annotation on the formal parameter.  (I think that is clearer, and it avoids the need for a temporary variable.)  Can you explain why you moved it here rather than doing that?

src/systemTest/resources/JDK7-omitmethods.txt
```diff
@@ -2,21 +2,52 @@
...
-^java7\.util7\.Arrays\.hashCode\(java\.lang\.Object\[\]\)$
-^java7\.util7\.Arrays\.hashCode\(long\[\]\)$
-^java7\.util7\.Arrays\.hashCode\(short\[\]\)$
-^java7\.util7\.Collection\.hashCode\(\)$
-^java7\.util7\.Collections\.shuffle\(java7\.util7\.List\)$
-^java7\.util7\.Comparator\.compare\(java\.lang\.Object, java\.lang\.Object\)$
-^java7\.util7\.List\.hashCode\(\)$
-^java7\.util7\.Random\.<init>\(\)$
-^java7\.util7\.Set\.hashCode\(\)$
+^\Qjava7.util7.AbstractList.hashCode()\E$
```
mernst: Please add a README file in `src/systemTest/resources` (or, better, a Gradle target) that explains how to generate `JDK7-omitmethods.txt` and `JDK7-sef-methods.txt`.

mernst (changes requested):
Thanks for these changes.  I have a few questions and concerns.

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -589,11 +569,34 @@ public void runJDKTest() {
             "java7.util7.Vector.removeRange(int, int) exclude",
             "java7.util7.Vector.writeObject(java.io.ObjectOutputStream) exclude",
             "java7.util7.WeakHashMap.eq(java.lang.Object, java.lang.Object) ignore", // Travis
-            "java7.util7.WeakHashMap.putAll(java7.util7.Map) exclude",
             "java7.util7.WeakHashMap.removeMapping(java.lang.Object) exclude",
             "java7.util7.WeakHashMap.resize(int) ignore",
             "java7.util7.WeakHashMap.transfer(java7.util7.WeakHashMap.Entry[], java7.util7.WeakHashMap.Entry[]) ignore",
-            "java7.util7.WeakHashMap.unmaskNull(java.lang.Object) ignore"
+            "java7.util7.WeakHashMap.unmaskNull(java.lang.Object) ignore",
```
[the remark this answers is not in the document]
caseyxing: Absolutely! Was in a rush to get systemTest pipeline working, but will definitely fix that.

src/systemTest/resources/README-JDK7.txt
```diff
@@ -0,0 +1,8 @@
+To generate the methods files 'JDK7-sef-methods.txt' and 'JDK7-omitmethods.txt':
+
+- JDK7-sef-methods.txt: Copy the generated JDK-sef-methods.txt from the output of the 
+    MethodListGen tool. Manually remove the files present in JDK8 but not JDK7.
```
mernst: Which files are these?  Can you provide a grep command that does this?

caseyxing: I'll clarify this.

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -974,7 +992,8 @@ public void runCollectionGenerationTest() {
             options,
             "collectiongen.Day.valueOf(java.lang.String) exclude",
             "collectiongen.AnInputClass.hashCode() ignore",
-            "collectiongen.Day.values() ignore");
+            "collectiongen.Day.values() ignore",
+            "collectiongen.AnInputClass.toString() ignore");
```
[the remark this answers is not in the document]
caseyxing: I think we discussed this in an email (though we only got to investigating one of them).

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -1498,8 +1520,8 @@ public void runBadCollectionSizeTest() {
         systemTestEnvironmentManager.createTestEnvironment("bad-collection-size");
     RandoopOptions options = createRandoopOptions(testEnvironment);
     options.addTestClass("collections.BadCollection");
-    options.setOption("generated_limit", "10");
-    options.setOption("output_limit", "10");
+    options.setOption("generated_limit", "5");
```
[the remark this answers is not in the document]
caseyxing: I think we discussed this in an email (though we only got to investigating one of them).

src/systemTest/java/randoop/main/RandoopSystemTest.java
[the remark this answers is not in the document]
caseyxing: Ended up reverting coverage changes checks for investigation in differences.

src/main/java/randoop/reflection/OperationModel.java
```diff
@@ -305,7 +305,7 @@ public void addClassLiterals(
    */
   public static MultiMap<Type, TypedClassOperation> readOperations(@Nullable Path file)
       throws OperationParseException {
-    return readOperations(file, false);
+    return readOperations(file, true);
```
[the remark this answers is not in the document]
caseyxing: This was reverted by you; thanks!

src/systemTest/resources/README-JDK7.txt
```diff
@@ -2,8 +2,433 @@ To generate the methods files 'JDK7-sef-methods.txt' and 'JDK7-omitmethods.txt':
...
     MethodListGen tool. Replace every instance of 'java.util' within the file to 'java7.util7'.
-    You may need to manually remove the methods referring to those present in JDK8 but not JDK7,
-    though at the time of this writing, no such difference exists.
+    You may need to manually remove the methods referring to those present in JDK8 but not JDK7.
+
+    You can generate the list of methods to exclude by adding logging to failed method parsing to the throw RandoopUsageError block in
+    OperationModel.readOperation(EntryReader, boolean) and making it continue rather than throwing an exception.
+    At the time of this writing, the list of methods removed are (you can 'subtract lines of one file from
+    another by 'grep -F -x -v -f subtract.txt all.txt > result.txt':
+        java7.util7.ArrayList$SubList.spliterator()
```
mernst: Can you put this long text in another file to make the file more readable?  I'm not sure whether it is subtract.txt or result.txt.  Thanks.

caseyxing: Sure thing!

PR #813, opened 2020-10-06
Add tasks to publish to Maven repositories

smillst:
This publishes the fat jars.  I did that because they contain shadowed versions of the dependencies.

* `./gradlew publishLocalPublicationToMavenLocal` publishes the artifacts to the local Maven repository.
* `./gradlew publishRemotePublicationToMavenRepository -PSONATYPE_NEXUS_USERNAME=username -PSONATYPE_NEXUS_PASSWORD=password` publishes the artifacts to the Nexus repository.  If the version number ends in `SNAPSHOT`, the release is published as a snapshot and no signing is required. 
* If this is not a SNAPSHOT release, then add `-Psigning.secretKeyRingFile=file -Psigning.password=secret -Psigning.keyId=id` to the command.

I have only tested the publishing to local since we have yet to set up the remote repository.


gradle-mvn-push.gradle
```diff
@@ -0,0 +1,104 @@
...
+            developerConnection = 'scm:git:ssh://git@github.com/randoop/randoop.git'
+        }
+
+        licenses {
+            license {
+                name = 'The MIT License'
+                url = 'http://opensource.org/licenses/MIT'
+                distribution = 'repo'
+            }
+            // TODO:
```
smillst, line 33: What license does this use?

agent/covered-class/covered-class.gradle
```diff
@@ -57,3 +57,32 @@ shadowJar {
...
 }
+
+apply from: rootProject.file("gradle-mvn-push.gradle")
+
+final coveredClassPom(publication) {
+    // Don't use publication.from components.java which would publish the skinny jar as randoop.jar.
+    // Information that is in all pom files is configured in randoop/gradle-mvn-push.gradle.
+    publication.pom {
+        name = 'Randoop\'s covered-class agent'
+        // description = ''
```
smillst, line 68: You should add a description.

agent/replacecall/replacecall.gradle
```diff
@@ -59,3 +59,32 @@ shadowJar {
...
 }
+
+apply from: rootProject.file("gradle-mvn-push.gradle")
+
+final replacecallPom(publication) {
+    // Don't use publication.from components.java which would publish the skinny jar as randoop.jar.
+    // Information that is in all pom files is configured in randoop/gradle-mvn-push.gradle.
+    publication.pom {
+        name = 'Randoop\'s replacecall agent'
+        // description = ''
```
smillst, line 70: You should add a description.

gradle-mvn-push.gradle
mernst: I can't find a short name for it.  That short name might not be useful, because Oracle keeps changing their license terms for a given-named license.  It is not restrictive.  It seems close to [Attribution-ShareAlike 4.0 International (CC BY-SA 4.0)](https://creativecommons.org/licenses/by-sa/4.0/?ref=chooser-v1)3.

mernst (approved):

gradle-mvn-push.gradle
smillst: So should I add a license block with that links to the license you provided?  It's not clear from your comment.

smillst:
@mernst, I don't have write access to this repository, so I can't assign you.

gradle-mvn-push.gradle
mernst: Sorry for the ambiguity.  I perhaps didn't understand your question.  If you want to include a license, that is the best one to include.  If so, I would add a comment saying it's close to that and including a link to the actual license for clarity.

smillst: I copied this line from `randoop/license.txt` and beyond that I don't have any insight into this file or its license. Your comment said I could add one if I want, but I don't know if we need one or not.

mernst: Can you clarify what you mean by "this file"?  Are you asking about the license for `gradle-mvn-push.gradle`?

Regarding whether the gradle file needs to have a license block, that seems to be a gradle question.  I believe that some people won't use a project that lacks one.

smillst: Here's the text of randoops `license.txt`

```
Unless otherwise noted, Randoop is distributed under the MIT license,
which appears below.  Some files in Randoop's distribution are
distributed under a different license. These exceptions are:

Randoop source code:
  src/randoop/util/StreamRedirectThread.java (License in file)

Experimental subject programs (not part of Randoop's implementation)
  systemTests/ (various licenses)
  tests/ (various licenses)
```

I've added a license block for The MIT License.  I don't know if we need one for `src/randoop/util/StreamRedirectThread.java`.  It's in the Jar file, so I assume so. We are not distributing the experimental subject programs in Maven, so we don't have to worry about them.

mernst: The file is not used, so I will open a pull request to remove it from Randoop, which makes this question moot.

PR #887, opened 2021-04-12
Use fastutil library for primitive collections

mernst:


markro49:
This looks good.

mernst:
Don't use fastutil:
 * fastutil attempts compatibility with the Java Collections Framework, but that makes it error-prone:  boxing occurs in foreach loops and other places that are easy to overlook.  It's better to use an API that is less error-prone.
 * the author claims that a priority queue can depend on the order of insertion, so use of the fastutil data structures is misleading to someone who assumes the standard definition of "priority queue".  The author doesn't understand priority queues and does not care to: See https://github.com/vigna/fastutil/issues/253 .

PR #890, opened 2021-04-22
Use `mapList`

mernst:


src/main/java/randoop/sequence/Variable.java
```diff
@@ -84,11 +84,7 @@ public Variable copyWithIndexUpdated(Map<Integer, Integer> map) {
   }
 
   public static List<Integer> statementIndexList(List<Variable> values) {
-    List<Integer> result = new ArrayList<>(values.size());
-    for (Variable value : values) {
-      result.add(value.getDeclIndex());
-    }
-    return result;
+    return CollectionsPlume.mapList(Variable::getDeclIndex, values);
```
markro49, line 87: Why no "import org.plumelib.util.CollectionsPlume" in this file?  If not needed, how is it different for others?

markro49 (approved):
Looks good.  I have some questions for my own knowledge on use of mapList, main one is when to use lambda and when use to method reference operator.

src/main/java/randoop/sequence/Variable.java
mernst: This particular change is within a comment (see line 93/89 below, or expand the code view upward to see the opening comment).

mernst:
> when to use lambda and when use to method reference operator.

The method reference operator can always be replaced by a lambda.  The method reference is more concise, however.

PR #927, opened 2021-06-02
added --test-package command-line option

Ivan-Kocherhin:
Hi there. We try to utilize Randoop for generation regression tests for a huge and legacy code.
Potentially it can be a robust solution, but coverage is the key here.
We have got a better coverage when ran Randoop for each package separately, because in that way protected and package-private classes and methods are covered. In addition, I had to write down every class present in the package. It would be easier just define package to cover, but not all classes inside. I didn`t find this option inside Randoop, so decided to implement by my own.


src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
 
+  /**
+   * Returns classes in the given package. Does not include classes in sub-packages. Ignores
+   * non-accessible classes.
+   *
+   * @param packageName a package name
+   * @param accessibility the accessibility predicate
+   * @return classes in package {@code packageName}
+   */
+  public static Set<@ClassGetName String> getClassnamesFromPackage(
```
mernst, line 1184: Why are all the new methods `public`?

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+    }
+    return classnames;
+  }
+
+  /**
+   * Returns list of jars and directories, present in classpath.
+   *
+   * @return List of jars and directories, present in classpath
+   */
+  public static List<File> getClassLocationsForCurrentClasspath() {
```
mernst, line 1304: Creating a list is not necessary.  This method can return an array, eliminating the need for the loop and the `urls` variable.  (Then, you might find the method so small that you can inline it.)

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+
+  /**
+   * Searches in the given directory for classes with the given package.
+   *
+   * @param directory a directory
+   * @param packageName a package name
+   * @param accessibility the accessibility predicate
+   * @return classes with the given package that were found in the specified directory
+   */
+  public static Set<@ClassGetName String> searchInDirectory(
```
mernst, line 1207: The name `search` is not descriptive.  It would be better for the name to be something like `classesInDirectory`.  The same comment applies to the next method as well.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+      } else if (location.isDirectory()) {
+        classnames.addAll(searchInDirectory(location, packageName, accessibility));
+      }
+    }
+
+    return classnames;
+  }
+
+  /**
+   * Searches in the given directory for classes with the given package.
```
mernst, line 1200: Does "in the directory" include in subdirectories?  Please make the documentation more explicit.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+      try (Stream<Path> allFiles = Files.walk(Paths.get(packageDirectory.getPath()))) {
+        @SuppressWarnings("signature")
+        Set<@ClassGetName String> classnames =
+            allFiles
+                .filter(e -> e.toFile().isFile()) // to ignore directories
+                .map(Path::toString)
+                .map(
+                    absolutePath ->
+                        absolutePath.replace(
+                            directory.getPath() + File.separator, "")) // to delete absolute path
```
mernst, line 1222: What does it mean to "delete an absolute path"?  This looks like it is removing characters, but it isn't deleting any files.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+                    relativePath -> relativePath.startsWith(packageNameAsFile)) // to filter package
+                .filter(relativePath -> relativePath.endsWith(".class")) // to only get classes
+                .map(
+                    relativePath ->
+                        relativePath.substring(0, relativePath.length() - ".class".length()))
+                .map(ifClassName -> ifClassName.replace(File.separator, "."))
+                .filter(
+                    classname -> {
+                      try {
+                        return classname.startsWith(packageName)
```
mernst, line 1233: It is awkward to do a partial test for the package name on line 1224 and more here.  Please put the entire test for package name in one place.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+   * @return classes with the given package that were found in the specified directory
+   */
+  public static Set<@ClassGetName String> searchInDirectory(
+      File directory, String packageName, AccessibilityPredicate accessibility) {
+    String packageNameAsFile = packageName.replace(".", File.separator) /*.concat(File.separator)*/;
+    File packageDirectory =
+        new File(directory.getPath().concat(File.separator).concat(packageNameAsFile));
+    if (packageDirectory.exists()) {
+      try (Stream<Path> allFiles = Files.walk(Paths.get(packageDirectory.getPath()))) {
+        @SuppressWarnings("signature")
```
mernst, line 1214: Anytime you suppress a warning, write a comment to explain why the code is definitely correct and the suppressed warning is definitely a false positive warning.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+        Set<@ClassGetName String> classnames =
+            allFiles
+                .filter(e -> e.toFile().isFile()) // to ignore directories
```
mernst, line 1217: I find this long sequence of stream operations a bit awkward.  I suspect the code would be cleaner and clearer as a for loop with statements inside.  You might also consider passing a visitor to `walkFileTree`.  However, if you prefer this idiom, that is OK.  Just group the operations so that like functionality is grouped.  For example, currently the filtering occurs in multiple different places.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1147,6 +1173,145 @@ private static boolean shouldOmitClass(String classname) {
...
+   *
+   * @param jarFile a jar file containing classes
+   * @param packageName a package name
+   * @param accessibility the accessibility predicate
+   * @return classes in package {@code packageName} in the given jar file
+   */
+  public static Set<@ClassGetName String> searchInJar(
+      File jarFile, String packageName, AccessibilityPredicate accessibility) {
+    Set<@ClassGetName String> classnames = new HashSet<>();
+    @ClassGetName String classname = "randoop.main.GenInputsAbstract";
```
mernst, line 1269: Is the `@ClassGetName` type annotation necessary?  I would expect it to be inferred.

mernst (changes requested):
Thanks for this feature!  It will make Randoop easier to use.

I made some edits to your code, and I have left some comments about suggested improvements.  Could you please address those and then ping me for another review?

Thanks again.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: Made methods private because they are not used anywhere besides GetInputAbstract

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: Good idea! I inlined it and moved creation of File object to loop which called that method, so new list is not created now.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: Changed names to getClassesWithPackageFromDirectory and getClassesWithPackageFromJar, hope these names are more descriptive and understandable

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: Directory must be root directory with .class files. Extended documentation to explain what directory is expected and provided examples.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: It meant to return relative path from the directory root. I changed stream operations with for loop, so now this operation is assignment to relativePath variable and I hope it is more readable even without any comments.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: All tests, except accessibility.isAccessible are now placed in filter provided to listFiles method. accessibility.isAccessible test is separate because it tests class, not a file or filename. Also, some additional operations to perform this check are needed, e. g. transformation of filename to ClassGetName form.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: I needed to perform some operations to transform filename to ClassGetName form. Wrote comments explaining operations and why suppression is needed.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: Thanks for your advice! We don`t add classes in subdirectories, so I replaced Files.walk() with for loop called on directory.listFiles() which only shows files for one directory. It allows to skip a lot of extra work, easily filter files and makes code shorter and more readable, so I hope this solution is better than previous.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: Removed type annotation, works fine.

src/main/java/randoop/main/GenInputsAbstract.java
mernst: Thanks for addressing my feedback.  I appreciate it.  As a general rule, if you accepted a suggestion, you don't need to respond.  (This advice is from 
https://homes.cs.washington.edu/~mernst/advice/github-pull-request.html#respond-to-reviews .)

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1148,6 +1170,117 @@ private static boolean shouldOmitClass(String classname) {
...
+      } else if (location.isDirectory()) {
+        classnames.addAll(getClassesWithPackageFromDirectory(location, packageName, accessibility));
+      }
+    }
+
+    return classnames;
+  }
+
+  /**
+   * Searches in the given directory for classes with the given package. Directory must be root
```
mernst, line 1198: I don't understand what a "root directory" is.  In typical file system parlance, this would be "/" in Unix or "C:\" in Windows.
Also, this method does *not* search the given directory.  It searches one subdirectory of the given directory, where the subdirectory corresponds to the package name.  Could you please clarify the comment?

I was confused before about whether the search is deep or shallow (that is, examines subdirectories or only files directly present in the given directory), but I think clarifying the documentation will prevent that confusion.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1148,6 +1170,117 @@ private static boolean shouldOmitClass(String classname) {
...
+    File packageDirectory =
+        new File(directory.getPath().concat(File.separator).concat(packageNameAsFile));
+    if (packageDirectory.exists() && packageDirectory.isDirectory()) {
+      Set<@ClassGetName String> classnames = new HashSet<>();
+      for (File file :
+          packageDirectory.listFiles(f -> f.isFile() && f.getName().endsWith(".class"))) {
+
+        String relativePath =
+            file.getPath()
+                .substring(directory.getAbsolutePath().length() + File.separator.length());
```
mernst, line 1219: I believe that if `directory` is a relative pathname, then `file` will be too, in which case trimming off enough characters for `directory`'s absolute pathname would be incorrect.  Could you double-check that?
I wonder if the [`relativize`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Path.html#relativize(java.nio.file.Path)) method would be useful here instead of doing string manipulation.  String manipulation is a code smell.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1148,6 +1170,117 @@ private static boolean shouldOmitClass(String classname) {
...
+        String internalForm = relativePath.substring(0, relativePath.length() - ".class".length());
+        @SuppressWarnings("signature") // classname must have @ClassGetName annotation to be added
+        // to classnames set, but we assign string without any annotations
+        @ClassGetName String classname = internalForm.replace(File.separator, ".");
```
mernst, line 1223: Please use [`Signatures.internalFormToClassGetName`](http://plumelib.org/reflection-util/api/org/plumelib/reflection/Signatures.html#internalFormToClassGetName-java.lang.String-) rather than performing string manipulation.  Then you won't need to suppress any warnings, either.

mernst (changes requested):
Thanks for the improvements.  This is definitely better.  I have a few more suggestions.

src/main/java/randoop/main/GenInputsAbstract.java
Ivan-Kocherhin: I tried, and when searching in jars it works just fine, but this method does not work well in directories. I am using Windows and have '\' separator instead of '/', so it simply returns the same string. Should I update `Signatures.internalFormToClassGetName` method to replace '\' too or should I ensure string passed to it has correct separator? Also, there are `Signatures.classfilenameToBinaryName` and `Signatures.binaryNameToClassGetName` methods, that might work. I am not sure what solution will be correct in this situation. Thank you for your help.

mernst: Thank you for pointing that out.  The variable name `internalForm` is misleading -- on Unix, that variable holds a string in internal form, but on Windows it does not.  I suggest changing that variable name to avoid confusion.

Ivan-Kocherhin: I decided to use `Signatures.classfilenameToBinaryName` and `Signatures.binaryNameToClassGetName` methods. They perform all string and type manipulations needed, so `internalForm` variable and `@SuprressWarnings` annotations can be deleted.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1148,6 +1169,110 @@ private static boolean shouldOmitClass(String classname) {
...
+   * @param accessibility the accessibility predicate
+   * @return classes with the given package that were found in the specified directory
+   */
+  private static Set<@ClassGetName String> getClassesWithPackageFromDirectory(
+      File directory, String packageName, AccessibilityPredicate accessibility) {
+    String packageNameAsFile = packageName.replace(".", File.separator);
+    File packageDirectory = // to find needed directory so we don`t need to check every file
+      directory.toPath().resolve(packageNameAsFile).toFile();
+    if (packageDirectory.exists() && packageDirectory.isDirectory()) {
+      Set<@ClassGetName String> classnames = new TreeSet<>();
```
mernst, line 1213: Because this is a private helper method whose results are always put in a `TreeSet` by the caller, using a `TreeSet` here rather than a list is not necessary and in fact is slightly inefficient.  It's not a problem if you prefer it, though.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -1148,6 +1169,110 @@ private static boolean shouldOmitClass(String classname) {
...
+      for (File file :
+          packageDirectory.listFiles(f -> f.isFile() && f.getName().endsWith(".class"))) {
+
+        String relativePath =
+            directory.toPath().relativize(file.toPath()).toString();
+        @ClassGetName String classname = Signatures.binaryNameToClassGetName(
```
mernst, line 1219: Every binary name is a classgetname (see the [type hierarchy](https://checkerframework.org/manual/#signature-annotations)), so the call to `binaryNameToClassGetName` is a no-op (except in erroneous circumstances).  You can remove it.  You also don't need the `@ClassGetName` annotation; it will be inferred.

mernst (approved):
Thanks for this feature!  The code looks good now.

Please double-check that the code still works after my changes.  Then I will merge the pull request.

Thanks again.

Ivan-Kocherhin:
Code works fine, you can merge it. Thanks for your review and help!

mernst:
Terrific!  Thanks so much for this contribution.

PR #932, opened 2021-06-09
Add --add-dependencies command-line option

Ivan-Kocherhin:
Hello! We continue trying to make the Randoop configuration process for generating regression tests suite with wider coverage more user-friendly. 
I am a bit automating the process. 
Documentation says:

> Suppose you want to test method A.m. You should direct Randoop to call not just A.m, but also methods that create A.m's inputs. More concretely, if A.m's parameter type is B, then you should tell Randoop to also call methods or constructors that create objects of type B. When a test calls A.m, all of the methods that set up state or create argument values are also being implicitly tested.

That means we should provide dependencies of tested classes and methods. 
To make it easier, we add a command-line option for automatically finding dependencies of desired classes and methods, then putting the result into the testing classes list.


src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -108,6 +118,16 @@ protected GenInputsAbstract(
...
 
+  /**
+   * Automatically add dependencies to the classes under test. For any classes, methods, and
+   * constructors specified by the user, this option adds their argument classes. (The user
+   * specifies them using the {@code --testjar}, {@code --classlist}, {@code --testclass}, and
+   * {@code --methodlist} options.) Recommended to use with {@code --require-covered-classes}
+   * option.
+   */
+  @Option("Add classes that are method/constructor arguments")
+  public static boolean add_dependencies = false;
```
mernst, line 129: Most command-line arguments that add classes/methods under test start with `--test`.  (And the others arguably should, but that change would be disruptive to users.)  I suggest this one should follow the same convention.

mernst (changes requested):
Thanks for this feature.  It's better than the manual's current suggestion of running shell commands to compute dependencies.

Does computation of dependent classes compute the transitive closure?  That is, does it work recursively?

Consider this code:

```java
class A {}

class B {
  B(A a);
}

class C {
  C(B b);
}
```

Suppose that the user specifies C as a class under test.  This code will add B as a class under test, but A is also needed.  Does this code add A?  Please document whether that is the case.  If possible, it would be best to include A as well.

Here is another example.  In the following code:

```java
abstract class D {} // cannot be instantiated

class E extends D {}

class F {
  F(D d);
}
```

the only way to instantiate a `F` is by passing an object with run-time class `E`.  Will `E` be added to the classes under test?  If not, please document this limitation, or fix it by including each class that is a subtype of any of the dependent ones.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -108,6 +118,16 @@ protected GenInputsAbstract(
   @Option("File that lists methods under test")
   public static Path methodlist = null;
 
+  /**
+   * Automatically add dependencies to the classes under test. For any classes, methods, and
+   * constructors specified by the user, this option adds their argument classes. (The user
+   * specifies them using the {@code --testjar}, {@code --classlist}, {@code --testclass}, and
+   * {@code --methodlist} options.) Recommended to use with {@code --require-covered-classes}
```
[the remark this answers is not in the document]
Ivan-Kocherhin: It`s the second one. This suggestion was added because we can end up with more tests for dependencies than for code we actually wanted to test. The logic behind it is dependencies should be tested separately, doesn`t matter using Randoop or hand-written tests.

Ivan-Kocherhin:
@mernst 
Thank you for your review and interest in my work! The answers to your questions:

1. No, it doesn't add dependencies recursively. So in your example, only class B will be added. I'm afraid if adding all dependencies recursively in some cases we will end up testing a lot of extra classes and functionality of libraries and/or frameworks instead of just testing our code. But your remark gave me an idea of adding depth option (let's call it `--test-add-dependencies-depth`), so user will decide how deep dependencies will be added. So, e. g., if `--test-add-dependencies-depth=1` only B class will be added, but when `--test-add-dependencies-depth=2+` class A will be added too.
2. No, it doesn't add child classes of abstract classes and interfaces. I agree this would be great functionality, but I'm not sure it can be done without loading and testing all classes under classpath, what in my opinion is not a good solution.

mernst:
@Ivan-Kocherhin Thanks for the information.

I agree there is a tradeoff regarding computing recursively:  the ability to generate more objects, vs. the danger of defocusing Randoop from the targeted classes.  (Randoop was designed for testing entire projects rather than specific classes.)  I don't know what the right tradeoff is, and it might differ from project to project.

It's possible to read .class files without loading them (say, via a library like ASM), but you are right that that would require extra logic in Randoop and would again increase the number of targeted classes.

In each case, please do document the code's current assumptions and functionality.  You can do this in the Javadoc, which will appear in the manual.  That will help users know what to expect.

Thanks again!

src/main/java/randoop/main/GenTests.java
```diff
@@ -293,6 +300,11 @@ public boolean handle(String[] args) {
       classNameErrorHandler = new WarnOnBadClassName();
     }
 
+    if (test_add_dependencies) {
```
mernst, line 303: I expected the implementation to be something like

```java
    if (test_add_dependencies) {
      classnames.addAll(getDependentClassnamesFromClassnames(classnames, accessibility));
      classnames.addAll(getDependentClassnamesFromMethodList(accessibility));
      for (int i=1; i<test_add_dependencies_depth; i++) {
        classnames.addAll(getDependentClassnamesFromClassnames(classnames, accessibility));
      }
    }
```

This may be less efficient than other possible implementations, but it permits a much simpler implementation of `getDependentClassnamesFromMethodList` which is agnostic to depth `getDependentClassnamesFromClassnames`.  (You could use your previous, simpler implementations.)

Is there a reason not to prefer an approach like this?

mernst (changes requested):
I have a couple of questions about the design, before I take a closer look at the method implementations.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -121,6 +121,36 @@ protected GenInputsAbstract(
...
+  @Option("Add classes that are method/constructor arguments")
+  public static boolean test_add_dependencies = false;
+
+  /**
+   * Configures the depth of adding dependendencies when using the {@code --test-add-dependencies}
+   * option. When this option is 1, only the direct dependencies of tested classes and methods are
+   * added. When this option is 2, dependencies of dependendencies are also added. When this option
+   * is 3, dependencies of dependendencies of dependendencies are also added. And so forth.
+   *
+   * <p>If code contains circular dependencies, Randoop will add those dependencies repeatedly.
```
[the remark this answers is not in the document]
Ivan-Kocherhin: When we call method `GenTests.getDependentClassnamesFromClassnamesWithDepth` recursively, we use dependencies classnames created inside the method instead of using classnames provided to method in arguments. So when searching dependencies depth X, we only know dependencies depth X-1, but not previous. So in situation when tested class A uses B and B uses A again (this code has a high coupling and is a red flag of bad architecture, but unfortunately such situations are still possible) and we search for dependencies depth 2 we only know about class B, not A. In the end we will have both classes A and B, but searching for the same classes will cost a lot of extra resources.

src/main/java/randoop/main/GenTests.java
Ivan-Kocherhin: Good suggestion! I think it can be improved by adding intermediate variables:
```java
if (test_add_dependencies) {
      classnames.addAll(getDependentClassnamesFromClassnames(classnames, accessibility));
      classnames.addAll(getDependentClassnamesFromMethodList(accessibility));
      Set<@ClassGetName String> searchDependenciesFor = classnames;
      Set<@ClassGetName String> dependencies;
      for (int i = 2; i <= test_add_dependencies_depth; ++i) {
        dependencies = getDependentClassnamesFromClassnames(searchDependenciesFor, accessibility);
        classnames.addAll(dependencies);
        searchDependenciesFor = dependencies;
      }
    }
```
In this case simpler implementation can be used and it is more efficient because we don`t search for dependencies of every class over and over again.

src/main/java/randoop/main/GenInputsAbstract.java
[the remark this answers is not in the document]
mernst: I agree that from an implementation point of view, Randoop will notice classes repeatedly.

I think that from a user point of view, the classes only appear once, as if the user had specified them in `--testclass` without duplication.  (Am I right about that?)

Since this Javadoc is user documentation, I would drop this remark.  You could add it as a code comment, where it will be visible only to people working on the Randoop source code (the main people who should care).

src/main/java/randoop/main/GenTests.java
mernst: This is great.  (It's what I had in mind, but I wanted to paste in a simpler version.)  Thanks for making this change.

src/main/java/randoop/main/GenTests.java
```diff
@@ -1335,4 +1358,150 @@ private Path getResourceDirectoryPath(String resourceDirectory) {
...
+   * if it is a parameter to a method/constructor. Does not return omitted or non-accessible
+   * classes. Does not return dependencies for non-accessible methods/constructors.
+   *
+   * @param classnames names of dependent classes
+   * @param accessibility accessibility predicate
+   * @return classnames of dependencies
+   */
+  public static Set<@ClassGetName String> getDependentClassnamesFromClassnames(
+      Set<@ClassGetName String> classnames, AccessibilityPredicate accessibility) {
+    Set<@ClassGetName String> dependenciesClassnames = new TreeSet<>();
```
mernst, line 1373: This code seems to be copied from `addOperationsUsingSignatures`.
Rather than duplicating code, please refactor it to avoid the duplication.  I have done that in https://github.com/randoop/randoop/pull/938.

src/main/java/randoop/main/GenTests.java
```diff
@@ -1335,4 +1358,150 @@ private Path getResourceDirectoryPath(String resourceDirectory) {
...
+  }
+
+  /**
+   * Tests whether a method should be omitted. Returns true if the signature matches any pattern in
+   * omit_method list, false otherwise.
+   *
+   * @param signature signature of method to test
+   * @return true if method should be ommited
+   */
+  private static boolean shouldOmitMethod(String signature) {
```
mernst, line 1439: This implementation is nice and simple, but it does not respect the Randoop documentation, which requires checking other classes in the type hierarchy: https://randoop.github.io/randoop/manual/#option:omit-methods .

I think it is better to make a regular `OmitMethodsPredicate` rather than reimplementing part of its functionality.

src/main/java/randoop/main/GenTests.java
```diff
@@ -1335,4 +1358,150 @@ private Path getResourceDirectoryPath(String resourceDirectory) {
...
+  /**
+   * Adds parameter types of method to collection if method is accessible by accessibility predicate
+   * and parameter type is not String, primitive, should not be omitted and is accessible by
+   * accessibility predicate
+   *
+   * @param method method
+   * @param classnames collection to add parameter types to
+   * @param accessibilityPredicate accessibility predicate
+   */
+  private static void addMethodParameterTypesIfShould(
```
mernst, line 1457: When https://github.com/randoop/randoop/pull/939 is merged, you will be able to have just one method `addParameterTypesIfShould` rather than being forced to have two overloads, one for methods and one for constructors.

src/main/java/randoop/main/GenTests.java
```diff
@@ -1335,4 +1358,150 @@ private Path getResourceDirectoryPath(String resourceDirectory) {
...
+   */
+  private static void addExecutableParameterTypesIfShould(
+      Executable executable,
+      Collection<@ClassGetName String> classnames,
+      AccessibilityPredicate accessibilityPredicate) {
+    for (Class<?> parameterType : executable.getParameterTypes()) {
+      String parameterName = parameterType.getName();
+      if (!shouldOmitClass(parameterName)
+          && !parameterType.isPrimitive()
+          && !parameterType.equals(String.class) // weird behavior when these are added
```
mernst, line 1500: Can you explain what the weird behavior is?  Can you provide a test that exposes the weird behavior?  Then perhaps I can fix the underlying issue.

src/main/java/randoop/main/GenInputsAbstract.java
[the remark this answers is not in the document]
mernst: Also, my change to the `test_add_dependencies` loop may mean that the comment isn't true even from an implementation standpoint.  (Please see my changes and let me know if they look wrong.)

mernst (changes requested):
Thanks, this is looking good.
I have made a few refactorings to simplify this pull request, and I have given suggestions about how to take advantage of them.

src/main/java/randoop/main/GenTests.java
Ivan-Kocherhin: When String class is under test, this error is continuously outputted:
```
Error selecting inputs for operation: java.lang.String.replaceAll : (java.lang.String, java.lang.String, java.lang.String) -> java.lang.String
randoop.main.RandoopBug: getSequencesForType(java.lang.String.replaceAll : (java.lang.String, java.lang.String, java.lang.String) -> java.lang.String, 0, true) neededType=java.lang.Str
ing
        at randoop.generation.ComponentManager.getSequencesForType(ComponentManager.java:186)
        at randoop.generation.ForwardGenerator.selectInputs(ForwardGenerator.java:770)
        at randoop.generation.ForwardGenerator.createNewUniqueSequence(ForwardGenerator.java:457)
        at randoop.generation.ForwardGenerator.step(ForwardGenerator.java:208)
        at randoop.generation.AbstractGenerator.createAndClassifySequences(AbstractGenerator.java:320)
        at randoop.main.GenTests.handle(GenTests.java:514)
        at randoop.main.Main.nonStaticMain(Main.java:71)
        at randoop.main.Main.main(Main.java:31)
```
Seems as error is thrown in this code inside ComponentManager
```java
    if (onlyReceivers && neededType.isNonreceiverType()) {
      throw new RandoopBug(
          String.format(
              "getSequencesForType(%s, %s, %s) neededType=%s",
              operation, i, onlyReceivers, neededType));
    }
```
So I think adding nonreceiver types should be avoided and I will change code to test if type is nonreceiver instead of testing for equality with concrete classes

src/main/java/randoop/main/GenTests.java
Ivan-Kocherhin: I agree, but this method is private method of `OperationModel` variable, so I couldn't use it. Also, using `OperationModel` for this purposes is not the best approach. So I decided to refactor `getOperationsFromFile` more and moved its logic to new class called `TypedClassOperationProvider`, so now both `GenTests` and `OperationModel` use one implementation located in a new class instead of `GenTests` using `OperationModel`.

Ivan-Kocherhin:
@mernst Sorry for bothering you. I see you added some commits, but you didn't approved PR or requested changes. Could you please update the status of PR? Thank you.

src/main/java/randoop/main/GenTests.java
mernst: Can you explain why "using `OperationModel` for this purposes is not the best approach"?  I don't think I understand the reasoning behind that conclusion.

src/main/java/randoop/main/GenTests.java
```diff
@@ -293,6 +299,24 @@ public boolean handle(String[] args) {
       classNameErrorHandler = new WarnOnBadClassName();
     }
 
+    if (test_add_dependencies) {
+      // The first two lines are in this order to avoid having getDependentClassnamesFromClassnames
+      // act on classes added by getDependentClassnamesFromMethodList.
+      classnames.addAll(getDependentClassnamesFromClassnames(classnames, accessibility));
+      classnames.addAll(getDependentClassnamesFromMethodList(accessibility, omit_methods));
```
mernst, line 306: It is inconsistent that `getDependentClassnamesFromClassnames` takes `classnames` as an argument, but `getDependentClassnamesFromMethodList` uses a field.  Method implementations and uses are easiest to understand when they do not use external state.  Could you change `getDependentClassnamesFromMethodList`?

src/main/java/randoop/main/GenTests.java
```diff
@@ -1335,4 +1359,93 @@ private Path getResourceDirectoryPath(String resourceDirectory) {
...
+   * Returns names of classes that methods in {@code methodlist} depend on. Does not add
+   * dependencies of methods that should be omitted. Does not add classes that are not accessible or
+   * should be omitted.
+   *
+   * @param accessibilityPredicate an accessibility predicate
+   * @param omitMethods each regex indicates methods that should not be called directly in generated
+   *     tests
+   * @return classnames of dependencies
+   */
+  public static Set<@ClassGetName String> getDependentClassnamesFromMethodList(
```
mernst, line 1402: Please make this method take `methodList` as an argument rather than depending on the value of a field.

src/main/java/randoop/main/GenTests.java
```diff
@@ -1335,4 +1359,93 @@ private Path getResourceDirectoryPath(String resourceDirectory) {
...
+    }
+    return classnames;
+  }
+
+  /**
+   * Adds the parameter types of {@code executable} to {@code classnames} if the parameter type is
+   * not String or primitive, should not be omitted, and is accessible by accessibility predicate.
+   *
+   * @param executable a method or constructor
+   * @param classnames collection to add parameter types to
```
mernst, line 1433: A method without side effects is easier to understand than one with side effects.  Why not have this method return a list rather than side-effecting an argument?

src/main/java/randoop/reflection/TypedClassOperationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Set;
+import org.plumelib.util.EntryReader;
+import org.plumelib.util.UtilPlume;
+import randoop.condition.SpecificationCollection;
+import randoop.main.RandoopUsageError;
+import randoop.operation.TypedClassOperation;
+import randoop.operation.TypedOperation;
+import randoop.types.ClassOrInterfaceType;
+
+/** Provides TypedClassOperation objects from given class types or methods. */
```
mernst, line 21: What does it mean to "provide" an object?  This comment doesn't say much more than the class name, so it does not explain the class to a reader.

mernst (changes requested):

src/main/java/randoop/main/GenTests.java
Ivan-Kocherhin: From my point of view `OperationModel` is not responsible for reading files and returning objects from them. It was ok when this functionality was used only inside `OperationModel`, but not when we need to perform the same logic in different places of the program. Also, this logic is needed in `GenTests` before `OperationModel` object is created. Maybe better than creating new class I should move this logic to `public static` methods, similar to `OperationModel.readOperations()`?

PR #949, opened 2021-07-01
Use Gradle 7.1.

smillst:


build.gradle
```diff
@@ -606,15 +607,6 @@ jar {
   from sourceSets.main.output
 }
 
-// want to archive locally
```
smillst, line 609: I didn't replace this with anything, because I couldn't find anywhere it was called.  It's not needed to upload artifacts to Maven central or to a local Maven repository.

mernst:
Thanks!

PR #997, opened 2022-01-04
Use the version of the Checker Framework as specified by the CF plugin.

smillst:


build.gradle
```diff
@@ -25,7 +25,7 @@ plugins {
   id "com.gorylenko.gradle-git-properties" version "2.3.2"
 }
 
-def checkerFrameworkVersion = "3.21.0" // Checker Framework version
+ext.checkerFrameworkVersion = '3.20.0' // Checker Framework version
```
smillst, line 28: I think using `ext` will cause dependabot to change the version number.  I rolled back to `3.20.0` to test this.

mernst (approved):

mernst:
I merged this so that we can see whether dependabot reacts.

PR #1005, opened 2022-01-10
Run CI under JDK 17

mernst:


markro49:
I'm pretty sure I've figured this out.  The special system used to install JDK 17 did not have the correct dependencies.   I was able to successfully run all the tests in the docker container after I installed the additional package 'x11-utils'.

mernst:
Thanks for figuring this out!

PR #1027, opened 2022-03-21
Make contracts symmetric, fixing an error in deb125d

mernst:


src/main/java/randoop/contract/EqualsHashcode.java
```diff
@@ -20,7 +20,10 @@ public boolean evaluate(Object... objects) {
     Object o1 = objects[0];
     Object o2 = objects[1];
 
-    return !o1.equals(o2) || o1.hashCode() == o2.hashCode();
+    if (o1.equals(o2)) {
+      return o1.hashCode() == o2.hashCode();
+    }
+    return true;
```
markro49, line 26: I'm unfamiliar with this part of Randoop.  Why isn't the last changed line return false?

src/main/java/randoop/contract/EqualsTransitive.java
```diff
@@ -24,7 +24,10 @@ public boolean evaluate(Object... objects) {
     Object o2 = objects[1];
     Object o3 = objects[2];
 
-    return !(o1.equals(o2) && o2.equals(o3)) || o1.equals(o3);
+    if (o1.equals(o2) && o2.equals(o3)) {
+      return o1.equals(o3);
+    }
+    return true;
   }
```
markro49, line 31: same comment

src/main/java/randoop/contract/EqualsHashcode.java
mernst: The Javadoc of `evaluate` says:

> Returns true if the contract was satisfied or was not applicable.

If two objects are `equals` to one another, their hash codes must be equal.
If two objects are not `equals` to one another, then there is no further requirement.

markro49 (approved):

PR #1114, opened 2022-12-23
Set size of collections when creating them

mernst:


markro49 (approved):
Looks good.  I did not verify that for the new collections with a fixed integer size that the integer value was correct.

PR #1185, opened 2023-05-10
Add note about jacocoagent.jar to bloodhound info

markro49:


mernst (approved):
I made the change in `GenInputsAbstract.java`.  It gets automatically copied from there to the manual.

PR #1213, opened 2023-07-07
Don't run updateDevTOC and updateUserTOC on Windows.

smillst:


mernst (approved):
This change is OK with me, but wouldn't it be better to test whether the perl program is installed?  Then this would work for any user who doesn't have it installed.

smillst:
Building the manual was already checking for Windows, so I just added more checks.

PR #1216, opened 2023-07-11
Compare arrays using `Assert.assertArrayEquals()`

varuniy:
Fixed Issue 1107: Added functionality for assertArrayEquals for arrays of literal types (primitives, boxed primitives, Strings, Classes, Enums)


varuniy:
@mernst

varuniy:
@mernst

varuniy:
@mernst

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+ * A check recording the current state of an array during execution.
+ *
+ * <p>ObserverEqArray checks are not checks that must hold of all objects of a given class (unlike a
+ * check like {@link EqualsReflexive}, which must hold for any objects, no matter its execution
+ * context).
+ */
+public final class ObserverEqArray extends ObjectContract {
+
+  /** The run-time result of calling the observer: an array of literals. */
+  public Object value;
```
mernst, line 23: I don't think these three variables need to be public.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+    if (cls == Class.class || cls == String.class || cls.isEnum()) {
+      return true;
+    }
+    return PrimitiveTypes.isBoxedPrimitive(cls) || cls.isPrimitive();
+  }
+
+  @Override
+  public String toCodeString() {
+    StringBuilder b = new StringBuilder();
+    if (Array.getLength(value) < MAX_ARRAY_LENGTH) {
```
mernst, line 82: Instead of  testing against MAX_ARRAY_LENGTH in toString, an ObserverEqArray object should not even be created if the array's length is longer than that value.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * Prints the code string of the second parameter (instantiation of array) in assertArrayEquals
+   * e.g. prints the second parameter in assertArrayEquals(var, new int[] {1,2,3})
+   *
+   * @return String that represents an instantiation of an array equal to value
+   */
+  private String printArray() {
+    String finalString = "";
+    if (value == null) {
```
mernst, line 104: Is it necessary to handle the possibility that the array is null?  Could other code in Randoop address that instead, permitting this code to be a bit shorter?  (I'm not sure about this, I'm just asking.)
When the value is null, there is no need to call an array equality method in JUnit.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+      return ArrayUtils.toObject((boolean[]) value);
+    } else {
+      return (Object[]) value;
+    }
+  }
+
+  /**
+   * Helper method that prints the components of the array
+   *
+   * @return String that represents the components of the array
```
mernst, line 145: The documentation should clarify that the output includes the enclosing "{...}".

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+    int length = Array.getLength(value);
+    if (value.getClass().getComponentType().isEnum()) {
+      for (int i = 0; i < length; i++) {
+        if (Array.get(value, i) == null) {
+          finalString += "null";
+        } else {
+          finalString += new EnumValue((Enum<?>) Array.get(value, i)).getValueName();
+        }
+        if (i < length - 1) {
+          finalString += ", ";
```
mernst, line 158: You don't need this logic.  Use StringJoiner instead.  That is also more efficient than repeated String concatenation.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+   */
+  private String printArrayComponents() {
+    String finalString = "{";
+    int length = Array.getLength(value);
+    if (value.getClass().getComponentType().isEnum()) {
+      for (int i = 0; i < length; i++) {
+        if (Array.get(value, i) == null) {
+          finalString += "null";
+        } else {
+          finalString += new EnumValue((Enum<?>) Array.get(value, i)).getValueName();
```
mernst, line 155: I suggest creating a "printArrayElement" method, so that you don't need three different for loops in this method.  (That will be a tiny bit less efficient, but such efficiency is not important and the reduction in code size, which speeds up the code due to fitting in the cache, might make up for it.)

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }
+    finalString += "}";
+    return finalString;
+  }
+
+  @Override
+  public boolean evaluate(Object... objects) throws Throwable {
+    assert objects.length == 1;
+    return Arrays.equals(toObjectArray(value), toObjectArray(objects[0]));
```
mernst, line 187: Should this use the JUnit methods?  This differs from the code as written, because it does not use a delta when comparing floating-point arrays (float and double).

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+  static TypeTuple inputTypes = new TypeTuple(Arrays.asList(JavaTypes.OBJECT_TYPE));
+
+  @Override
+  public TypeTuple getInputTypes() {
+    return inputTypes;
+  }
+
+  @Override
+  public String toCommentString() {
+    return null;
```
mernst, line 205: Please don't return null here.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  @Override
+  public String get_observer_str() {
+    return "ObserverEqArray";
+  }
+
+  @Override
+  public String toString() {
+    return "randoop.ObserverEqArray, value=" + StringsPlume.escapeJava(printArrayComponents());
```
mernst, line 215: You don't need to quote this for appearance in Java source code.
Also, I would simplify this to "ObserverEqArary(" + printArrayComponents() + ")".

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -13,12 +13,7 @@
 import randoop.ExecutionOutcome;
 import randoop.NormalExecution;
 import randoop.NotExecuted;
-import randoop.contract.EnumValue;
-import randoop.contract.IsNotNull;
-import randoop.contract.IsNull;
-import randoop.contract.ObjectContract;
-import randoop.contract.ObserverEqValue;
-import randoop.contract.PrimValue;
+import randoop.contract.*;
```
mernst, line 16: Please do not make gratuitous changes to import statements.  The style of Randoop is not to use wildcards.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -215,6 +222,22 @@ public RegressionChecks generateTestChecks(ExecutableSequence eseq) {
     return checks;
   }
 
+  /**
+   * Returns true if the class (representing an array type) is a literal.
+   *
+   * @param cls -- the class to be tested
+   * @return true iff the class is a primitive, boxed primitive, String, Class, or Enum
+   */
+  private boolean isLiteralType(Class<?> cls) {
```
mernst, line 231: Don't define this twice.  Eliminate it here or in ObserverEqArray.java.
Also, the definitions differ because in ObserverEqArray.java, no test for accessibility is performed.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,217 @@
...
+                printArray(), FLOATING_POINT_DELTA));
+      } else {
+        b.append(String.format("org.junit.Assert.assertArrayEquals(x0, %s);", printArray()));
+      }
+    }
+    return b.toString();
+  }
+
+  /**
+   * Prints the code string of the second parameter (instantiation of array) in assertArrayEquals
```
mernst, line 97: End sentences with punctuation.  (But don't end `@param` and `@return` clauses -- which are not full sentences -- with a period unless a full sentence follows.)

mernst (changes requested):
This is looking good.  Thanks!

I have some suggestions for improving it further.

By the way, please make pull request titles descriptive, not just "Fix Issue 1107".
There is a "Development" item in the right-hand column of the pull request webpage where you can link to issues that the pull request fixes.  (I did that already.)

src/main/java/randoop/contract/ObserverEqArray.java
varuniy: Do you mean that I should use assertArrayEquals? If so, should I return "true" or "false" depending on if an assertion exception is thrown? And should I change this for the equals() method too?

mernst: Yes, use assertArrayEquals as you said in your comment.

Do not change the equals() method.  Doing so would make equals() non-transitive and violate the contract of equals().

mernst:
@varuniy If you are done with your work, please un-assign yourself.  If you are still assigned, I'll assume that you still have some work to do and the PR is not blocked on me.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,205 @@
...
+      return new EnumValue((Enum<?>) element).getValueName();
+    } else if (element.getClass() == Class.class) {
+      return ((Class<?>) element).getCanonicalName() + ".class";
+    } else {
+      return Value.toCodeString(element);
+    }
+  }
+
+  @Override
+  public boolean evaluate(Object... objects) throws Throwable {
```
mernst, line 166: This method should return true or false, rather than throwing an exception.

mernst (changes requested):

src/main/java/randoop/contract/ObserverEqArray.java
varuniy: Randoop fails if I try to surround these statements in a try-catch block because it warns that I am ignoring an AssertionError. How should I get around this?

varuniy: I just added a @SuppressWarnings("AssertionFailureIgnored") tag. Let me know if you think this is the wrong fix, or there is a better way.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * Creates a new ObserverEqArray.
+   *
+   * @param value the run-time result of calling the observer: an array of literals
+   * @param isAccessible the accessibility predicate
+   */
+  public ObserverEqArray(Object value, AccessibilityPredicate isAccessible) {
+    this.value = value;
+    if (!isLiteralType(value.getClass().getComponentType(), isAccessible)) {
```
mernst, line 39: It is also acceptable for the array to consist only of `null` values.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * Returns true if values of the class can be represented as literals in Java source code.
+   *
+   * @param cls a class
+   * @param isAccessible the accessibility predicate
+   * @return true iff the class is a primitive, boxed primitive, String, Class, or accessible Enum
+   */
+  public static boolean isLiteralType(Class<?> cls, AccessibilityPredicate isAccessible) {
+    if (cls == Class.class || cls == String.class) {
```
mernst, line 55: Should the `Class.class` case also be associated with an `isAccessible()` test, as the `isEnum()` case is?  (I don't know the answer to this, but if the answer is "no", please document that fact in a comment.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+    if (!(o instanceof ObserverEqArray)) {
+      return false;
+    }
+    ObserverEqArray other = (ObserverEqArray) o;
+    return Arrays.equals(toObjectArray(value), toObjectArray(other.value));
+  }
+
+  @Override
+  public int hashCode() {
```
mernst, line 78: The `Arrays` class contains hash code routines for every primitive type.  When the array type is primitive, it is more efficient to use those, which don't require the construction of a new array.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,211 @@
...
+    return Arrays.hashCode(toObjectArray(value));
```
mernst, line 79: Because of Java's (statically unsound) array subtyping rules, you don't need to call `toObjectArray()`.

In fact, the body can be just

```java
Arrays.hashCode(value);
```

By analogy, this code compiles and runs:
```java
import java.util.Arrays;
import java.util.Date;

public class ArraySubtyping {

  public static void main(String[] args) {
    Date[] da = new Date[] {new Date()};
    test(da);
  }

  public static void test(Date[] da) {
    Arrays.hashCode(da);
  }

  public static void test(double[] da) {
    Arrays.hashCode(da);
  }
}
```

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,211 @@
...
+    return "new " + value.getClass().getCanonicalName() + arrayComponentsToString();
+  }
+
+  /**
+   * Converts an Object into its array representation.
+   *
+   * @param value the given Object
+   * @return its array representation
+   */
+  private Object[] toObjectArray(Object value) {
```
mernst, line 109: I don't think this method is necessary.  Overloads of the methods that you are calling should statically dispatch to the correct implementation, without creating any new arrays.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+    return sj.toString();
+  }
+
+  /**
+   * Returns one literal value, as it would appear in Java code
+   *
+   * @return the Java code for the literal value
+   */
+  private String literalValueToString(Object element) {
```
mernst, line 151: I don't think this new method is necessary.  It's better to extend `Value.toCodeString()` and then use it in `arrayComponentsToString()`.

src/main/java/randoop/contract/ObserverEqArray.java
mernst: Always write a justifying comment on any `@SuppressWarnings` annotation.  (I did this.)

mernst (changes requested):

src/main/java/randoop/contract/ObserverEqArray.java
varuniy: When I was changing Value's toCodeString() method to include Classes and Enums, I noticed that there was a method in the Value class called arrayLengthOk(Object a) that "returns true if the given array is shorter than the --string-maxlen=N parameter." Should I use this method instead of the constant MAX_ARRAY_LENGTH in RegressionCaptureGenerator?

src/main/java/randoop/contract/ObserverEqArray.java
varuniy: Not sure if this answers your question, but isn't Class.class always accessible? (built-in)

mernst (approved):
Thanks for this enhancement!

PR #1239, opened 2023-09-05
Unify Checker Framework versions.

smillst:
Make sure the the Checker Framework plugin uses the same version of the Checker Framework that is used as a dependencies of the project.  This should fix the failures in #1238 and #1237.


mernst (approved):
Thanks!

PR #1257, opened 2023-11-14
Cleanup ClassLiteralExtractor

Kuro1Fury:


src/main/java/randoop/util/ClassFileConstants.java
```diff
@@ -662,31 +673,53 @@ static void longConstant(Long value, ConstantSet cs) {
...
-      for (Class<?> x : cs.classes) {
-        map.add(clazz, new NonreceiverTerm(JavaTypes.CLASS_TYPE, x));
-      }
+      addToConstantMap(cs, map);
     }
     return map;
   }
+
+  /**
+   * Add the constants in a ConstantSet to the given map.
```
mernst, line 682: Are they added as keys?  As values?  If keys, what are the corresponding values?

mernst (changes requested):
This looks good.  I have a question before it is merged.

src/main/java/randoop/util/ClassFileConstants.java
Kuro1Fury: The comment is fixed: "Add all constant values from the given ConstantSet as NonreceiverTerms to their corresponding class in the given map."

mernst:
Thanks!

PR #1258, opened 2023-11-15
Create tests for nonaccessible methods

varuniy:
Randoop can now call inaccessible methods using reflection. If you wish to utilize this behavior, use the command-line argument "--use-reflection=true". Future developers can expand on this feature by adding similar behavior for inaccessible constructors, enums, and more.


src/main/java/randoop/operation/MethodCall.java
```diff
@@ -92,45 +100,102 @@ public void appendCode(
...
+   * MethodCall}. The variable name is formed by appending the method name and the simple names of
+   * the parameter types, separated by underscores. For array types, "Array" is appended instead of
+   * "[]".
+   *
+   * @return the unique variable name
+   */
+  private String getVariableNameForMethodObject() {
+    StringBuilder signature = new StringBuilder();
+    // Append method name
+    signature.append(getMethod().getName());
```
mernst, line 191: This isn't necessarily unique.  There might be a method named "foo" in multiple classes (especially with overriding).  Should you add the class name?

src/main/java/randoop/operation/MethodCall.java
```diff
@@ -92,45 +100,102 @@ public void appendCode(
...
+    // Append parameter types
+    Class<?>[] parameterTypes = getMethod().getParameterTypes();
+    for (int i = 0; i < parameterTypes.length; i++) {
+      signature.append("_");
+      signature.append(parameterTypes[i].getSimpleName().replace("[]", "Array"));
```
mernst, line 196: The type simple name might not be unique.  There could be two (overloaded) methods, with signatures:
 * foo(java.sql.Date)
 * foo(java.util.Date)

src/main/java/randoop/operation/TypedOperation.java
```diff
@@ -367,9 +368,11 @@ public static TypedClassOperation forConstructor(Constructor<?> constructor) {
    * Constructs a {@link TypedOperation} for a method object.
    *
    * @param method the reflective method object
+   * @param accessibilityPredicate AccessibilityPredicate to check if the method is accessible
```
mernst, line 371: You don't need to repeat the parameter's type, which is apparent nearby, in both the code and the generated API documentation.

src/main/java/randoop/output/JUnitCreator.java
```diff
@@ -258,6 +259,27 @@ public CompilationUnit createTestClass(
       }
     }
 
+    if (!Globals.makeAccessibleMap.isEmpty()) {
+      StringBuilder sb = new StringBuilder();
```
mernst, line 264: Please use a more descriptive variable name than `sb`.

mernst (changes requested):
Overall, this looks good.  Here are a few things that can be improved.

varuniy:
Hi Mike,

I have implemented your requested changes. However, for the method that does the variable name generation, isn't it quite likely that the variable name will end up being super long? If so, will Randoop automatically omit these from generated tests? Should I check to see if the length of a method variable fits a given criteria before I even create the MethodCall object?

Thanks!

src/main/java/randoop/test/EmptyExceptionCheck.java
```diff
@@ -26,7 +26,9 @@ public class EmptyExceptionCheck extends ExceptionCheck {
    * <p>Appends comment indicating that catch is being ignored.
```
mernst, line 26: The `appendCatchBehavior()` changes seem independent of the other changes in this pull request.  Can they be pull-requested separately, so that they can be merged faster and so that this pull request is simpler?

src/docs/manual/index.html
```diff
@@ -975,6 +975,9 @@ <h2 id="command-line-options">Command-line options</h2>
  <p>Using <code>--methodlist</code> is less common, and more error-prone, than <code>--testjar</code>,
  <code>--test-package</code>, <code>--classlist</code>, or <code>--testclass</code>. See the notes about <a href="https://randoop.github.io/randoop/manual/#specifying-methods">specifying methods that may
  appear in a test</a>.
+            <li id="option:use-reflection"><b>--use-reflection=</b><i>boolean</i>.
```
mernst, line 979: This pull request adds a command-line option `--use-reflection` that is adjacent to `--omit-classes`, but I think it should be next to `--only-test-public-members` which it is more similar to.  Also, let's improve the name `--use-reflection` which is an implementation detail.  The real question is whether private, protected, and or package-private methods should be called.
So maybe there should be an enum indicating how many methods should be invoked:
 1. public only
 2. public + some package-private (based on `--junit-package-name`)
 3. public + package-private
 4. public + package-private + protected
 5. public + package-private + protected + private

The current Randoop code supports #1 and #2.
Your code adds #5, though I don't think that `--only-test-public-members` is respected if `--use-reflection` is passed.  Having a single command-line argument will prevent that kind of inconsistency.  (We will keep the `--only-test-public-members` command-line argument for a while, for backward compatibility.)

It's true that for protected and private methods (and some package-private ones), they will have to be called reflectively.

In MethodCall, I've renamed the `isAccessible` boolean field to `callReflectively`, which I think is what it is really about.  I think that will make the code clearer.

Another way of saying the above is that the current code conflates "accessibility" with *whether* a method should be called.  They used to be the same, before your extended functionality, but now they should be separated for clarity:
 * which methods should be called.
 * whether a method must be called reflectively.

mernst (changes requested):

mernst:
@varuniy Ping.

PR #1304, opened 2024-03-16
Fuzzing for numbers & strings ("impurity")

776styjsu:


mernst:
@776styjsu Please ping me when this is ready for a code review.  Thanks!

776styjsu:
@mernst This is ready to be reviewed.

mernst:
@776styjsu Could you fix the "misc" CI job?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    // Do not fuzz void, char, boolean, or byte.
+    if (outputType.isVoid()
+        || outputType.runtimeClassIs(char.class)
+        || outputType.runtimeClassIs(boolean.class)
+        || outputType.runtimeClassIs(byte.class)) {
+      return new ImpurityAndNumStatements(sequence, 0);
+    }
+
+    Class<?> outputClass = outputType.getRuntimeClass();
+    List<Method> methodList = new ArrayList<>();
```
mernst, line 77: Please document and/or use a more descriptive name.  `methodList` gives no more information than the type.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    try {
+      if (outputClass.isPrimitive()) { // fuzzing primitive numbers
+        sequence = getFuzzedSequenceForPrimNumber(sequence, outputClass);
```
mernst, line 80: I would be inclined to write out "Primitive", since the identifier is so long already.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+      } else if (outputClass == null) {
+        throw new RuntimeException("Output class is null");
+      }
+    } catch (NoSuchMethodException e) {
+      throw new RuntimeException("Initialization failed due to missing method", e);
+    }
+
+    Sequence output = sequence;
+    for (int i = 0; i < methodList.size() - 1; i++) {
+      Method method = methodList.get(i);
```
mernst, line 104: Use a foreach loop rather than an indexed loop.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+            methodList.get(methodList.size() - 1),
+            outputType,
+            fuzzStatementOffset,
+            outputType.runtimeClassIs(short.class));
+
+    return new ImpurityAndNumStatements(output, fuzzStatementOffset.getOffset());
+  }
+
+  /**
+   * Create a sequence for fuzzing an object of a given type using the given method. This overload
```
mernst, line 120: What is the "given type"?  Does it have some relation to `executable` or to `sequence`?  I have this same question for many uses of "given type".

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   * @return a sequence with the fuzzing statement appended at the end
+   */
+  private static Sequence createSequence(
+      Sequence sequence, Executable executable, FuzzStatementOffset fuzzStatementOffset) {
+    Type outputType = determineOutputType(executable);
+    return createSequence(sequence, executable, outputType, fuzzStatementOffset, false);
+  }
+
+  /**
+   * Create a sequence for fuzzing an object of a given type using the given method. This overload
```
mernst, line 135: What is the "given type"?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   * @param sequence the sequence to append the fuzzing sequence to
+   * @param executable the method to be invoked to fuzz the object
+   * @param outputType the type of the object to be fuzzed
+   * @param fuzzStatementOffset the offset counter for the number of fuzzing statements added
+   * @param explicitCast whether to perform an explicit cast for the right-hand side of the fuzzing
+   *     statement
+   * @return a sequence with the fuzzing statement appended at the end
+   * @throws IllegalArgumentException if the method is not a method of the given type
+   */
+  private static Sequence createSequence(
```
mernst, line 147: The name "createSequence" is too generic.  Please use a more descriptive name that communicates its purpose or use to readers.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    }
+  }
+
+  /**
+   * Determine the output type of the given method.
+   *
+   * @param executable the method to determine the output type of
+   * @return the output type of the given method
+   */
+  private static Type determineOutputType(Executable executable) {
```
mernst, line 188: "determine" doesn't convey anything.  I would change the method name to `outputType()` or `getOutputType()`.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   */
+  private static List<Type> getInputTypeList(
+      Executable executable, NonParameterizedType declaringType) {
+    List<Type> inputTypeList = new ArrayList<>();
+    if (!Modifier.isStatic(executable.getModifiers()) && executable instanceof Method) {
+      inputTypeList.add(declaringType);
+    }
+    for (Class<?> clazz : executable.getParameterTypes()) {
+      inputTypeList.add(
+          clazz.isPrimitive() ? PrimitiveType.forClass(clazz) : new NonParameterizedType(clazz));
```
mernst, line 215: This appears a couple of times.  Should it be abstracted out?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    }
+    return inputTypeList;
+  }
+
+  /**
+   * Calculate the index of the input parameters in the given sequence.
```
mernst, line 221: This makes the assumption that "the input parameters" are the last _n_ statements in the sequence.  When there is an important precondition like that, please state it in the documentation.  Also, what does "the input parameters" mean?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+  private static List<Integer> calculateInputIndex(Sequence sequence, List<Type> inputTypeList) {
+    List<Integer> inputIndex = new ArrayList<>();
+    for (int i = 0; i < inputTypeList.size(); i++) {
+      inputIndex.add(sequence.size() - inputTypeList.size() + i);
+    }
+    return inputIndex;
+  }
+
+  /**
+   * Get a sequence with the fuzzing statement appended at the end for fuzzing a primitive number.
```
mernst, line 236: What primitive number is being fuzzed?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   *
+   * @param outputClass the class of the primitive number to be fuzzed
+   * @return a fuzzed value for the primitive number
+   */
+  private static Object getFuzzedValueForPrim(Class<?> outputClass) {
+    double randomGaussian = GAUSSIAN_STD * Randomness.nextRandomGaussian(1);
+    if (outputClass == int.class) {
+      return (int) Math.round(randomGaussian);
+    } else if (outputClass == short.class || outputClass == byte.class) {
+      return (short) Math.round(randomGaussian); // Unified handling for short and byte
```
mernst, line 261: Earlier, the class said that `byte` was not supported.  This seems to contradict that.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+      return (float) randomGaussian;
+    } else if (outputClass == double.class) {
+      return randomGaussian;
+    } else {
+      throw new RuntimeException("Unexpected primitive type: " + outputClass.getName());
+    }
+  }
+
+  /**
+   * Get the method (in a list) that can be used to fuzz primitive numbers.
```
mernst, line 274: If this always returns a singleton list, why does it return a list at all?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    } else if (outputClass == double.class) {
+      methodList.add(Double.class.getMethod("sum", double.class, double.class));
+    } else if (outputClass == float.class) {
+      methodList.add(Float.class.getMethod("sum", float.class, float.class));
+    } else if (outputClass == long.class) {
+      methodList.add(Long.class.getMethod("sum", long.class, long.class));
+    } else if (outputClass == short.class) {
+      methodList.add(Integer.class.getMethod("sum", int.class, int.class));
+    } else if (outputClass == byte.class) {
+      throw new NoSuchMethodException("Byte fuzzing is not supported yet");
```
mernst, line 297: `NoSuchMethodException` is intended for problems with dynamic dispatch, not for situations like this.  Use `RandoopBug` or `Error` or the like.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+          "Unable to find suitable method for class: "
+              + outputClass.getName()
+              + " in primitive number fuzzing");
+    }
+
+    return methodList;
+  }
+
+  /**
+   * Get a sequence with the fuzzing statement appended at the end for fuzzing a String.
```
mernst, line 314: This says "the fuzzing statement", but two lines below says "the fuzzing sequence".

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   *
+   * @param sequence the sequence to append the fuzzing sequence to
+   * @param fuzzingOperationIndex the index of the fuzzing operation to perform
```
mernst, line 317: The index in what?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   */
+  private static Sequence getFuzzedSequenceForString(
+      Sequence sequence, int fuzzingOperationIndex, FuzzStatementOffset fuzzStatementOffset)
+      throws IllegalArgumentException, IndexOutOfBoundsException {
+    sequence = initializeWithStringBuilder(sequence, fuzzStatementOffset);
+
+    Object stringValue = getStringValue(sequence);
+    int stringLength = stringValue.toString().length();
+
+    if (stringLength == 0) {
```
mernst, line 331: Since this is a precondition for callers, it should be stated in the method documentation.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    List<Sequence> fuzzingSequenceList =
+        performFuzzingOperation(fuzzingOperationIndex, stringLength);
+
+    List<Sequence> temp = new ArrayList<>(Collections.singletonList(sequence));
+    temp.addAll(fuzzingSequenceList); // Assuming these are sequences that need to be concatenated
+    return Sequence.concatenate(temp); // Concatenate all sequences together
+  }
+
+  /**
+   * Initialize a sequence with a StringBuilder object.
```
mernst, line 345: What does it mean to "initialize a sequence"?  Does this create a new singleton sequence containing `new StringBuffer()`?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   *
+   * @param sequence the sequence to initialize with a StringBuilder object
+   * @param fuzzStatementOffset the offset counter for the number of fuzzing statements added
+   * @return a sequence with the fuzzing statement appended at the end
```
mernst, line 349: What is "the fuzzing statement"?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+  /**
+   * Get the String value from the given sequence.
+   *
+   * @param sequence the sequence to get the String value from
+   * @return the String value from the given sequence
+   * @throws IllegalArgumentException if the String value cannot be obtained
+   */
+  private static Object getStringValue(Sequence sequence) {
+    try {
+      // Original String value is the 2nd last statement in the sequence
```
mernst, line 371: Please explain how you know this.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+      // This does not indicate error, ignore this fuzzing operation.
+      throw new IllegalArgumentException(
+          "Error obtaining String value. Will ignore this" + " fuzzing operation.", e);
+    }
+  }
+
+  /**
+   * Perform a fuzzing operation on a String.
+   *
+   * @param operationIndex the index of the fuzzing operation to perform
```
mernst, line 384: It will make the code more readable and less error-prone if this is an `enum`.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+  private static List<Sequence> performFuzzingOperation(int operationIndex, int stringLength) {
+    switch (operationIndex) {
+      case 0:
+        return Collections.singletonList(fuzzInsertCharacter(stringLength));
+      case 1:
+        return Collections.singletonList(fuzzRemoveCharacter(stringLength));
+      case 2:
+        return fuzzReplaceCharacter(stringLength);
+      case 3:
+        return fuzzSelectSubstring(stringLength);
```
mernst, line 397: Why do 2 of these 4 methods return a list, but the other two return a list element?  Please be consistent.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+    Sequence startIndexSequence = Sequence.createSequenceForPrimitive(startIndex);
+    Sequence endIndexSequence = Sequence.createSequenceForPrimitive(endIndex);
+    return Arrays.asList(startIndexSequence, endIndexSequence);
+  }
+
+  /**
+   * Get the method (in a list) that can be used to fuzz strings.
+   *
+   * @param stringFuzzingStrategyIndex the index of the fuzzing strategy to use
+   * @return a list of methods that can be used to fuzz strings
```
mernst, line 468: Here, can the method be used to fuzz strings, or can the list be used to fuzz strings?

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+
+    if (methodList.isEmpty()) {
+      // Should be unreachable
+      throw new NoSuchMethodException("No suitable method found for class String");
+    }
+
+    return methodList;
+  }
+
+  /** A helper class to store the result of the Impurity component. */
```
mernst, line 498: Isn't the result an extended sequence?

src/main/java/randoop/util/EquivalenceChecker.java
```diff
@@ -0,0 +1,50 @@
...
+    map.put(byte.class, Byte.class);
+    map.put(char.class, Character.class);
+    map.put(double.class, Double.class);
+    map.put(float.class, Float.class);
+    map.put(int.class, Integer.class);
+    map.put(long.class, Long.class);
+    map.put(short.class, Short.class);
+    map.put(void.class, Void.class);
+    PRIMITIVE_TO_BOXED = Collections.unmodifiableMap(map);
+  }
```
mernst, line 27: Replace this by a use of `PrimitiveTypes.toBoxedType()`.

src/main/java/randoop/util/EquivalenceChecker.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Determines if two Class objects represent equivalent types, considering both primitive types
+   * and their boxed counterparts as equivalent. For instance, int.class and Integer.class are
+   * considered equivalent.
+   *
+   * @param c1 the first class to compare
+   * @param c2 the second class to compare
+   * @return true if the classes represent equivalent types, false otherwise
+   */
+  public static boolean equivalentTypes(Class<?> c1, Class<?> c2) {
```
mernst, line 38: "equivalent" is a vague name.  Make it more descriptive, even though it must be longer.

mernst (changes requested):
Please add a test of the feature.

src/main/java/randoop/generation/Impurity.java
```diff
@@ -0,0 +1,518 @@
...
+   * @return a sequence with additional fuzzing statements appended at the end, and a count of the
+   *     number of fuzzing statements added to the sequence
+   */
+  public static ImpurityAndNumStatements fuzz(Sequence sequence) {
+    // A counter to keep track of the number of fuzzing statements added to the sequence
+    FuzzStatementOffset fuzzStatementOffset = new FuzzStatementOffset();
+
+    Type outputType = sequence.getLastVariable().getType();
+
+    // Do not fuzz void, char, boolean, or byte.
```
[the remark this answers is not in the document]
776styjsu: The boxed primitive does not yet enter GrtImpurity, so no boxed version is passed in the call to fuzz. I will add handling for these cases soon.

776styjsu:
@mernst I updated the code as you suggested. Please review it. Thanks!

src/main/java/randoop/compile/SequenceCompiler.java
```diff
@@ -173,6 +173,15 @@ private boolean compile(
         compiler.getTask(
             null, fileManager, diagnostics, new ArrayList<String>(compilerOptions), null, sources);
     Boolean succeeded = task.call();
+
+    //    // Print the diagnostics if compilation failed
```
mernst, line 177: Please do one of:
 * Make this unconditional
 * Put this behind an "if" statement.
 * Delete this.

src/main/java/randoop/contract/CompareToEquals.java
```diff
@@ -10,10 +10,13 @@
  * <pre>(x0.compareTo(x1) == 0) == x0.equals(x1)</pre>
  */
 public class CompareToEquals extends ObjectContract {
+  /** The singleton instance of this class. */
   private static final CompareToEquals instance = new CompareToEquals();
 
-  private CompareToEquals() {}
+  /** Singleton pattern. */
```
mernst, line 16: The documentation should be something like "Creates the singleton CompareToEquals; is only ever called once.  Clients should call {@link #getInstance}."

src/main/java/randoop/contract/CompareToEquals.java
```diff
@@ -10,10 +10,13 @@
...
+  protected CompareToEquals() {}
```
mernst, line 17: Why is this protected?  What other method needs to call it?  Shouldn't all clients call `getInstance`?

src/main/java/randoop/generation/GrtImpurityAndNumStatements.java
```diff
@@ -0,0 +1,31 @@
...
+
+import randoop.sequence.Sequence;
+
+/**
+ * Represents the result of fuzzing a sequence of statements, and the number of additional
+ * statements that were used to fuzz the inputs.
+ */
+class GrtImpurityAndNumStatements {
+
+  /** The sequence that create the inputs. */
```
mernst, line 11: What "inputs"?  Relate this to fuzzing.

Nit: "create" should be "creates".

src/main/java/randoop/generation/GrtImpurityAndNumStatements.java
```diff
@@ -0,0 +1,31 @@
...
+  public Sequence sequence;
+
+  /**
+   * The number of additional statements that were executed to fuzz the inputs. These fuzzing
+   * statements are the last `numStatements` statements in the sequence. If `numStatements` is 0,
```
mernst, line 16: Good comment.
Nit: each occurrence of `` `numStatements` `` should be `{@code numStatements}`.

src/main/java/randoop/generation/GrtImpurityAndNumStatements.java
```diff
@@ -0,0 +1,31 @@
...
+   * then the sequence does not include any fuzzing statements.
+   */
+  public int numStatements;
+
+  /**
+   * Creates a new GrtImpurityAndNumStatements object.
+   *
+   * @param sequence the sequence that create the inputs.
+   * @param numStatements the number of additional statements that were executed to fuzz the inputs
```
mernst, line 25: This is confusing.  Does `sequence` *create* the inputs, or does it *fuzz* the inputs?  Please clarify.

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -148,8 +149,23 @@ public RegressionChecks generateTestChecks(ExecutableSequence eseq) {
...
-            PrimValue.EqualityMode equalityMode =
-                var.getType().isPrimitive() ? EQUALSEQUALS : EQUALSMETHOD;
+            PrimValue.EqualityMode equalityMode;
+
+            // GRT Impurity option's short fuzzing strategy results in ambiguous
+            // assertions for primitive numbers. Use == for primitives since
+            // Java automatically unboxes them when comparing with ==, which
+            // removes the ambiguity.
+            // See randoop.generation.GrtImpurity for more details.
+            if (GenInputsAbstract.impurity) {
```
mernst, line 159: It looks to me like this code is no different depending on the value of `impurity` (which should probably be named `grtImpurity` for clarity).

PrimValue.EqualityMode only applies to `String`s and primitives.  The two branches seem to set `equalityMode` identically, just using syntactically different tests.

Does this code have an effect?

mernst (changes requested):
I have a few preliminary comments.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+import randoop.types.NonParameterizedType;
+import randoop.types.Type;
+import randoop.types.TypeTuple;
+import randoop.util.Randomness;
+
+/**
+ * Implements the "GRT Impurity" component, as described in "GRT: Program-Analysis-Guided Random
+ * Testing" by Ma et. al (ASE 2015): https://people.kth.se/~artho/papers/lei-ase2015.pdf.
+ *
+ * <p>The GRT Impurity component is a fuzzing mechanism that alters the states of (or creates new)
```
mernst, line 29: This is an accurate comment.  Since it is a fuzzing mechanism, I suggest renaming this to "GRT Fuzzing" instead of "GRT Immutability".  This introduces an unfortunate terminology difference with the paper, but the paper's terminology is terrible.  This class should clearly state that it is an implementation of whan is called "Immutability" in the GRT paper.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+   * states and potentially improve coverage for the unit test generated by Randoop by triggering
+   * more branches.
+   *
+   * <p>Different types have different fuzzing strategies:
+   *
+   * <ul>
+   *   <li><b>Numbers (int, short, long, float, double, and their wrapper classes):</b> Fuzzed using
+   *       a Gaussian distribution with the original value as the mean (mu). The fuzzed value is
+   *       generated as mu + N(0, GAUSSIAN_STD). For short values, if the fuzzed value exceeds the
+   *       short range, it is truncated.
```
mernst, line 79: I don't see code for that truncation in this pull request.  (A narrowing conversion just throws away the higher-order bits.)
Also, why do other numbers overflow but this one is truncated?  That is inconsistent.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+   * <p>This method does not fuzz void, char, boolean, or byte types.
+   *
+   * @param sequence the (non-null) sequence to fuzz
+   * @return a sequence consisting of {@code sequence} with additional fuzzing statements appended
+   *     at the end, and a count of the number of fuzzing statements added. If no fuzzing statements
+   *     are added, the original sequence is returned, and the count is 0.
+   */
+  public static GrtImpurityAndNumStatements fuzz(Sequence sequence) {
+    // The number of fuzzing statements added to the sequence.
+    FuzzStatementOffset fuzzStatementOffset = new FuzzStatementOffset();
```
mernst, line 95: The documentation says it's a number, but the variable name uses "offset".  This is inconsistent.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+        || outputType.runtimeClassIs(byte.class)
+        || outputType.runtimeClassIs(Byte.class)) {
+      return new GrtImpurityAndNumStatements(sequence, 0);
+    }
+
+    Class<?> outputClass = outputType.getRuntimeClass();
+    List<Executable> fuzzingOperations = new ArrayList<>();
+
+    // Append statements for fuzzing operations based on the type of the output object,
+    // then get the fuzzing operations.
```
mernst, line 114: I don't know what it means to "Append statements for fuzzing operations" and "then get the fuzzing operations."  How are those different?

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+    Sequence output = sequence;
+
+    // TODO: Implement cast to improve short fuzzing output readability
+    //  (e.g. cast to Integer for Integer.sum when fuzzing short can make Integer.valueOf()
+    //  unnecessary).
+
+    // Append fuzzing operations to the output sequence.
+    Iterator<Executable> iterator = fuzzingOperations.iterator();
+    while (iterator.hasNext()) {
+      Executable executable = iterator.next();
```
mernst, line 147: Why isn't this a foreach loop (a Java enhanced for loop)?

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+    return new GrtImpurityAndNumStatements(output, fuzzStatementOffset.offset);
+  }
+
+  /**
+   * Create a new sequence with a fuzzing operation statement appended to the given sequence.
+   *
+   * @param sequence the sequence to append the fuzzing operations to
+   * @param fuzzingOperation the method to be invoked to fuzz the object
+   * @param outputType the output type of the fuzzing operation
+   * @param fuzzStatementOffset the offset counter for the number of fuzzing statements added
```
mernst, line 162: Explain the semantics of "offset counter", or drop that part of the comment and just say "the number of fuzzing statements added".
Is this the number of fuzzing statements added so far?  Added ever?  Is it side-effected by this method?

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+   * Get the list of input types for the given executable.
+   *
+   * @param executable the executable to get the input types of
+   * @param declaringType the type that declares the given executable
+   * @return the list of input types for the given executable
+   */
+  private static List<Type> getInputTypeList(
+      Executable executable, NonParameterizedType declaringType) {
+    List<Type> inputTypeList = new ArrayList<>();
+    if (!Modifier.isStatic(executable.getModifiers()) && executable instanceof Method) {
```
mernst, line 217: What about inner class constructors?

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+      inputTypeList.add(declaringType);
+    }
+    for (Class<?> cls : executable.getParameterTypes()) {
+      inputTypeList.add(Type.forClass(cls));
+    }
+    return inputTypeList;
+  }
+
+  /**
+   * Calculate the indices of statements in the sequence that are the inputs for the fuzzing
```
mernst, line 227: What is "the fuzzing operation"?  None is mentioned here.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+   * operation. Precondition: The input statements are the last n statements in the sequence, where
```
mernst, line 228: Does the fuzzing operation take multiple inputs, or are there multiple fuzzing operations?

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+   * n is the number of input types.
+   *
+   * @param sequence the sequence to calculate the input indices of
+   * @param inputTypeListSize the number of input types of the given executable
+   * @return the indices of statements in the sequence that are the inputs for the fuzzing operation
```
mernst, line 233: Note that it is a sequential list of numbers, here or above.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+
+  /**
+   * Get a list of sequences that represent the inputs for the fuzzing operation for String.
+   *
+   * @param operation the String fuzzing operation to perform
+   * @param stringLength the length of the string to be fuzzed, for generating valid random indices
+   * @return a list of sequences that represent the inputs for the fuzzing operation
+   * @throws IllegalArgumentException if an invalid enum value is passed
+   */
+  private static List<Sequence> getStringFuzzingInputs(
```
mernst, line 380: Why does this return a list of sequences rather than a single sequence?

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+
+  /**
+   * Get the String value from the given sequence. Precondition: The String value is the last
+   * statement in the sequence.
+   *
+   * @param sequence the sequence to get the String value from
+   * @return the String value from the given sequence
+   * @throws IllegalArgumentException if the String value cannot be obtained
+   */
+  private static Object getStringValue(Sequence sequence) {
```
mernst, line 363: Explain why this doesn't return String (or make it return String).

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+    try {
+      return sequence.getStatement(sequence.size() - 1).getValue();
+    } catch (Exception e) {
+      throw new IllegalArgumentException(
+          "Invalid sequence, last statement does not have a String value");
```
mernst, line 368: There wasn't a test that the value is String, but this error message says that was the problem.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+    }
+  }
+
+  /**
```
mernst, line 372: Make the next 6 methods after this one be instance methods in enum values.  This method can probably be dispensed with, except as an abstract method on the enum class itself.

src/main/java/randoop/generation/GrtImpurity.java
```diff
@@ -0,0 +1,513 @@
...
+        break;
+      default:
+        throw new IllegalArgumentException(
+            "Invalid enum value was passed to getStringFuzzingMethod: " + operation);
+    }
+
+    return methodList;
+  }
+
+  /**
```
mernst, line 494: Use the JDK's `AtomicInteger` rather than inventing a new class.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -701,6 +701,30 @@ public enum BloodhoundCoverageUpdateMode {
   @Option("Maximum length of Strings in generated tests")
   public static int string_maxlen = 1000;
 
+  /**
+   * The "GRT Impurity" technique from the GRT paper attempts to fuzz the inputs of methods used in
+   * tests. Randoop by default only start with a fixed set of primitive/String inputs to use as
+   * arguments to methods. When GRT Impurity is enabled, Randoop will attempt to fuzz the
+   * primitive/String inputs, improving coverage while making the tests less repetitive.
```
mernst, line 708: How does this make the tests less repetitive?

src/main/java/randoop/util/EquivalenceChecker.java
```diff
@@ -0,0 +1,30 @@
+package randoop.util;
+
+import randoop.types.PrimitiveTypes;
+
+/**
+ * Utility class for checking equivalence between objects in randoop.generation.Detective. Mainly
```
mernst, line 6: I guess it is used by fuzzing as well as detective, or you wouldn't have included it in this pull request.

src/systemTest/java/randoop/main/RandoopSystemTest.java
```diff
@@ -1922,6 +1922,23 @@ public void runJDKSpecificationsTest() {
         testEnvironment, options, ExpectedTests.SOME, ExpectedTests.NONE, coverageChecker);
   }
 
+  /** Test GRT Impurity component {@link randoop.generation.GrtImpurity} */
+  @Test
+  public void runGrtImpuritySeedCollectionTest() {
```
mernst, line 1927: This is a nice test.  Thanks.

src/main/java/randoop/contract/CompareToEqualsWithImpurity.java
```diff
@@ -0,0 +1,42 @@
+package randoop.contract;
+
+/**
```
mernst, line 3: The comment makes it sound like Randoop should, everywhere, not perform the CompareToEquals contract on StringBuffer.  Therefore, I don't think this should be specefic to GRT Impurity fuzzing, but should be a direct change to CompareToEquals.  If you agree, could you please do that in a separate pull request?

mernst (changes requested):
Thanks, this is looking better.  I left some comments for further improvement.

Also, you don't need to write Javadoc like this:

```
   * @throws RandoopBug if an unexpected exception occurs
   * @throws NoSuchMethodException if the StringBuilder constructor cannot be found
   * @throws IllegalArgumentException if invalid sequence or String fuzzing operation is passed
```

These should all be impossible (they indicate a bug in Randoop), and the user can't do anything about them, so there is no point to documenting them, which bloats the documentation without providing any real value.

src/main/java/randoop/contract/CompareToEquals.java
776styjsu: CompareToEqualsWithImpurity (subclass of CompareToEquals) calls it to initialize itself

mernst: That makes sense.  Please add a comment to that effect.  However, I later suggested removing CompareToEqualsWithImpurity, in which case this could be made private again.

src/main/java/randoop/test/RegressionCaptureGenerator.java
776styjsu: It does have an effect. 
When fuzzing short, GRT Impurity creates:
```
int int12 = java.lang.Integer.sum((int) (short) 10, (int) (short) 43);
java.lang.Integer int13 = java.lang.Integer.valueOf(int12);
short short14 = int13.shortValue();
```
Where a java.lang.Integer is created as part of the fuzzing process.
This cannot be handled properly without the change because:
```
equalityMode = var.getType().isPrimitive() ? EQUALSEQUALS : EQUALSMETHOD;
```
and java.lang.Integer is not primitive with
```
var.getType().isPrimitive()
```
So I changed the code there to force the boxed type to use EQUALSEQUALS, which solves this problem due to autoboxing.

mernst: Ah, it is using `==` for boxed types.  Thanks for the clarification.

`==` should not be used for two boxed types.  Here is why:

`Integer.valueOf(10) == Integer.valueOf(10)`  →  true
`Integer.valueOf(1000) == Integer.valueOf(1000)`  →  false !

src/main/java/randoop/generation/GrtImpurity.java
776styjsu: Truncation (I should use a better word) is done automatically by Integer.shortValue() when fuzzing shorts. I should have mentioned this explicitly. I assumed that fuzzing a short outside its range would be significantly more common so I mentioned it, but I agree that I need to be consistent in my explanation.

What do you recommend? Should I mention that overflow/underflow can occur, which contradicts the Gaussian distribution for all primitive number types? Or do you think it's unnecessary to mention this at all?

mernst: It's not "truncation", but modulus or bit masking.
I don't think it is necessary to mention in a Javadoc comment.

src/main/java/randoop/test/RegressionCaptureGenerator.java
776styjsu: Thank you for pointing that out! It looks like Randoop always generates ```PrimValue``` assertions with a number on the right side of the ```==``` as a primitive type, regardless of whether a primitive or boxed type is being checked. Therefore, this shouldn't be a problem.

src/main/java/randoop/generation/GrtImpurity.java
776styjsu: Thank you for the suggestions!

src/main/java/randoop/generation/GrtImpurity.java
776styjsu: ```sequence.getStatement(sequence.size() - 1).getValue();``` returns Object. I'll explain this in the Javadoc.

776styjsu: It returns String now

src/main/java/randoop/util/EquivalenceChecker.java
776styjsu: Oops I forgot to remove it. It's not needed

src/main/java/randoop/contract/CompareToEqualsWithImpurity.java
776styjsu: CompareToEquals is only done on StringBuilder when GRT-Fuzzing is on. I'll make this change now since it doesn't hurt and I can remove CompareToEqualsWithImpurity. I'll try to understand this better in the future.

776styjsu:
@mernst Code updated, please review. Thanks!

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -148,8 +149,22 @@ public RegressionChecks generateTestChecks(ExecutableSequence eseq) {
               continue;
             }
             // System.out.printf("Adding objectcheck %s to seq %08X%n", poc, s.seq_id());
-            PrimValue.EqualityMode equalityMode =
-                var.getType().isPrimitive() ? EQUALSEQUALS : EQUALSMETHOD;
+            PrimValue.EqualityMode equalityMode;
+
+            // GRT Fuzzing option's short fuzzing strategy results in ambiguous assertions
+            // for primitive numbers. Use == for primitives since Java automatically
```
mernst, line 155: By "short", do you mean the `short` numeric type?  If so, please clarify, say, by putting `` ` `` around "short".

src/main/java/randoop/test/RegressionCaptureGenerator.java
```diff
@@ -148,8 +149,22 @@ public RegressionChecks generateTestChecks(ExecutableSequence eseq) {
...
+            // unboxes them when comparing with ==, which removes the ambiguity.
```
mernst, line 156: I agree with this comment for `short == Short` and `Short == short`, but I don't think it is true for `Short == Short`.  Is that fact irrelevant?  Please write a brief comment to discuss.

mernst (changes requested):
This is looking good!
I have a few comments.

src/main/java/randoop/test/RegressionCaptureGenerator.java
776styjsu: I think this fact is irrelevant here. When Randoop generates code for this type of assertion in the output unit test, it uses Value.toCodeString(Object). If the object in question is a boxed primitive, Randoop's method converts it to its primitive form and writes it as such in the output tests. For example, if the object is a short, the method will cast it to a primitive short by prepending `(short)` to the value in the generated code, so the case where `Short == Short` can never occur (and the same for other types of primitives/boxed numbers).

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -614,6 +616,23 @@ private void randoopConsistencyTests(Sequence newSequence) {
     }
   }
 
+  /** The numeric types that "GRT Fuzzing" fuzzes. */
+  // TODO: Why are byte and char omitted?
```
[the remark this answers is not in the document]
776styjsu: I am still uncertain whether these should be omitted. 

With `byte`, my primary concern is that they are sometimes treated as raw data rather than numeric values, and in such cases we may not want to fuzz them. Another potential problem with fuzzing bytes is their limited range, which increases the likelihood of overflow or underflow. This can cause the fuzzed value (bit-masked) to deviate from the specification (a Gaussian distribution centered around the original un-fuzzed input value). However, we decided to consider this as normal behavior and not to document it, as discussed earlier.

For `char`, the paper did not discuss how to fuzz it. One approach could be to use the original input char and classify it into categories such as "digit," "letter," "control character," etc. Depending on the category of the original input char, we might fuzz it by replacing it with a randomly selected character from the same category. We could choose to ignore `char` and not fuzz it. This decision is supported by the existence of GRT Constant Mining component. Consider the following example:
image
I assume that the figure represents a very common way of using `char` in a Java program. GRT Constant Mining will extract these constants and populate the initial SequenceCollection so when selecting an input to test this method, we can cover each case. Here, we don't want to fuzz the `char` and cause it to no longer covering these cases.

What do you think about this?

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -824,7 +858,8 @@ private InputsAndSuccessFlag selectInputs(TypedOperation operation) {
         }
       }
 
-      variables.add(totStatements + randomVariable.index);
+      variables.add(
+          totStatements + randomVariable.index + grtFuzzingAndNumStatements.numStatements);
```
mernst, line 862: Is the `GrtFuzzingAndNumStatements` data structure needed?  It seems to me that the number of statements could be computed by subtracting the length of `chosenSeq` before and after the call to `GrtFuzzing.fuzz`.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+ * shorter length by reducing the number of redundant sequences that do not side-effect the state of
+ * an object. That is not implemented here.]
+ *
+ * <p>For details about fuzzing, see {@link #fuzz}.
+ */
+@SuppressWarnings("NotJavadoc") // perhaps https://github.com/google/error-prone/issues/3904
+public class GrtFuzzing {
+
+  /**
+   * An enum representing the fuzzing operations for Strings. Each operation has a corresponding
```
mernst, line 43: Is it always exactly one corresponding method?

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+        int randomIndex = Randomness.nextRandomInt(stringLength + 1);
+        char randomChar = (char) (Randomness.nextRandomInt(95) + 32); // ASCII 32-126
+        Sequence randomIndexSequence = Sequence.createSequenceForPrimitive(randomIndex);
+        Sequence randomCharSequence = Sequence.createSequenceForPrimitive(randomChar);
+        return Sequence.concatenate(randomIndexSequence, randomCharSequence);
+      }
+
+      @Override
+      List<Executable> getMethods() throws NoSuchMethodException {
+        List<Executable> methodList = new ArrayList<>();
```
mernst, line 60: Throughout this enum, using `Arrays.asList()` is shorter and more efficient than `new ArrayList`.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+        methodList.add(StringBuilder.class.getMethod("substring", int.class, int.class));
+        return methodList;
+      }
+    };
+
+    /** Get the inputs for the fuzzing operation as a sequence. */
```
mernst, line 132: I'm a bit unclear on the relationship between "fuzzing operation" and method calls in a sequence.  Could you clarify that?  Also, is there always exactly one fuzzing operation?

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+    abstract Sequence getInputs(int stringLength);
+
+    /** Get the methods for the fuzzing operation. */
+    abstract List<Executable> getMethods() throws NoSuchMethodException;
```
mernst, line 136: I think a better name than `getMethods()` would be `get StringBuilderTransform()`, because it takes as input a StringBuilder and produces as output a StringBuilder.
Please also clarify the assumptions and relationship with `getInputs()`.  I think that it is assumed that no method in `getMethods()` takes any non-receiver arguments, except for exactly one method, which takes exactly `getInputs()` arguments.  Is that right?

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+  /** Do not instantiate. */
+  private GrtFuzzing() {
+    throw new Error("Do not instantiate");
+  }
+
+  /**
+   * Generate an extended fuzzed sequence for the given sequence.
+   *
+   * <p>This is the main entry point for GRT Fuzzing. This method appends additional fuzzing
+   * statements to the input sequence to fuzz the input object. This will generate more object
```
mernst, line 151: What is "the input object"?  I think it is the last value in the sequence.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+   * states and potentially improve coverage for the unit test generated by Randoop by triggering
```
mernst, line 152: You can drop this sentence; it was said in the class Javadoc.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+
+    // TODO: Implement cast to improve short fuzzing output readability
+    //  (e.g. cast to Integer for Integer.sum when fuzzing short can make Integer.valueOf()
+    //  unnecessary).
+
+    // Append fuzzing operation statements to the sequence.
+    for (Executable executable : fuzzingOperations) {
+      output =
+          appendFuzzingOperation(output, executable, getOutputType(executable), fuzzStatementCount);
+    }
```
mernst, line 228: Would it be possible to concatenate just two sequences with one another?  The `for` loop does a lot of intermediate concatenation, which is less efficient.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+
+    return new GrtFuzzingAndNumStatements(output, fuzzStatementCount.get());
+  }
+
+  /**
+   * Create a new sequence with a fuzzing operation statement appended to the given sequence.
```
mernst, line 234: This says "a fuzzing operation", but two lines below is "the fuzzing operations".  Also relate "fuzzing operation" to "fuzzing statement".

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+   *
+   * @param sequence the sequence to append the fuzzing operations to
+   * @param fuzzingOperation the method to be invoked to fuzz the object
+   * @param outputType the output type of the fuzzing operation
+   * @param fuzzStatementCount the number of fuzzing statements added, will be side-effected
+   *     (incremented).
```
mernst, line 240: Nit: In a Javadoc `@param`, `@return`, etc. clause, the initial text is a sentence fragment that does not end with a period unless followed by another sentence.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+      inputTypeList.add(declaringType);
+    }
+    for (Class<?> cls : executable.getParameterTypes()) {
+      inputTypeList.add(Type.forClass(cls));
+    }
+    return inputTypeList;
+  }
+
+  /**
+   * Calculate the indices of the input statements for the fuzzing operation. This method is called
```
mernst, line 308: The documentation speaks of "the fuzzing operation" and "the sequence", but it's not clear what these refer to, from reading the Javadoc in isolation.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+    } else {
+      throw new RuntimeException("Unexpected primitive type: " + cls.getName());
+    }
+  }
+
+  /**
+   * Get the methods for fuzzing primitive numbers of the given class.
+   *
+   * @param cls a numeric class
+   * @return a list of methods that will, together, be used to fuzz a number of the given class
```
mernst, line 367: Might other method calls (or at least other statements) be needed, for example to create values passed to the method calls returned by this method?  Please discuss.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+      methodList.add(Integer.class.getMethod("sum", int.class, int.class));
+    } else if (cls == double.class || cls == Double.class) {
+      methodList.add(Double.class.getMethod("sum", double.class, double.class));
+    } else if (cls == float.class || cls == Float.class) {
+      methodList.add(Float.class.getMethod("sum", float.class, float.class));
+    } else if (cls == long.class || cls == Long.class) {
+      methodList.add(Long.class.getMethod("sum", long.class, long.class));
+    } else if (cls == short.class || cls == Short.class) {
+      // Short doesn't have a sum method, so we use Integer.sum and get the short value
+      methodList.add(Integer.class.getMethod("sum", int.class, int.class));
```
mernst, line 383: Could the client just use `+` and a cast, rather than method calls?

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+   *
+   * @param cls a numeric class
+   * @return a list of methods that will, together, be used to fuzz a number of the given class
+   */
+  private static List<Executable> getNumberFuzzingMethods(Class<?> cls)
+      throws NoSuchMethodException {
+    List<Executable> methodList = new ArrayList<>();
+
+    if (cls == int.class || cls == Integer.class) {
+      methodList.add(Integer.class.getMethod("sum", int.class, int.class));
```
mernst, line 374: It would be more efficient, and clearer (logic would be localized), to `return Collections.singletonList(...);` here.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    return methodList;
+  }
+
+  /**
+   * Create and append all statements needed for the String fuzzing operation to the given sequence.
+   * Precondition: Length of the input String to be fuzzed is not 0.
+   *
+   * @param sequence the (non-empty) sequence to append the String fuzzing operation inputs to
```
mernst, line 397: If this must create a String with its last statement, please state that requirement here.

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+
+    if (stringLength == 0 && operation != StringFuzzingOperation.INSERT) {
+      return sequence; // Cannot remove/replace/substring an empty string
+    }
+
+    sequence = appendStringBuilder(sequence, fuzzStatementCount);
+
+    Sequence fuzzingInputsSequence = getStringFuzzingInputs(operation, stringLength);
+
+    return Sequence.concatenate(sequence, fuzzingInputsSequence);
```
mernst, line 416: This code is a bit inconsistent in that some statements are appended to `sequence` and some are put in `fuzzingInputsSequence`.  Why not put all the extra statements together, and then concatenate just once?

src/main/java/randoop/generation/GrtFuzzing.java
```diff
@@ -0,0 +1,480 @@
...
+    return operation.getInputs(stringLength);
+  }
+
+  /**
+   * Get a list of methods for fuzzing the input String based on the given operation.
+   *
+   * @param operation the string fuzzing operation to perform
+   * @return a list of methods that will be used to fuzz the input String
+   */
+  public static List<Executable> getStringFuzzingMethod(StringFuzzingOperation operation)
```
mernst, line 476: This method is not needed.  Please inline it at any calls.

mernst (changes requested):
Here are some comments, mainly on `GrtFuzzing.java`.

src/main/java/randoop/generation/GrtFuzzing.java
776styjsu: Added as TODO

src/main/java/randoop/generation/GrtFuzzing.java
776styjsu: I don't think we should try to concatenate two sequences right now, but we might in the future. If we create a sequence for concatenation from a list of statements, the statements must have non-negative indexes. However, most fuzzing statements have inputs, so the first statement must have a negative index, which invalidates the sequence. The only other way to create a sequence is to repeatedly call createSequence like the current implementation, which is unfortunately not efficient but works.

776styjsu:
@mernst I updated the code, please review it. Thanks!

src/main/java/randoop/generation/GrtFuzzerRegistry.java
```diff
@@ -0,0 +1,29 @@
+package randoop.generation;
```
mernst, line 1: Please move this file's contents into `GrtFuzzer`.

src/main/java/randoop/generation/GrtBaseFuzzer.java
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Append fuzzing statements to {@code sequence}. If the type is unsupported by this fuzzer, the
+   * implementation should simply return the original sequence.
+   *
+   * @param sequence the (non-null, non-empty) sequence whose <em>last</em> value will be fuzzed
+   * @return a new sequence with additional fuzzing statements, or the original sequence if no
+   *     fuzzing was performed
+   * @throws RandoopBug if fuzzing fails for an unexpected reason
+   */
+  public abstract Sequence fuzz(Sequence sequence);
```
mernst, line 71: Is there an assumption that `sequence` has already been executed?

src/main/java/randoop/generation/GrtNumericFuzzer.java
```diff
@@ -0,0 +1,129 @@
...
+        && runtimeClass != boolean.class
+        && runtimeClass != Boolean.class
+        && (runtimeClass.isPrimitive() || PrimitiveTypes.isBoxedPrimitive(runtimeClass));
+  }
+
+  @Override
+  public Sequence fuzz(Sequence sequence) {
+    Type inputType = sequence.getLastVariable().getType();
+    try {
+      // 1. Re-emit last value as a literal to avoid repeating impure stmt
```
mernst, line 88: I don't understand why it needs to be re-emitted.  Isn't the value in the last variable of the sequence, and can't that variable be used?

src/main/java/randoop/generation/GrtStringFuzzer.java
```diff
@@ -0,0 +1,259 @@
...
+          Sequence.createSequence(
+              SB_CTOR_OP, Collections.singletonList(strLitSeq), Collections.singletonList(0));
+      stmts.addAll(sbSeq.statements.toJDKList());
+
+      // 2) Extra inputs for the chosen fuzz operation (indices / chars)
+      Sequence opInputs = strFuzzingOp.inputs(strToFuzz.length());
+      stmts.addAll(opInputs.statements.toJDKList());
+
+      // 3) Method calls that perform the fuzzing operation
+      stmts.addAll(convertExecutables(strFuzzingOp.methods()));
```
mernst, line 107: This seems a bit heavy-handed.  I think the work can be done in a single constant assignment (if you want to do the fuzzing at test generation time rather than at test execution time) or as a string concatenation (which would require a new class such as the PlusOperation class).

src/main/java/randoop/generation/GrtStringFuzzer.java
```diff
@@ -0,0 +1,259 @@
...
+     * @return a sequence of inputs for this operation
+     */
+    abstract Sequence inputs(int length);
+
+    /**
+     * Initialize the list of methods for a given operation.
+     *
+     * @param name the name of the method
+     * @param params the parameter types of the method
+     * @return a list of {@link Executable} objects representing the method and its toString method
```
mernst, line 243: I think the `toString()` for `StringBuilder` is not necessary if the generated tests don't use StringBuilder.  Then this can return an Executable rather than a list.  (More generally,  I think several things will be simplified.)

src/main/java/randoop/generation/GrtStringFuzzer.java
```diff
@@ -0,0 +1,259 @@
...
+     */
+    private static List<Executable> init(String name, Class<?>... params) {
```
mernst, line 245: `init` is not a descriptive name for what this method returns.

src/main/java/randoop/generation/GrtStringFuzzer.java
```diff
@@ -0,0 +1,259 @@
...
+      EnumMap<StringFuzzingOperation, List<Executable>> m =
+          new EnumMap<>(StringFuzzingOperation.class);
+      m.put(INSERT, init("insert", int.class, char.class));
+      m.put(REMOVE, init("deleteCharAt", int.class));
+      m.put(REPLACE, init("replace", int.class, int.class, String.class));
+      m.put(SUBSTRING, init("substring", int.class, int.class));
+      METHOD_CACHE = Collections.unmodifiableMap(m);
+    }
+
+    /** Return the list of methods for this operation. */
```
mernst, line 224: More specifically, is it a list of methods such that running them all in sequence performs the given operation?  Please clarify.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -810,6 +810,21 @@ private InputsAndSuccessFlag selectInputs(TypedOperation operation) {
       Variable randomVariable = varAndSeq.var;
       Sequence chosenSeq = varAndSeq.seq;
 
+      // Fuzz the inputs for method calls and constructors to increase tests diversity.
+      // See randoop.generation.GrtFuzzing for details.
+      int chosenSeqSize = chosenSeq.size();
+      int chosenSeqSizeAfterFuzzing = chosenSeqSize;
```
mernst, line 816: I hope that the fuzzing can be made to be always exactly one statement, by not using StringBuffers.

mernst (changes requested):
@RockingMat Here are some simplification suggestions.  Some of the questions may require input from @776styjsu.

RockingMat:
@mernst, @776styjsu  I have made some changes based on your comments and simplified the string and numeric fuzzer. Both fuzzers now do fuzzing during generation time. Let me know what other changes are necessary.

src/main/java/randoop/operation/PlusOperation.java
```diff
@@ -0,0 +1,168 @@
+package randoop.operation;
```
mernst, line 1: This class is never used.  Is it needed?  (Or maybe we want to retain it, unused, just in case it is needed in the future, rather than deleting it.)

mernst (changes requested):

src/main/java/randoop/operation/PlusOperation.java
RockingMat: I think we can delete it. There is another pull request that is open for the plus operation I believe.

mernst:
@RockingMat There is missing documentation.

PR #1340, opened 2024-06-29
CompareToEquals ignores StringBuilder check

776styjsu:
StringBuilder implements the Comparable interface. However, it does not override the equals method. This causes a false contract failure when this check is performed on StringBuilder. The changes are that the contract will always evaluate to true if it involves a StringBuilder.


mernst (approved):
Thanks!

Issue #1344, opened 2024-07-03
Missing null flag usage


AgusNolasco:
https://github.com/randoop/randoop/blob/4e08395bdfc78e51e2cfd4ebc005e05c58b7a581/src/main/java/randoop/generation/ForwardGenerator.java#L719

PR #1345, opened 2024-07-03
Add forbid null flag (#1344)

AgusNolasco:
Using randoop, i could observe that setting forbid-null=true does not forbid null actually. Adding this change in the if statement condition does it.

### DIFF
--- a/src/main/java/randoop/generation/ForwardGenerator.java
+++ b/src/main/java/randoop/generation/ForwardGenerator.java
@@ -717,6 +717,7 @@ public class ForwardGenerator extends AbstractGenerator {
       // The user may have requested that we use null values as inputs with some given frequency.
       // If this is the case, then use null instead with some probability.
       if (!isReceiver
+          && !GenInputsAbstract.forbid_null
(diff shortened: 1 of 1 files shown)


mernst (approved):
Thank you for noticing and fixing this bug.  I appreciate it.

There is another change that should be made along with this one (in `GenInputsAbstract.checkOptionsValid()`), but I can't edit this pull request.  In the future, could you check the "maintainers may edit" box when creating a pull request?  (Or, maybe the problem is that your fork of Randoop is private.)

In any event, thanks again for this patch!

AgusNolasco:
> Thank you for noticing and fixing this bug. I appreciate it.
> 
> There is another change that should be made along with this one (in `GenInputsAbstract.checkOptionsValid()`), but I can't edit this pull request. In the future, could you check the "maintainers may edit" box when creating a pull request? (Or, maybe the problem is that your fork of Randoop is private.)
> 
> In any event, thanks again for this patch!

Thanks you and sorry, i did know about the option for allowing edits by maintainers. I will check it the next time.

PR #1347, opened 2024-07-04
GRT Elephant Brain

776styjsu:
Manage method sequences (to create inputs) in the SequenceCollection with the exact types obtained at run-time. This feature diversifies the input object types by using dynamic type information to find objects that cannot be generated using static type information alone.


776styjsu:
Current implementation has some problems at the moment. Will be fixed soon.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +281,38 @@ public Set<Sequence> getAllSequences() {
...
+      return;
+    }
+    ReferenceValue lastValue = lastValues.get(0);
+    Type expectedType = lastValue.getType();
+    Type actualType = Type.forClass(lastValue.getObjectValue().getClass());
+
+    if (actualType.isSubtypeOf(expectedType) && !actualType.equals(expectedType)) {
+      TypedOperation castOperation = TypedOperation.createCast(expectedType, actualType);
+
+      // Get the last variable in the sequence that has the expected type, cast it to the actual
```
mernst, line 305: Don't we know this is the first variable in the last statement in the sequence?  Is there a more straightforward way to get it than making a list of all the variables of the given type (which seems like unnecessary work)?

mernst (changes requested):

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +280,36 @@ public Set<Sequence> getAllSequences() {
...
+   * subtype of its static type, cast it to its dynamic type. This allows Randoop to create input
+   * objects that cannot be created using static type information alone for the method under test.
+   *
+   * @param eSeq the executable sequence to apply dynamic type casting on
+   */
+  private void castToRunTimeType(ExecutableSequence eSeq) {
+    List<ReferenceValue> lastValues = eSeq.getLastStatementValues();
+    if (lastValues.isEmpty()) {
+      return;
+    }
```
RockingMat, line 296: You can add a check for if lastValues.getObjectValue() is null here.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +280,36 @@ public Set<Sequence> getAllSequences() {
...
+    ReferenceValue lastValue = lastValues.get(0);
```
RockingMat, line 297: It may be worth handling multiple return values here.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +280,36 @@ public Set<Sequence> getAllSequences() {
...
+    Type declaredType = lastValue.getType();
+    Type runTimeType = Type.forClass(lastValue.getObjectValue().getClass());
+
+    if (!runTimeType.equals(declaredType) && runTimeType.isSubtypeOf(declaredType)) {
+      TypedOperation castOperation = TypedOperation.createCast(declaredType, runTimeType);
```
RockingMat, line 302: The original paper only downcasts with a certain probability(50%)

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -235,6 +236,14 @@ private void initializeRuntimePrimitivesSeen() {
 
     eSeq.execute(executionVisitor, checkGenerator);
 
+    // Dynamic type casting permits calling methods that do not exist on the declared type.
+    if (GenInputsAbstract.cast_to_run_time_type && eSeq.isNormalExecution()) {
+      castToRunTimeType(eSeq);
+      // Re-execute the sequence after applying dynamic type casting.
```
mernst, line 242: Please only re-execute the sequence if necessary -- that is, if `castToRunTimeType` *needed* to insert a cast because it changed the variable type.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +279,38 @@ public Set<Sequence> getAllSequences() {
...
+   * Randoop to call methods on it that do not exist in the supertype.
+   *
+   * <p>This implements the "GRT Elephant-Brain" component, as described in "GRT:
+   * Program-Analysis-Guided Random Testing" by Ma et. al (ASE 2015):
+   * https://people.kth.se/~artho/papers/lei-ase2015.pdf.
+   *
+   * @param eSeq an executable sequence; may be side-effected
+   */
+  private void castToRunTimeType(ExecutableSequence eSeq) {
+    List<ReferenceValue> lastValues = eSeq.getLastStatementValues();
```
mernst, line 294: I don't think we want to consider all the values of the last statement, but only the return value of the last call.  A side effect to an existing value cannot change the value's run-time class.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +279,38 @@ public Set<Sequence> getAllSequences() {
...
+    if (lastValues.isEmpty()) {
+      return;
+    }
+    ReferenceValue lastValue = lastValues.get(0);
+    Type declaredType = lastValue.getType();
+    Type runTimeType = Type.forClass(lastValue.getObjectValue().getClass());
+
+    if (!runTimeType.equals(declaredType) && runTimeType.isSubtypeOf(declaredType)) {
```
mernst, line 302: Under what circumstance can `runTimeType.isSubtypeOf(declaredType)` be false?  If it should always be true, then please write it as an assertion rather than in the `if` condition.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +279,38 @@ public Set<Sequence> getAllSequences() {
...
+      TypedOperation castOperation = TypedOperation.createCast(declaredType, runTimeType);
+
+      // Get the first variable of the last statement and cast it to the run-time type.
+      Variable variable = eSeq.sequence.firstVariableForTypeLastStatement(declaredType, false);
```
mernst, line 306: As noted above, I think that Elephant Brain should only work on the return value of the method called.  So the logic in `firstVariableForTypeLastStatement` is not necessary.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -270,6 +279,38 @@ public Set<Sequence> getAllSequences() {
...
+
+      if (variable != null) {
```
mernst, line 308: How can `variable` be null?  If it is null, does that indicate a bug in Elephant Brain?  If so, please make it an assertion rather than an `if` condition.

src/main/java/randoop/sequence/Sequence.java
```diff
@@ -699,18 +699,64 @@ private Statement getStatementWithInputs(int index) {
...
+    for (Variable var : this.lastStatementVariables) {
+      if (matchesVariable(var, type, onlyReceivers)) {
+        possibleVars.add(var);
       }
     }
     return possibleVars;
   }
 
+  /**
+   * Return the first value of type {@code type} that is produced by, or might be side-effected by,
```
mernst, line 711: As noted elsewhere, I think the logic that is needed is to return the single value produced by the method call in the last statement (which should be to a non-void method).

mernst (changes requested):

src/main/java/randoop/generation/ForwardGenerator.java
RockingMat: Doesn't the code already do this by grabbing lastValues.get(0)?

RockingMat:
I have addressed the comments from the most recent code review @mernst. I don't understand how the matchesVariables method was created in this PR but is used in methods that weren't created in this PR such as allVariablesForTypeLastStatement.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -235,6 +236,17 @@ private void initializeRuntimePrimitivesSeen() {
 
     eSeq.execute(executionVisitor, checkGenerator);
 
+    // Dynamic type casting permits calling methods that do not exist on the declared type.
+    if (GenInputsAbstract.cast_to_run_time_type && eSeq.isNormalExecution()) {
+      Sequence oldSeq = eSeq.sequence;
+      castToRunTimeType(eSeq);
```
mernst, line 242: Since this method may change its argument, I think it would be more appropriate as an instance method rather than a method defined outside `ExecutableSequence`.

Whether or not you do that, can some or all of the body of the `if` be folded into the routine, to make the call site cleaner and simpler?

src/main/java/randoop/sequence/Sequence.java
```diff
@@ -699,18 +699,30 @@ private Statement getStatementWithInputs(int index) {
...
-        possibleVars.add(i);
+    for (Variable var : this.lastStatementVariables) {
+      if (matchesVariable(var, type, onlyReceivers)) {
+        possibleVars.add(var);
       }
     }
     return possibleVars;
   }
 
+  /**
```
mernst, line 710: This is a reasonable change, but could you make a separate PR for it?  (You don't have to remove it from this PR; that will happen automatically when the other one is merged and master is merged into this one.)

mernst (changes requested):
This looks great.   I have just two comments.

mernst:
I fixed several crashes; now the problem is coverage, which might indicate a deeper problem that @776styjsu needs to investigate.

RockingMat:
Thank you @mernst for simplifying the pull request and adding the nullness annotation. This pull request is no ready to be rereviewed.

src/main/java/randoop/sequence/Sequence.java
```diff
@@ -715,6 +715,40 @@ public List<Variable> allVariablesForTypeLastStatement(Type type, boolean onlyRe
...
+   *
+   * <pre>{@code
+   * // Sequence of statements:
+   * Integer num = 5;
+   * String text = num.toString();
+   *
+   * // Retrieve the first Integer variable from the last statement
+   * Variable result = sequence.firstVariableForTypeLastStatement(Integer.class, false);
+   *
+   * // 'result' refers to 'num'
```
mernst, line 731: The comment says "Returns the first value of type {@code type} that is produced by, or might be side-effected by, the last statement."  However, the Integer `num` is neither produced by nor side-effected by `String text = num.toString();`.  So, either the comment (probably) or the example is wrong.  I think this method may return the first variable of the given type in the signature.

RockingMat:
I have fixed the comment. Thank you for the quick review @mernst!

PR #1351, opened 2024-07-11
Create specifications from `@NonNull` and `@Pure` annotations

RockingMat:
Included nonull annotations in building specifications and pure annotated methods were included as side-effect-free


src/main/java/randoop/operation/TypedOperation.java
```diff
@@ -69,6 +84,111 @@ public abstract class TypedOperation implements Operation, Comparable<TypedOpera
...
+    List<Precondition> preconditions = new ArrayList<>();
+    List<Postcondition> postconditions = new ArrayList<>();
+    List<ThrowsCondition> throwsConditions = new ArrayList<>();
+
+    // Read parameter annotations
+    Parameter[] parameters = method.getParameters();
+    for (Parameter parameter : parameters) {
+      AnnotatedType annotatedType = parameter.getAnnotatedType();
+      for (Annotation annotation : annotatedType.getAnnotations()) {
+        if (annotation.annotationType().getSimpleName().equals("NonNull")) {
```
mernst, line 135: Please also check for "NotNull".

mernst (changes requested):

src/main/java/randoop/operation/TypedOperation.java
RockingMat: @mernst I changed the HashSets to LinkedHashSets. Let me know if there are any other changes that would be useful.

src/main/java/randoop/operation/TypedOperation.java
```diff
@@ -70,6 +84,83 @@ public abstract class TypedOperation implements Operation, Comparable<TypedOpera
...
+      Method method, @ClassGetName String className) {
+    String methodName = method.getName();
+    List<String> parameterNames =
+        CollectionsPlume.mapList(Parameter::getName, method.getParameters());
+    List<@ClassGetName String> parameterTypes =
+        CollectionsPlume.mapList(Class::getName, method.getParameterTypes());
+
+    OperationSignature operation =
+        OperationSignature.forConstructorName(className, methodName, parameterTypes);
+    Identifiers identifiers = new Identifiers("receiver", parameterNames, "result");
```
mernst, line 126: Is a `result` identifier always needed?  What if the method has `void` return type?

mernst (changes requested):
I have one concern; maybe adding a test regarding it would  be a good idea.

src/main/java/randoop/operation/TypedOperation.java
RockingMat: The Identifiers class should assign a "result" identifier in it's constructor regardless of the return type. If the return type is void I do not think the method would have a pure annotation.

Issue #1261, opened 2023-11-21
Randoop generates uncompilable sequences because "reference to assertEquals is ambiguous"


varuniy:
`Randoop Version: 4.3.2`
`Output of java -version:` 
> openjdk version "17.0.5" 2022-10-18
> OpenJDK Runtime Environment Temurin-17.0.5+8 (build 17.0.5+8)
> OpenJDK 64-Bit Server VM Temurin-17.0.5+8 (build 17.0.5+8, mixed mode, sharing)

`Operating System Version:`
> Edition:	  Windows 10 Home
> Version:   22H2
> Installed on:  5/‎3/‎2021
>OS build:  19045.3693
> Serial number:  004802701057
> Experience:  Windows Feature Experience Pack 1000.19053.1000.0

**Steps I took:**

1. Create a directory called `test` anywhere
2. Inside this directory, create three files called `MyClass1.java` and `MyClass2.java` and `myclasses.txt`
3. Paste the following contents into `MyClass1.java`, `MyClass2.java`, and `myclasses.txt` respectively:
```
import java.util.*;

public class MyClass1 {

    public Object input;

    public MyClass1 (Object input) {
        this.input = input;
    }

    public String privateMethod1 (Object input) {
        return input.toString();
    }
} 
```
```
import java.util.*;

public class MyClass2 {

    public Integer input;

    public MyClass2 (Integer input) {
        this.input = input;
    }

    public Integer privateMethod2(Integer input) {
        return input * 2;
    }
}
```
```
MyClass1
MyClass2
```
4. Compile `MyClass1.java` and `MyClass2.java` using `javac MyClass1` and `javac MyClass2`
5. Change into the directory containing the `test` subdirectory and run the following command: 
`java -classpath "test/;[path to randoop-all-4.3.2.jar]"  randoop.main.Main gentests --classlist="test/myclasses.txt" --time-limit=4`

**Note:** If trying to replicate this error on Linux, you may need to edit some of the above commands. Notably, please change the ";" to ":"

This was the output I generated:

```
Randoop for Java version 4.3.2.
Will try to generate tests for 2 classes.
PUBLIC MEMBERS=10
Explorer = ForwardGenerator(steps: 0, null steps: 0, num_sequences_generated: 0;                                             
    allSequences: 0, regresson seqs: 0, error seqs: 0=0=0, invalid seqs: 0, subsumed_sequences: 0, num_failed_output_test: 0;
[...]

PR #1354, opened 2024-07-13
Fix "reference to assertEquals is ambiguous" issue

776styjsu:
Fixes https://github.com/randoop/randoop/issues/1261


mernst (approved):
Thanks!

PR #1359, opened 2024-07-19
Fix Randoop failure when output assertion checks involve a char with the value '\''

776styjsu:
Randoop would generate an assertion like:
`org.junit.Assert.assertTrue("'" + char2 + "' != '" + ''' + "'", char2 == ''');`
if the method output contains the character `'`. 
`'''` is problematic.


mernst (approved):
Thanks!

PR #1367, opened 2024-08-17
Add option to include `assertArrayEquals(boolean[], boolean[])` definition in generated test suite

776styjsu:
This feature is for user with JUnit 4.11 or below, where JUnit does not have a built-in `assertArrayEquals(boolean[], boolean[])` defined.


mernst:
The `misc` job is failing because of missing Javadoc.

src/main/java/randoop/contract/ObserverEqArray.java
```diff
@@ -134,6 +135,9 @@ public String toCodeString() {
     } else if (value.getClass().getComponentType() == float.class) {
```
mernst, line 135: I suggest to, at the beginning of this class, set a variable to either "org.junit.Assert.assertArrayEquals" or "assertArrayEquals".  Then use it throughout.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -849,6 +849,16 @@ public enum InputSelectionMode {
...
 
+  /**
+   * If true, Randoop will add a definition of {@code assertArrayEquals(boolean[], boolean[])} to
+   * the generated JUnit tests and use it instead for checking the equality of boolean arrays. This
+   * option is a backward compatibility feature as {@code assertArrayEquals(boolean[], boolean[])}
+   * is not available in JUnit versions 4.11 and earlier.
+   */
+  @Option(
+      "If true, include an assertArrayEquals(boolean[], boolean[]) implementation in generated JUnit tests")
+  public static boolean legacy_boolean_array_check = false;
```
mernst, line 860: I would give this a name like "junit_pre_4_12", since that explains the reason.
More importantly, throughout please handle all arrays, not just boolean ones.
If we encountered the problem with boolean arrays, then it is likely to arise for other types of arrays as well, and it will be easier for us to handle them all now than to discover and fix the problem later.

mernst (changes requested):
This is a nice pull request -- small and clean.
However, please generalize it to other array assertions introduced in JUnit 4.12.

src/main/java/randoop/main/GenInputsAbstract.java
776styjsu: It’s odd that boolean[] was the only type that was missing in assertArrayEquals, while all other primitive array types were supported prior to version 4.12. According to the release notes for version 4.12:

> Assert.assertArrayEquals() previously existed for all primitive array types, except boolean[]. This has now been added for boolean[]

Given this, should we consider adding support for other primitive arrays just in case?

mernst: Thank you for clarifying.  I had assumed that all the array operations were added at the same time, uniformly.  Since only the boolean one is needed, your pull request is fine.

src/main/java/randoop/main/GenInputsAbstract.java
```diff
@@ -849,6 +849,10 @@ public enum InputSelectionMode {
   @Option("If true, use JUnit's reflective invocation; if false, use direct method calls")
   public static boolean junit_reflection_allowed = true;
 
+  /** Whether Randoop's generated tests will be run under JUnit 4.11 or earlier. */
+  @Option("True if generated JUnit tests will be run under JUnit 4.11 or earlier.")
+  public static boolean junit_pre_4_12 = false;
```
mernst, line 854: I suggest removing this command-line option.  I see no harm in always including the definition, which should work for all versions of JUnit.

mernst (changes requested):

mernst (approved):
Thanks!

PR #1370, opened 2024-08-28
add jacocoagent.jar to release

markro49:


mernst:
@markro49 I think we discussed that the code around `extractJacocoAgent` should be further simplified.  Should my assignment to this pull request be removed until that time?

markro49:
I could not remember the simplification we discussed.  I did simplify the release procedure by including jacocoagent.jar in build/libs.

PR #1374, opened 2024-09-01
Add an explicit cast to float for delta in assertArrayEquals(float[], float[], float)

776styjsu:
The code generated for `assertArrayEquals(float[], float[], float)` would cause uncompilale sequence. Prior to this change, the third argument was treated as a `double`, which did not match any of the `assertArrayEquals` method overloads.


mernst (approved):
Thanks!

PR #1381, opened 2024-09-11
Check if superclass is enum in isEnum method. 

RockingMat:
This caused an issue when using demand driven on the easymock library.

### DIFF
--- a/src/main/java/randoop/types/NonParameterizedType.java
+++ b/src/main/java/randoop/types/NonParameterizedType.java
@@ -175,7 +175,7 @@ public class NonParameterizedType extends ClassOrInterfaceType {
 
   @Override
   public boolean isEnum() {
-    return runtimeType.isEnum();
+    return runtimeType.isEnum() || (getRuntimeClass().getSuperclass() != null && getRuntimeClass().getSuperclass().isEnum());
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/main/java/randoop/types/NonParameterizedType.java
```diff
@@ -175,7 +175,9 @@ public boolean isBoxedPrimitive() {
 
   @Override
   public boolean isEnum() {
-    return runtimeType.isEnum();
+    return runtimeType.isEnum()
+        || (getRuntimeClass().getSuperclass() != null
+            && getRuntimeClass().getSuperclass().isEnum());
```
mernst, line 180: I was surprised by this code, since users cannot subclass an enum class.  However, `runtimeType` might be an enum _constant_ (as opposed to a class declared as `enum`).  I think this is worth a brief comment given my initial surprise.

For more details, https://docs.oracle.com/javase/specs/jls/se22/html/jls-8.html#jls-8.9, says "The permitted direct subclasses ([§8.1.6](https://docs.oracle.com/javase/specs/jls/se22/html/jls-8.html#jls-8.1.6)) of E are the anonymous classes implicitly declared by the enum constants that have a class body."

mernst (changes requested):

PR #1385, opened 2024-09-14
implementing all logs

RockingMat:


RockingMat:
I am having an issue where the checks are failing because some of the options in GenInputAbstract which I did not create are lacking documentation. Should I add a javadoc comment for all of them or is there some way to get around this?

mernst:
@RockingMat It's a bit annoying that CI warns about nearby problems that you didn't cause, just because you edited a line very near them.  (But this is a necessary evil.)  You just have to fix the sins of the people who coded before you.  I did this.

PR #1386, opened 2024-09-15
Fix Value.toCodeString to Correctly Handle Enum Instances of Anonymous Subclasses

776styjsu:
The old condition fails for enum constants that are instances of anonymous inner classes when they override methods. In such cases, `value.getClass().isEnum()` and `valueType.isEnum()` return false, leading to the object being incorrectly treated as a numeric type, and fails Randoop.


mernst:
Would it be possible to add a test case?

mernst:
@776styjsu CI is failing.

776styjsu:
Can no longer reproduce the problem.

PR #1390, opened 2024-09-23
Add LifecycleMethodManager for Managing Start-Stop Method Pairs

776styjsu:
Introduces LifecycleMethodManager to ensure every start method (e.g., starting a network service) has a corresponding stop method invoked at the end of its test sequence, reducing potential resource leaks.


src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,275 @@
...
+    TypedClassOperation classOperation = (TypedClassOperation) operation;
+    if (!classOperation.isMethodCall()) {
+      return false;
+    }
+
+    Method method = ((MethodCall) classOperation.getOperation()).getMethod();
+
+    // Check if the method matches any start method in the method pairs
+    for (MethodPair pair : methodPairs) {
+      if (method.getName().equals(pair.getStartMethodName()) && isValidPairMethod(method)) {
```
mernst, line 81: You don't need to repeatedly call `isValidPairMethod`.  Lift that outside the loop.

src/main/java/randoop/sequence/Statement.java
```diff
@@ -23,22 +23,46 @@ public final class Statement {
...
-  // of Variables, but a list of RelativeNegativeIndex objects.
-  // See that class for an explanation.
+  /**
+   * The list of values used as input to the statement. NOTE that the inputs to a statement are not
+   * a list of Variables, but a list of RelativeNegativeIndex objects. See that class for an
+   * explanation.
+   */
   final List<RelativeNegativeIndex> inputs;
 
+  /** Indicates if this statement is a pair start method. */
```
mernst, line 33: "True if" gives the polarity of the variable, unlike "indicates if".

src/main/java/randoop/sequence/Statement.java
```diff
@@ -176,6 +200,16 @@ public Type getDeclaringClass() {
     return null;
   }
 
+  /** Getter for the pair start method flag. */
```
mernst, line 203: This comment doesn't give any information.  State what it returns:  true if this statement is a call to a lifecycle start method.

src/main/java/randoop/sequence/Statement.java
```diff
@@ -176,6 +200,16 @@ public Type getDeclaringClass() {
...
+  public boolean isPairStart() {
+    return isPairStart;
+  }
+
+  /** Getter for the pair stop flag. */
```
mernst, line 208: See comment immediately above.

mernst (changes requested):
Thanks for this improvement.  I have left a few comments.

src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * <ul>
+ *   <li>{@code start()} and {@code stop()}
+ *   <li>{@code open()} and {@code close()}
+ *   <li>{@code run()} and {@code end()} ...
+ * </ul>
+ */
+public class MethodPairManager {
+  /** A list tracking method pairs, where each pair consists of a start and stop method. */
+  private final List<MethodPair> methodPairs = new ArrayList<>();
```
mernst, line 39: Why is this a `List` rather than, say, a map?  A map would make lookups more efficient.

src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,284 @@
...
+
```
mernst, line 40: Please state the invariants of the following fields.  For example, are they lazily filled in?  When is one set of values equal to another set of values, and when is it a subset?

src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,284 @@
...
+  /** Maps types to their corresponding stop operations. */
+  private final Map<Type, TypedOperation> stopOperationsByType = new HashMap<>();
+
+  /** The set of stop operations, for quick lookup. */
+  private final Set<TypedOperation> stopOperations = new HashSet<>();
+
+  /** Constructs a new MethodPairManager with default method pairs. */
+  public MethodPairManager() {
+    // Default method pairs
+    methodPairs.add(new MethodPair("start", "stop"));
```
mernst, line 53: What if a class has only one of the two methods?  (In that case they must not be lifecycle methods that have to be called in pairs.)

src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Returns the pair method type for the given operation. We assume that the operation is either a
+   * start or stop method. Not both.
```
mernst, line 125: This fact should be documented in the class documentation, not only here in this method.

src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  /**
+   * Finds a matching stop/close method for the given start operation and method pair.
+   *
+   * @param startOperation the start operation
+   * @param pair the MethodPair containing the stop method name
+   * @return the corresponding stop operation, or null if not found
+   */
+  public TypedOperation findMatchingStopMethod(
```
mernst, line 178: Since this returns an operation, I think it should be named `findMatchingStopOperation()`.

src/main/java/randoop/generation/MethodPairManager.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  /**
+   * Caches the stop operation for the given start operation and method pair.
+   *
+   * @param startOperation the start operation to cache the stop operation for
+   * @param pair the MethodPair containing the stop method name
+   */
+  private void cacheStopOperation(TypedClassOperation startOperation, MethodPair pair) {
+    TypedOperation stopOperation = findMatchingStopMethod(startOperation, pair);
```
mernst, line 162: `cacheStopOperation()` calls `findMatchingStopMethod()`, but I would think the opposite calling convention would make more sense.  Is there an invariant for `findMatchingStopMethod()` that it should be called only when the value is not yet cached?  (That doesn't make sense to me given that `findMatchingStopMethod()` is public and clients have no way of knowing information about caching.)

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -544,7 +572,8 @@ private ExecutableSequence createNewUniqueSequence() {
    * @param times the number of times to repeat the {@link Operation}
    * @return a new {@code Sequence}
    */
-  private Sequence repeat(Sequence seq, TypedOperation operation, int times) {
+  private Sequence repeat(
+      Sequence seq, TypedOperation operation, int times, PairMethodType pairMethodType) {
```
mernst: The new formal parameter needs to be documented in the Javadoc.

mernst (changes requested):
I have a few questions about data structure invariants.

PR #1484, opened 2025-05-18
Obtain constants from LDC information

mernst:
@776styjsu This greatly simplifies the previous implementation.  Is it still correct?


mernst:
@Kuro1Fury Could you please review this pull request?

PR #1490, opened 2025-05-22
Fix GRT character fuzz

776styjsu:


776styjsu:
@mernst This PR fixes an issue where Impurity couldn't fuzz `char` values. Could you please review it? Thanks!

PR #1516, opened 2025-06-11
Object.getClass() produce precise runtime types under Elephant Brain

776styjsu:


src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -395,14 +401,41 @@ private static Object[] getRuntimeValuesForVars(
...
+
+  /**
+   * Checks if the last operation in the sequence is a call to {@code java.lang.Object#getClass()}.
+   *
+   * @param seq the sequence to check
+   * @return true iff the last statement in {@code seq} is a call to {@code
+   *     java.lang.Object#getClass()}.
+   */
+  private static boolean lastOpIsGetClass(Sequence seq) {
+    if (seq.statements.isEmpty()) {
```
mernst: If this happens, is it a bug?  If so, please throw an exception rather than returning.  If it's not a bug, please briefly comment when it can happen.

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -420,6 +453,24 @@ public boolean castToRunTimeType() {
     Type declaredType = lastValue.getType();
     Type runTimeType = Type.forClass(lastValue.getObjectValue().getClass());
 
+    // If the last operation is a call to Object.getClass(), then
```
mernst: Briefly state the purpose of this.  What is wrong with `Class<?>`?  Under what circumstances does it make a difference whether the result is `Class<ObjectRuntimeType>` rather than `Class<?>`?  Thanks.

(Is this because of a weakness in Randoop, or is it because the different type enables more tests to be output, or is it because the different type enables certain tests to compile that otherwise would not compile, or something else?)

776styjsu:
@mernst It’s ready for review. The CI is now complaining about a large number of missing comments throughout the codebase. I'm not sure why it's only flagging them now and not earlier.

mernst:
@776styjsu The CI failure is (at the very end of the output):
```
Here are the test failures: spotlessCheck
```
so it only requires reformatting.

src/test/java/randoop/types/JDKTypesTest.java
```diff
@@ -37,6 +37,7 @@ public void collectionsMapTest() throws IllegalAccessException {
       }
     }
 
+    System.out.println("JDKTypes collection types: " + collectionTypes);
```
mernst: Should this be removed before the branch is merged?

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -235,9 +238,23 @@ private void initializeRuntimePrimitivesSeen() {
...
     // Dynamic type casting permits calling methods that do not exist on the declared type.
-    boolean cast = eSeq.castToRunTimeType();
-    // Re-execute the sequence after applying dynamic type casting.
-    if (cast) {
+    if (GenInputsAbstract.cast_to_run_time_type && eSeq.isNormalExecution()) {
+      eSeq.castToRunTimeType();
+
+      // Special-case getClass(): when run-time casting is enabled and the last op is
+      // Object.getClass(),
+      // convert Class<?> to Class<ConcreteType> to avoid emitting an uninstantiated Class<T>.
```
mernst: Here, I think you mean the _type_ `Class<T>`, not an object of that type.  Right?

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -396,57 +412,103 @@ private static Object[] getRuntimeValuesForVars(
...
     }
 
+    TypedOperation castOp = TypedOperation.createCast(declaredType, runTimeType);
+    Variable var = sequence.firstVariableForTypeLastStatement(declaredType, false);
+    if (var == null) {
+      return;
+    }
+
+    sequence = sequence.extend(castOp, Collections.singletonList(var));
+    refineClassReturnTypeForGetClass();
```
mernst: I think the call to `refineClassReturnTypeForGetClass();` (or to a variant of it) would be better placed in `TypedOperation.createCast()` instead.

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -396,57 +412,103 @@ private static Object[] getRuntimeValuesForVars(
...
-            "Run-time type %s [%s] is not a subtype of declared type %s [%s]",
-            runTimeType, runTimeType.getClass(), declaredType, declaredType.getClass());
+            "Run-time type %s is not a subtype of declared type %s", runTimeType, declaredType);
 
     if (runTimeType.equals(declaredType)) {
-      return false;
+      return; // Nothing to do
     }
 
+    TypedOperation castOp = TypedOperation.createCast(declaredType, runTimeType);
```
mernst: This can be moved after the `if` statement just below, to avoid doing the work if it will not be used.

src/main/java/randoop/sequence/ExecutableSequence.java
776styjsu: I have a couple of concerns about moving `refineClassReturnTypeForGetClass()` into `TypedOperation.createCast()`.

First, there’s a separation of concerns issue. From my understanding, `createCast()` is simply a factory method where it builds an unchecked cast operation when both the source and target types are already known. Choosing what type to cast to falls outside its responsibility.

Second, deciding whether refinement is needed would require `TypedOperation.createCast()` to inspect the operation itself. For example, checking if the last operation was a call to `Object.getClass()` and inspecting runtime types. That forces the method to become non-static. 

For these reasons I think it's cleaner to keep `refineClassReturnTypeForGetClass()` in `ExecutableSequence`.

I’m also concerned about the recent change that moves `refineClassReturnTypeForGetClass()` to the end of `castToRunTimeType()`. The problem is that the cast is skipped when the runtime type of the object to be cast is an uninstantiated generic, and the return type of `Object.getClass()` at runtime is an uninstantiated generic. Placing the call at the end makes it unreachable due to early returns.


Please correct me if I’m misunderstanding anything, and I’d appreciate hearing your thoughts on this.

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -396,23 +412,22 @@ private static Object[] getRuntimeValuesForVars(
...
-   *
-   * @return true if the cast was performed, false otherwise (in which case no side effect occurs)
    */
-  public boolean castToRunTimeType() {
+  public void castToRunTimeType() {
     if (!GenInputsAbstract.cast_to_run_time_type || !this.isNormalExecution()) {
-      return false;
+      return;
     }
     List<ReferenceValue> lastValues = this.getLastStatementValues();
```
mernst, line 435: Abstract out the code from here until just before `assert runTimeType.isSubtypeOf(declaredType)`.  Put it in a method named something like `ExecutableSequence.getRunTimeType()`, which may return null.  Move the logic that is currently in `refineClassReturnTypeForGetClass()` into the new method.  Currently, `refineClassReturnTypeForGetClass()` has a side effect, but within the new method it won't have a side effect, it will just cause the new method to return a good value.

The code in routine `castToRunTimeType()` can depend on the value returned by the new method `ExecutableSequence.getRunTimeType()`, without any complex or confusing logic here, and without the need to call a method with a side effect after `castToRunTimeType()` runs.

Another advantage is that there will be only one place in the code that creates a cast, rather than a cast sometimes being created by `castToRunTimeType()` and sometimes being done by a fixup routine `refineClassReturnTypeForGetClass()` when `castToRunTimeType()` fails to create a needed cast.

src/main/java/randoop/sequence/ExecutableSequence.java
mernst: Thanks for making these points.  I had overlooked the fact that `castToRunTimeType()` sometimes returns early even when a cast is needed.  Therefore, `refineClassReturnTypeForGetClass()` would not have had any effect, as you pointed out.

I suggest that the logic in `refineClassReturnTypeForGetClass()` that determines the correct type should run earlier in this method, in order to ensure that the call to `createCast()` two lines above is correct and always runs, without a need for other code to run after `castToRunTimeType()` to correct places where it did not create the cast that it should have.

Please see the concrete suggestion above.  (It starts "Abstract out the code from here until just before ...".)

776styjsu:
@mernst It is ready for review, Thanks!

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -396,56 +412,106 @@ private static Object[] getRuntimeValuesForVars(
...
 
     ReferenceValue lastValue = lastValues.get(0);
     Type declaredType = lastValue.getType();
-    Type runTimeType = Type.forClass(lastValue.getObjectValue().getClass());
+    Type runTimeType = getRunTimeType(lastValues, declaredType);
+    if (runTimeType == null) {
+      return; // nothing to cast to
+    }
+
+    Variable var = sequence.firstVariableForTypeLastStatement(declaredType, false);
```
mernst: This gets the first variable for a given type.

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -396,56 +412,106 @@ private static Object[] getRuntimeValuesForVars(
...
+
+    Type runTimeType;
+
+    if (lastOpIsGetClass()) {
+      // Special-case getClass(): when run-time casting is enabled and the last op is
+      // `Object.getClass()`, convert `Class<?>` to `Class<ConcreteType>` to avoid emitting the
+      // uninstantiated type "Class<T>". By default, method `Type#forClass` (required to find the
+      // run-time type to cast to in cast-to-run-time-type) on wildcard generics carries over type
+      // variables, which can produce uncompilable "Class<T>" in generated tests.
+      ReferenceType elemType = lastValues.get(lastValues.size() - 1).getType();
```
mernst: This gets the last element of `lastValues`.

src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -396,56 +412,106 @@ private static Object[] getRuntimeValuesForVars(
...
+      if (elemType.isGeneric()) {
+        GenericClassType g = (GenericClassType) elemType;
+        elemType =
+            g.instantiate(Collections.nCopies(g.getTypeParameters().size(), JavaTypes.OBJECT_TYPE));
+      }
+      // cast to Class<ConcreteType>
+      runTimeType = JavaTypes.CLASS_TYPE.instantiate(Collections.singletonList(elemType));
+    } else {
+      // Ordinary case: use the dynamic class of the returned value.
+      runTimeType = Type.forClass(lastValues.get(0).getObjectValue().getClass());
```
mernst, line 491: This gets the first element of `lastValues`.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -234,9 +234,10 @@ private void initializeRuntimePrimitivesSeen() {
     eSeq.execute(executionVisitor, checkGenerator);
 
     // Dynamic type casting permits calling methods that do not exist on the declared type.
-    boolean cast = eSeq.castToRunTimeType();
-    // Re-execute the sequence after applying dynamic type casting.
-    if (cast) {
+    if (GenInputsAbstract.cast_to_run_time_type && eSeq.isNormalExecution()) {
+      eSeq.castToRunTimeType();
+
+      // Re-execute the sequence after applying dynamic type casting.
```
mernst, line 240: Why is it better to always re-execute?  It seems that if `castToRunTimeType()` made no change, that re-execution is not necessary.

776styjsu:
> I have a concern about discrepancies regarding which value of the last statement is used.

@mernst Thanks. Those two indices,
`ReferenceType elemType = lastValues.get(lastValues.size() - 1).getType();`
and
`runTimeType = Type.forClass(lastValues.get(0).getObjectValue().getClass());`
are different on purpose:

For `getClass()` special-case, we need the receiver object to build `Class<ConcreteType>`, and the receiver is always the last element of `lastValues` since `getClass()` always has a `lastValues` of `[output, receiver]`. In fact, `lastValues.size() - 1` always evaluates to `1` there.

For ordinary calls, we need the return value itself, which is always the first element.

I've added comments to the code to make these clear. The code is now ready for review with these changes, thanks!

PR #1583, opened 2025-06-28
enum handling

RockingMat:


mernst (changes requested):
Can you add a test that fails on master but passes on this branch, to show that this code has an effect on Randoop?

RockingMat:
@mernst from what I can tell it doesn't seem like any unique enums are being generated which impact test generation (equivalent enums might be generated with OperationExtractor). Rather, the constant frequency map is updated. This isn't currently used in Randoop but will be used by ConstantMining.

mernst (approved):

PR #1592, opened 2025-07-12
Add replacecall dependencies

mernst:


agent/replacecall/src/main/java/randoop/grt_mock/javax/mail/Service.java
```diff
@@ -0,0 +1,27 @@
+package randoop.grt_mock.javax.mail;
```
mernst: I think these can be in `randoop.mock` rather than a separate `randoop.grt_mock`.  Other users may find these useful.

agent/replacecall/src/main/resources/default-replacements.txt
```diff
@@ -1,7 +1,7 @@
 // Default replacement file
 
 // Apply default mock for System.exit that throws randoop.SystemExitCalledError.
-java.lang.System randoop.mock.java.lang.System
+// java.lang.System randoop.mock.java.lang.System  # Do not replace System.exit() for GRT
```
mernst: What is the reason that GRT doesn't want to replace System.exit()?  Because it is run on full systems rather than on unit tests?

We won't want to directly change `default-replacements.txt`, but we could create a new file or add a command-line switch to make it easy for users to specify whether `System.exit()` should be mocked.

ant.m
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
mernst: What is the purpose of this file?

agent/replacecall/src/main/resources/default-replacements.txt
776styjsu: The test harness (JUnitTestRunner) calls System.exit() to return its status, and mocking it would result in false failures.

mernst: Rather than answering the question in the pull request, where it will never be seen again, please add a comment to the source code, where it will answer the question for anyone who reads the code in the future.  (This is a general rule.)  Thanks!

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -167,6 +187,56 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
         }
       }
 
+      // Remove some replacements.
+      if (replacecall_exclude != null || replacecall_exclude_file != null) {
+        Set<MethodSignature> excludeSignatures = new LinkedHashSet<>();
```
mernst, line 192: I don't see that the order of this set matters.  Why is it a `LinkedHashSet` rather than just a `HashSet`?

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -167,6 +187,56 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+            }
+          } catch (IOException e) {
+            System.err.printf(
+                "Error reading replacement exclusion file %s:%n %s%n",
+                replacecall_exclude_file, e.getMessage());
+            System.exit(1);
+          }
+        }
+
+        // 1) Remove any method-level replacement entries that exactly match excluded signatures.
```
mernst, line 218: Using two loops rather than one is complicated, and it's inefficient.  I expect `excludeSignatures` to be smaller, so iterate over it rather than iterating over `replacementMap`.

Furthermore, no mutation of `excludeSigatures` is necessary.  Iterate over it once, and for each item test whether it exactly matches some element in `replacementMap`.  If so, do the work in "1)".  Otherwise, do the work in "2)".

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -260,6 +349,121 @@ private static Set<String> loadExclusions(@Owning Reader exclusionReader, String
...
+  private static @Nullable MethodSignature parseMethodSignature(String text) {
+    String s = text.trim();
+
+    int colonPos = s.indexOf(':');
+    if (colonPos > 0 && s.indexOf('(') > colonPos) {
+      String fqName = s.substring(0, colonPos);
+      int dotPos = fqName.lastIndexOf('.');
+      if (dotPos > 0) {
+        String owner = fqName.substring(0, dotPos);
+        String name = fqName.substring(dotPos + 1);
```
mernst, line 373: Why bother parsing `fqName` into `owner` and `name`, given that the only way they are used is being concatenated back together, exactly like `fqName`?

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -68,12 +70,30 @@ public class ReplaceCallAgent {
...
 
   /** Exclude transformation of classes in the listed packages. */
   @SuppressWarnings("WeakerAccess")
   @Option("file listing packages whose classes should not be transformed")
-  public static @Nullable Path dont_transform = null;
+  public static @MonotonicNonNull Path dont_transform = null;
+
+  /**
+   * Default replacement methods that should not be used. A comma-separated list of method
+   * signatures in colon/JVM-descriptor form, e.g. "java.lang.System.exit:(I)V". JVM method
```
mernst, line 82: Where is "colon/JVM-descriptor form" documented?  Please provide a URL or other reference.  (For example, method descriptors are defined in JVMS section 4.3.3, and Randoop fully-qualified signatures are defined at https://randoop.github.io/randoop/manual/#fully-qualified-signature.)  If you made it up, then I suggest using a standard format instead.

It is odd that one signature format is used for adding a method replacement (namely, Randoop fully-qualified signatures, defined at https://randoop.github.io/randoop/manual/#fully-qualified-signature), but a *different* signature format is used for excluding one.  I think that difference is likely to confuse users.  What is the benefit of using a different format?

Another benefit of using an existing format is the ability to just call `MethodSignature.of()`, which will eliminate more than half of this pull request.

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -68,12 +70,30 @@ public class ReplaceCallAgent {
...
+  public static @MonotonicNonNull String replacecall_exclude = null;
+
+  /**
+   * File listing method to exclude from default replacements. Each line is a method signature in
+   * colon/JVM-descriptor form, e.g. "java.lang.System.exit:(I)V". JVM method descriptors are
+   * defined in the Java Virtual Machine Specification, section 4.3.3.
+   */
+  @SuppressWarnings("WeakerAccess")
+  @Option("file listing methods not to replace, one per line")
+  public static @MonotonicNonNull Path replacecall_exclude_file = null;
```
mernst, line 96: There is already a `--dont-transform` command-line argument that excludes classes from replacement.  Why is a new command-line argument required?  How about permitting that file to contain both packages and methods?  That is consistent with how replacements are defined, and it avoids the need for a new command-line option.

PR #1595, opened 2025-07-21
Impurity Object Fuzzing

776styjsu:


src/main/java/randoop/main/GenTests.java
```diff
@@ -40,7 +40,6 @@
 import org.checkerframework.checker.regex.qual.Regex;
 import org.checkerframework.checker.signature.qual.ClassGetName;
 import org.checkerframework.checker.signature.qual.Identifier;
-import org.checkerframework.dataflow.qual.Pure;
```
mernst: Can the changes in this file be pull-requested and tested separately?

776styjsu: I can split this file into a separate PR, but it will have a few minor issues.

This change replaces ```annotation instanceof Pure || annotation instanceof SideEffectFree``` with a name-based check (```randoop.org.checkerframework…```). Until the corresponding renaming lands, that old ```instanceof``` check always returns false, so the fuzzer will treat every method as impure. Meanwhile the doc still says we skip ```@Pure```/```@SideEffectFree``` operations, which would be temporarily untrue.

If we separate it, we have to live with misleading fuzzer docs. I'm happy to split if you’re okay with that. One argument in favor of splitting is that this isolated change is smaller and more likely to be merged first, so the docs would become accurate again soon after.

mernst: Where is the corresponding renaming?  Can that be pull-requested together with this change?

776styjsu: Sorry for not being clear, I was referring to the changes in ```GenTests.java```, where
```annotation instanceof Pure || annotation instanceof SideEffectFree``` (original)
was replaced with
```annotationName.equals(PURE_ANNOTATION) || annotationName.equals(SIDE_EFFECT_FREE)``` (new).

And thanks for the question, it helped me clarify a few things. I think I might have assumed there was a convention to revert such changes when splitting PRs. But if it's okay to keep the change in both, then separating them shouldn’t be a problem.

Comment (mernst) [reply src/main/java/randoop/main/GenTests.java]: A common workflow is:
 * create a new PR that is a subset of this one
 * review and merge that PR
 * merge the mainline into this PR, which makes this PR smaller and easier to review

File: src/main/java/randoop/generation/GrtObjectFuzzer.java [hunk 2]
```diff
@@ -0,0 +1,382 @@
...
+import randoop.types.ClassOrInterfaceType;
+import randoop.types.Type;
+import randoop.types.TypeTuple;
+import randoop.util.Randomness;
+
+/**
+ * Fuzzer that applies a single side-effecting operation to a variable within a test sequence to
+ * explore the stateful behavior (impurity) of that object.
+ *
+ * <p>Specifically, this fuzzer:
```
mernst: What is the input?  Is it a sequence, or a sequence and an argument index, or something else?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+   * @param sideEffectOps a set of side-effecting operations to add to the fuzzer (not annotated
+   *     with Checker Framework's {@code @Pure} or {@code @SideEffectFree})
+   * @param cm the component manager to use for getting sequences for types (should not be null)
+   */
+  public void initializeIfNeeded(Set<TypedOperation> sideEffectOps, ComponentManager cm) {
+    if (initialized) {
+      return;
+    }
+    operationsByType.clear();
+    unionCache.clear();
```
mernst: Why is it necessary to call `clear()`?  Might these maps contain entries?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+   */
+  private final Map<Type, List<TypedOperation>> operationsByType = new HashMap<>();
+
+  /** Component manager to get sequences for types. */
+  private @MonotonicNonNull ComponentManager componentManager;
+
+  /** Whether this fuzzer has been initialized. */
+  private boolean initialized = false;
+
+  /** Cache of resolved unions so we don't recompute ancestor walks. */
```
mernst: What is a "union"?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+
+    // Keep track of the sequences to concatenate and the index of the necessary variable in each.
+    List<Sequence> sequencesToConcat = new ArrayList<>(formals.size());
+    List<Integer> varIndicesInEachSeq = new ArrayList<>(formals.size());
+    int targetParamPos = -1; // Initialize to an invalid position.
+
+    // Collect input sequences for each formal parameter.
+    for (int i = 0; i < formals.size(); i++) {
+      Type formalType = formals.get(i);
+      if (formalType.isAssignableFrom(typeToFuzz) && i == fuzzParamPos) {
```
mernst: 1. Put the cheaper test, `==`, first in the short-circuiting conjunction.
2. If the `==` evaluates to true, then is it possible that `isAssignableFrom()` returns false?  Or should that be an assertion within the body of the `if` whose predicate is `==`?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+          "Variable to fuzz is not part of the sequence to fuzz. "
+              + "Variable sequence: "
+              + variable.sequence
+              + ", sequence to fuzz: "
+              + sequence);
+    }
+  }
+
+  /**
+   * Returns a list of operations that can be applied to the given type.
```
mernst: Please define "can be applied".  Are these a set of operations such that the type is a subtype of at least one of the formal parameters?  Or must there be exactly one formal parameter?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+   *
+   * @param t the type to check for applicable operations
+   * @return a list of operations that can be applied to the type, or an empty list if no operations
+   *     are applicable
+   */
+  private List<TypedOperation> getApplicableOps(Type t) {
+    Type root = erase(t);
+    return unionCache.computeIfAbsent(
```
mernst: What is the relationship between "applicable" and "union"?  Should one or both of them be renamed?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+
+  /**
+   * Returns a list of operations that can be applied to the given type.
+   *
+   * @param t the type to check for applicable operations
+   * @return a list of operations that can be applied to the type, or an empty list if no operations
+   *     are applicable
+   */
+  private List<TypedOperation> getApplicableOps(Type t) {
+    Type root = erase(t);
```
mernst: What is the name `root` mnemonic for?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+   * returns the raw type; otherwise it returns the type unchanged.
+   *
+   * @param t the type to erase (non-null)
+   * @return the erased/canonical type
+   */
+  private Type erase(Type t) {
+    if (t == null) {
+      throw new IllegalArgumentException("type cannot be null");
+    }
+    Type raw = t.getRawtype(); // Randoop's API: raw type for generics
```
mernst: If the underlying Randoop API is named "raw", should this method also be named "raw" rather than "erased" or "canonical"?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+    return (raw != null) ? raw : t;
```
mernst: Under what circumstances does `getRawType()` return `null`?  That is, what is the need for this method?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+   * @return a list of erased types including {@code t} and all ancestors
+   */
+  private List<Type> ancestorsAndSelf(Type t) {
+    Type start = erase(t);
+
+    // Non-class/interface types (primitives, arrays) have no ancestors.
+    if (!(start instanceof ClassOrInterfaceType)) {
+      List<Type> singleton = new ArrayList<>(1);
+      singleton.add(start);
+      return singleton;
```
mernst: Can this be just `Collections.singletonList(start)`?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    java.util.Deque<Type> queue = new java.util.ArrayDeque<>();
```
mernst: Given that this is used as a queue, a better declaration is
```suggestion
    Queue<Type> queue = new java.util.ArrayDeque<>();
```
Don't forget to import `java.util.Queue`.

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+    java.util.LinkedHashSet<Type> visited = new java.util.LinkedHashSet<>();
+
+    queue.add(start);
+    while (!queue.isEmpty()) {
+      Type next = erase(queue.removeFirst());
+      if (!visited.add(next)) {
+        continue; // already processed
+      }
+      if (next instanceof ClassOrInterfaceType) {
```
mernst: How can this be false?  If it cannot be, then remove the `if`.  The cast on the next line will tell us if we're wrong and a non-`ClassOrInterfaceType` flows to this location.

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,382 @@
...
+        ClassOrInterfaceType ci = (ClassOrInterfaceType) next;
+        ClassOrInterfaceType sup = ci.getSuperclass();
+        if (sup != null) {
+          queue.addLast(erase(sup));
+        }
+        for (Type itf : ci.getInterfaces()) {
+          if (itf != null) {
```
mernst: How can it happen that `itf` is null?  If it cannot be, then remove this check.

src/main/java/randoop/generation/AbstractGenerator.java
```diff
@@ -91,6 +91,13 @@ private long elapsedTime() {
    */
   protected final List<TypedOperation> operations;
 
+  /**
+   * A copy of {@link AbstractGenerator#operations} that contains all operations used to generate
```
mernst, line 95: Is this really "all operations"?  Detective uses other operations to generate sequences.

src/main/java/randoop/generation/ForwardGenerator.java
```diff
@@ -56,6 +57,12 @@ public class ForwardGenerator extends AbstractGenerator {
   /** The side-effect-free methods. */
   private final Set<TypedOperation> sideEffectFreeMethods;
 
+  /**
+   * The methods that might have side effects. This is the complement of {@link
+   * #sideEffectFreeMethods}.
+   */
+  private final Set<TypedOperation> sideEffectingMethods;
```
mernst: The pull request uses a variety of terms:  impure, side-effecting, mutating.  Choose one of them and use it consistently throughout.

src/main/java/randoop/generation/GrtFuzzer.java
```diff
@@ -27,7 +27,10 @@ public GrtFuzzer() {}
   /** List of all fuzzers. */
   @SuppressWarnings("ClassInitializationDeadlock")
   private static final List<GrtFuzzer> FUZZERS =
-      Arrays.asList(GrtNumericFuzzer.getInstance(), GrtStringFuzzer.getInstance());
+      Arrays.asList(
+          GrtNumericFuzzer.getInstance(),
+          GrtStringFuzzer.getInstance(),
+          GrtObjectFuzzer.getInstance());
```
mernst: Does this need to be instantiated so early?  The initialization code in `ForwardGenerator` is a bit complex and feels out of place.  Is there some other location where the fuzzers can be instantiated, so that the invocations of the fuzzer in `ForwardGenerator` don't have to possibly do initalization?

src/main/java/randoop/generation/ForwardGenerator.java
RockingMat: Which one works best? I know side-effecting is used in command line options. If I choose that, should I update all instances across codebase for mutating and impure?

src/main/java/randoop/generation/AbstractGenerator.java
RockingMat: I looked at the code for Detective and I don't see an additional store for more operations.

src/main/java/randoop/generation/GrtObjectFuzzer.java
RockingMat: The input is a sequence and a variable where the variable is in the sequence. The workhorse of the class is the Fuzz method which implements the interface method from GRTFuzzer.

mernst: Please add documentation to that effect.
(I had figured out the answer, and my question was intended to provoke documentation.  But my comment was not clear about that!)

src/main/java/randoop/generation/ForwardGenerator.java
mernst: "side-effecting" is probably best.  It is the most standard and easy-to-understand term.
Don't make unrelated changes in this pull request, but it would be good to create a new pull request to make the changes more globally.

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,284 @@
...
+      }
+      applicableOps = new ArrayList<>(opsSet);
+      typeToApplicableOps.put(rawType, applicableOps);
+    }
+
+    if (applicableOps.isEmpty()) {
+      return null;
+    }
+
+    // Refine: keep only operations that have at least one parameter truly assignable from
```
mernst: If this test is all about generics, please add a comment to that effect.  (And if it's not, comment what it is about.)

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,284 @@
...
+          break;
+        }
+      }
+    }
+
+    return compatibleOps.isEmpty() ? null : Randomness.randomMember(compatibleOps);
+  }
+
+  /**
+   * Chooses the index of a parameter in {@code mutationOp} whose type can accept the target
```
mernst: Types don't "accept" variables.  Use more direct language such as "assignable" or (better) "subtype".

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,284 @@
...
+
+        if (candidates.isEmpty()) {
+          // No sequence can satisfy this parameter - abort mutation.
+          return new VarAndSeq(variable, sequence);
+        }
+
+        Sequence candidateSeq = inputSequenceSelector.selectInputSequence(candidates);
+        Variable candidateVar = candidateSeq.randomVariableForTypeLastStatement(paramType, false);
+        if (candidateVar == null) {
+          // No variable of the required type in the candidate sequence.
```
mernst: Is this situation a bug?  Should the code throw an exception rather than returning a value?

RockingMat: This situation is just for the case where we chose a mutationop such that one of the parameters(not including the object we are fuzzing) is not in our component pool.

varuniy:
@mernst This is ready for re-review. Sorry for the delay, I had to get access to Tian's branch before I made a bugfix to pass the pipeline.

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,286 @@
...
+  /** The singleton instance. */
+  private static final GrtObjectFuzzer INSTANCE = new GrtObjectFuzzer();
+
+  /** Maps RAW type to mutating operations that have a parameter of that type. */
+  private final Map<Type, List<TypedOperation>> rawTypeToSideEffectingOps = new HashMap<>();
+
+  /** Component manager to get sequences for types. */
+  private @MonotonicNonNull ComponentManager componentManager;
+
+  /** Cache of applicable operations by raw type to avoid retraversing supertypes. */
```
mernst: What is the definition of "applicable"?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,286 @@
...
+  private GrtObjectFuzzer() {
+    // nothing to do; `initialize()` does the initialization.
+  }
+
+  /**
+   * Initialize this fuzzer with side-effecting operations and a component manager.
+   *
+   * @param mutators side-effecting operations used as mutators
+   * @param cm the component manager used to obtain sequences for required types
+   * @param selector strategy for choosing input sequences to satisfy non-fuzz parameters
```
mernst: Should "non-fuzz" be "non-fuzzed"?

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,286 @@
...
+      Set<TypedOperation> mutators, ComponentManager cm, InputSequenceSelector selector) {
+    if (this.componentManager != null) {
+      throw new RandoopBug("Do not call initialize multiple times.");
+    }
+    // Build the type-to-mutators map, for quick access later.
+    for (TypedOperation op : mutators) {
+      TypeTuple inputTypes = op.getInputTypes();
+      for (int i = 0; i < inputTypes.size(); i++) {
+        Type type = inputTypes.get(i).getRawtype();
+        rawTypeToSideEffectingOps.computeIfAbsent(type, k -> new ArrayList<>()).add(op);
```
mernst, line 93: I see that the list may have duplicates.  Is that intentional?  (Maybe, to increase the likelihood if there are two parameters of a given type??)

src/main/java/randoop/generation/GrtObjectFuzzer.java
```diff
@@ -0,0 +1,286 @@
...
+   */
+  private @Nullable TypedOperation selectMutationOperation(Type typeToFuzz) {
+    if (typeToFuzz.isNonreceiverType()) {
+      return null;
+    }
+
+    Type rawType = typeToFuzz.getRawtype();
+    // Start from a coarse, raw-type-based superset.
+    // Later code will refine this so the generic types match.
+    List<TypedOperation> applicableOps = typeToApplicableOps.get(rawType);
```
mernst: Abstract this code (through the blank line) into a method.

src/main/java/randoop/generation/GrtObjectFuzzer.java
varuniy: Yes, you are correct that a single `TypedOperation` can be added multiple times to the list for a raw type if it declares that raw type in more than one parameter slot. However, this doesn't actually change selection probability because we deduplicate later (we collect into a `LinkedHashSet` when building the candidate list). The duplicates are harmless functionally. Let me know if I'm overlooking something.

PR #1597, opened 2025-07-24
Use annotation name string comparison for method purity check

776styjsu:


776styjsu:
This reduces the size of the diff in #1595.

PR #1609, opened 2025-08-08
Enforce single-argument side-effect-free methods for regression checks

776styjsu:


mernst:
@776styjsu runNaiveCollectionsTest is timing out.  Could you investigate that?

mernst:
@776styjsu Can you provide a test case that fails before your changes and succeeds afterwards?

mernst:
Your test case (which is included in #1618) was very helpful!

PR #1625, opened 2025-08-15
Add `--replacecall-exclude` option to skip specific default method replacements

776styjsu:


agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -167,6 +179,54 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+            }
+          } catch (IOException e) {
+            System.err.printf(
+                    "Error reading replacement exclusion file %s:%n %s%n",
+                    excludeFile, e.getMessage());
+            System.exit(1);
+          }
+        }
+
+        // 1) Remove any method-level replacement entries that exactly match excluded signatures.
```
mernst: I find the logic of 1) here and 2) below a bit difficult to understand, and it uses several temporary data structures.
Can you rewrite it to iterate over `excludeSignatures` just once, taking different action depending on whether the method is in `replacementMap`?

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -233,6 +293,25 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
   }
 
+  /**
+   * Trims the given string and parses it as a {@link MethodSignature} to add to the set of
+   * excluded signatures.
+   *
+   * @param excludeSignatures the set of excluded method signatures to which the parsed signature is added
+   * @param text the string to parse as a method signature
+   */
+  private static void trimAndParseExcludedSignature(Set<MethodSignature> excludeSignatures, String text) {
```
mernst: It is surprising that two methods with very similar names `trimAndParseExcludedSignature` and `parseExcludedSignature` have very different contracts.  One does what the same says.  The other one adds to a specific list.  I suggest putting the call to `trim()` in the one that does parsing (and perhaps the warning as well), and rename or inline the one that adds to a list.

agent/replacecall/src/main/java/randoop/instrument/ReplaceCallAgent.java
```diff
@@ -260,6 +339,98 @@ private static Set<String> loadExclusions(@Owning Reader exclusionReader, String
...
+  }
+
+  /**
+   * Converts a JVM method descriptor to a comma-separated list of binary names suitable for
+   * {@link MethodSignature#of(String)}. For example,
+   *   "(I)V" -> "int"
+   *   "([Ljava/lang/String;I)V" -> "java.lang.String[], int"
+   * Returns null if the input doesn't look like a valid method descriptor.
+   */
+  private static @Nullable String descriptorParamsToBinaryNames(String methodDescriptor) {
```
mernst: The implementation of this method can be shorter by using methods from [reflection-util](https://plumelib.org/reflection-util/api/org/plumelib/reflection/Signatures.html).

mernst (changes requested):
Thanks for this useful functionality.  I have some questions and suggestions about how users will use it.

**The colon format**

1. Where is the colon format described?  Why don't you use instead a standard form, such as [method descriptor](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.3.3)?
2. Why is the colon format different than the format that is already used for replacements, in the replacement file?  The discrepancy is surprising and is likely to be inconvenient for users.
3. The method in colon form is immedately converted to the form required by `MethodSignature`.  Why not require the `MethodSignature` form?

I think the second choice is best, for consistency.

**The API**

This pull request adds two new command-line arguments.  It might be more convenient to users to avoid adding new command-line arguments.
Two possibilities are:

 * Permit the exclusion file `dont-transform.txt` to contain, in addition to packages, methods.
 * Permit the replacement file `replacecall-replacements.txt` to contain, in addition to a line containing two whitespace-separated method signatures without return types, a line containing a method signature, whitespace, and some special text (such as "DO NOT REPLACE").

PR #1632, opened 2025-08-25
Fix directory name for the distribution zip

zzjas:
After unzip the current release file [randoop-4.3.4.zip](https://github.com/randoop/randoop/releases/download/v4.3.4/randoop-4.3.4.zip), a directory named `extension 'base' property 'archivesName'-4.3.4` will be created.

This PR fixes the `build.gradle` to correctly set the name. With the fix, unzipping `randoop-4.3.4.zip` will create a dir named `randoop-4.3.4`.


mernst (approved):
Thanks very much for this fix!

PR #1731, opened 2025-12-25
Add type check before casting to GenericClassType in castToRunTimeType()

varuniy:


src/main/java/randoop/sequence/ExecutableSequence.java
```diff
@@ -479,7 +479,7 @@ public boolean castToRunTimeType() {
       // For a getClass() call, the index of receiver is 1 since the lastValues will always
       // be [output, receiver].
       ReferenceType elemType = lastValues.get(1).getType();
-      if (elemType.isGeneric()) {
+      if (elemType.isGeneric() && elemType instanceof GenericClassType) {
```
mernst: When this test fails, what is the class of `elemType`?  Should this method contain code to handle that case?  (Because `elemType` is still generic in that case.)

varuniy:
@mernst Ready for re-review.

mernst (approved):
Thanks!

PR #1803, opened 2026-05-18
Fix TF-IDF literal statistics inconsistency causing Randoop Bug

varuniy:
Fixed a bug where constants appearing only in static final field initializers were being tracked with zero usage frequency, causing RandoopBug exceptions in TfIdfSelector when [numUses=0] but [numClassesWith>0]. These constants appear in the bytecode constant pool but are never directly used in executable code. The fix skips any constant with frequency <= 0 during literal extraction, ensuring only genuinely used literals are mined for test generation.


varuniy:
@mernst Ready for review.

mernst (approved):
Thanks!

PR #1804, opened 2026-05-20
GRT String Fuzzing Bugfix

varuniy:
Fixed a bug with String Fuzzing


varuniy:
@mernst Ready for review

mernst (approved):
Thanks for noticing and fixing this bug.

PR #1805, opened 2026-05-22
Make Gradle handle submodules/worktrees

mernst:


smillst (approved):
I think this is ok.  You may want to look at the coderabbit comments.

PR #1818, opened 2026-06-02
Include last element in getAllValues loop

AgusNolasco:
Change loop condition to include the last element in getAllValues method.


mernst (approved):
Thank you for noticing and fixing this bug.

PR #1822, opened 2026-06-03
Fix RandoopBug crash when type variable is uninstantiated in generic library methods

varuniy:
## PR Description

**Problem**

Randoop crashes with `RandoopBug: no run-time class for a type variable` when analyzing or executing methods from generic library classes (e.g., ASM's `Frame<V>.merge(...)`) where the type variable `V` is never substituted with a concrete type. Two separate code paths hit this:

1. **Input selection**: `SequenceCollection.getSequencesForType()` calls `isNonreceiverType()` on a candidate type that is an uninstantiated `ExplicitTypeVariable`. The base implementation of `isNonreceiverType()` in `Type` calls `getRuntimeClass()`, which throws for all `ParameterType` instances.

2. **Execution**: `TypedClassOperationWithCast.execute()` tries to cast the return value of a method whose output type is still a type variable, again calling `getRuntimeClass()` and throwing.

**Fix**

- ParameterType.java: Override `isNonreceiverType()` to return `false` directly. A type variable can never be a primitive, boxed primitive, `String`, or `Class`, so no runtime class lookup is needed.

- TypedClassOperationWithCast.java: Guard the cast in `execute()` -- if the output type is a `ParameterType` (uninstantiated type variable), skip the cast and return the execution result as-is.

### DIFF
--- a/src/main/java/randoop/operation/TypedClassOperationWithCast.java
+++ b/src/main/java/randoop/operation/TypedClassOperationWithCast.java
@@ -5,6 +5,7 @@ import randoop.ExecutionOutcome;
 import randoop.NormalExecution;
 import randoop.condition.ExecutableSpecification;
 import randoop.types.ClassOrInterfaceType;
+import randoop.types.ParameterType;
 import randoop.types.Substitution;
 import randoop.types.Type;
 import randoop.types.TypeTuple;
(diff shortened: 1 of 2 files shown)


mernst (approved):
Thanks for the fix!

PR #1844, opened 2026-07-20
Fix ClassCastException crash from GRT Elephant-Brain casts

varuniy:
`UncheckedCast.execute()` let `ClassCastException` propagate uncaught when a reused elephant-brain-inserted cast targeted a stale/incorrect runtime subtype, crashing generation entirely; this catches the exception and returns an `ExceptionalExecution` instead, so Randoop discards/truncates the offending sequence like it does for any other reflective failure.


varuniy:
@mernst Ready for review

mernst (approved):
Thanks!

